PR #1397, opened 2024-09-12
Fix/rehydration

### LINKED ISSUE #1316: Orleans based Aggregate Roots does not need to rehydrate from the beginning if the grain is already in memory
We should have a similar mechanism to what we have for immediate projections were it doesn't rehydrate all events relevant if the grain is already in memory. Instead it should just apply any new events since the last rehydration.

woksin:
Fixed #1316


woksin:
This is not completely done yet. I would like to have more integration tests

Source/Clients/AspNetCore/Transactions/UnitOfWorkMiddleware.cs
```diff
@@ -21,6 +21,7 @@ public class UnitOfWorkMiddleware(IUnitOfWorkManager unitOfWorkManager, RequestD
     /// <returns>Awaitable task.</returns>
     public async Task InvokeAsync(HttpContext context)
     {
+        // TODO: Create Issue: Maybe there should be a try catch here and dispose the unit of work if an exception is thrown.
```
einari, line 24: What could cause an exception here? Are you seeing exceptions?
From what I can see from the current implementation all this does is create a new UnitOfWork instance and return this.

Source/Clients/DotNET/Aggregates/AggregateRootContext.cs
```diff
@@ -33,5 +35,5 @@ public class AggregateRootContext(
     public IUnitOfWork UnitOfWOrk { get; } = unitOfWork;
 
     /// <inheritdoc/>
-    public bool HasEventsForRehydration { get; set; }
+    public EventSequenceNumber NextSequenceNumber { get; set; } = nextSequenceNumber;
```
einari, line 38: We can probably remove the setter here.

einari (changes requested):
Great work! A few minor things to address.

Source/Clients/DotNET/Aggregates/IAggregateRootContext.cs
```diff
@@ -35,7 +35,7 @@ public interface IAggregateRootContext
     IUnitOfWork UnitOfWOrk { get; }
 
     /// <summary>
-    /// Gets the last processed <see cref="EventSequenceNumber"/>.
+    /// Gets the next <see cref="EventSequenceNumber"/> to process for the aggregate root in the unit of work.
```
einari, line 38:
```suggestion
    /// Gets or sets the next <see cref="EventSequenceNumber"/> to process for the aggregate root in the unit of work.
```

einari (changes requested):

PR #1402, opened 2024-09-13
WIP: Redesign Aggregate root unit of work

woksin:
## Summary

needed.

### Added

- Describe the added features

### Changed

### Fixed

### Removed

### Security

### Deprecated


einari:
I've started an issue for fixing the issues #1405 with an initial design I think can work.

PR #1421, opened 2024-09-20
Fix/deadlocking reactor

woksin:
### Added

- Some more logging

### Fixed

- DotNET Client Reactor deadlock issue for async Reactor event handlers because of manual call on ObserveMethod(...).Wait would some times block the thread that was running the Reactor event handler method. This problem was proven in integration tests and fixed by using an observable pattern for dealing with asynchronous OnNext delegates using `Observable.FromAsync` https://github.com/dotnet/reactive/issues/459
- The same issue as described above, but for Reducers as well
- CPU / thread pool thread hogging in. `AppendedEventsQueue` because of the `ManualResetEventSlim` and calling `.Wait` on it which would cause the thread to be blocked. Implemented `AsyncManualResetEvent` which is just a new async version of ManualResetEventSlim based off Stephen Toub's suggestion https://devblogs.microsoft.com/pfxteam/building-async-coordination-primitives-part-1-asyncmanualresetevent/ 
Avoiding `ManualResetEvent` is also a better solution because it uses busy spinning which is only optimized for when you expect the wait time to be short. https://learn.microsoft.com/en-us/dotnet/api/system.threading.manualreseteventslim?view=net-8.0#remarks
- Integration test speed increase (from 10-15 min to 35 sec on my machine)

### DIFF
--- a/Integration/Orleans.InProcess/OrleansFixture.cs
+++ b/Integration/Orleans.InProcess/OrleansFixture.cs
@@ -118,7 +118,7 @@ public class OrleansFixture(GlobalFixture globalFixture) : WebApplicationFactory
             _backupPerformed = true;
         }
         GlobalFixture.RemoveAllDatabases().GetAwaiter().GetResult();
-        base.Dispose(disposing);
+        base.Dispose(false);
     }
 
     protected virtual void ConfigureServices(IServiceCollection services)
--- a/.github/workflows/dotnet-build.yml
+++ b/.github/workflows/dotnet-build.yml
@@ -1,5 +1,8 @@
 name: .NET Build
 
+env:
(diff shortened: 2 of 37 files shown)


Integration/Orleans.InProcess/OrleansFixture.cs
```diff
@@ -118,7 +118,7 @@ protected override void Dispose(bool disposing)
             _backupPerformed = true;
         }
         GlobalFixture.RemoveAllDatabases().GetAwaiter().GetResult();
-        base.Dispose(disposing);
+        base.Dispose(false);
```
woksin, line 121: This sped up the integration tests a lot! Though there is probably something in Chronicle which causes it to be so slow on shutdown that we have to figure out some time

PR #1550, opened 2024-11-24
Fix/appended-events-queue

einari:
## Summary

needed.

### Added

- Describe the added features

### Changed

### Fixed

### Removed

### Security

### Deprecated

### DIFF
--- a/Source/Kernel/Grains/EventSequences/AppendedEventsQueue.cs
+++ b/Source/Kernel/Grains/EventSequences/AppendedEventsQueue.cs
@@ -8,6 +8,7 @@ using Cratis.Chronicle.Concepts.Events;
 using Cratis.Chronicle.Concepts.Observation;
 using Cratis.Chronicle.Grains.Observation;
 using Cratis.Chronicle.Tasks;
+using Cratis.Tasks;
 
 namespace Cratis.Chronicle.Grains.EventSequences;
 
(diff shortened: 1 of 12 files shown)


Source/Kernel/Grains/EventSequences/AppendedEventsQueue.cs
```diff
@@ -64,34 +69,87 @@ public Task Unsubscribe(ObserverKey observerKey)
...
                     }
-                    var observer = _grainFactory.GetGrain<IObserver>(subscription.ObserverKey);
-                    var partition = actualEvents[0].Context.EventSourceId;
-                    await observer.Handle(partition, actualEvents);
                 }
-
-                _queueEvent.Reset();
+                catch
+                {
+                    // We ignore any failures, the queue should never fail
```
woksin, line 147: Wouldn't it be good to log something here?

Source/Kernel/Grains/EventSequences/AppendedEventsQueue.cs
```diff
@@ -64,34 +69,87 @@ public Task Unsubscribe(ObserverKey observerKey)
...
+                            if (actualEvents.Count == 0)
+                            {
+                                continue;
+                            }
+                            var observer = _grainFactory.GetGrain<IObserver>(subscription.ObserverKey);
+                            var partition = group.Key;
+                            tasks.Add(observer.Handle(partition, actualEvents));
+                        }
+
+                        await Task.WhenAll(tasks);
```
woksin, line 142: While we're at it I think it could be useful to write an alternative path for when the events list only contains one event, which seems to be the case most of the time.

If we write a separate path for handling a single event we can avoid unnecessary allocations. It's a small performance optimization that in my opinion is pretty cheap :)

woksin (approved):
Approved but with some suggestions

PR #1638, opened 2025-01-08
Feature/observer config

woksin:
## Summary

Barebones implementation of configuration for observers. This will be fleshed out later

### Added

- Configuration for observers for configuring retry policies using asp net configuration system for the Cratis Server
  - SubscriberTimeout - defaults to 5 seconds
  - MaxRetryAttempts - defaults to 10
  - BackoffDelay - defaults to 1 second
  - ExponentialBackoffDelayFactor - defaults to 2
  - MaximumBackoffDelay - defaults to 600 seconds = 10 minutes

### DIFF
--- a/Source/Kernel/Concepts/Configuration/Observers.cs
+++ b/Source/Kernel/Concepts/Configuration/Observers.cs
@@ -11,7 +11,7 @@ public class Observers
     /// <summary>
     /// Gets the timeout in seconds for observer calling its subscriber.
     /// </summary>
-    public int SubscriberTimeoutInSeconds { get; init; } = 5;
+    public int SubscriberTimeout { get; init; } = 5;
 
     /// <summary>
     /// Gets the maximum number of retries that can be attempted on a failed observer partition.
@@ -24,15 +24,15 @@ public class Observers
     /// <summary>
     /// Gets the delay for attempting to retry a failed partition in seconds.
     /// </summary>
-    public int RetryDelayInSeconds { get; init; } = 1;
+    public int BackoffDelay { get; init; } = 1;
 
     /// <summary>
     /// Gets the retry delay exponential factor.
     /// </summary>
-    public int RetryDelayExponentialFactor { get; init; } = 2;
+    public float ExponentialBackoffDelayFactor { get; init; } = 2;
 
     /// <summary>
     /// Gets the max delay time in seconds for retrying a failed partition.
     /// </summary>
-    public int MaxRetryDelayInSeconds { get; init; } = 60 * 10;
+    public int MaximumBackoffDelay { get; init; } = 60 * 10;
 }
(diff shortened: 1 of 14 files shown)


Source/Kernel/Concepts/Configuration/Observers.cs
```diff
@@ -0,0 +1,38 @@
...
+    /// </summary>
+    /// <remarks>
+    /// 0 represents infinite number of retries.
+    /// </remarks>
+    public int MaxRetryAttempts { get; init; } = 10;
+
+    /// <summary>
+    /// Gets the delay for attempting to retry a failed partition in seconds.
+    /// </summary>
+    public int RetryDelayInSeconds { get; init; } = 1;
```
einari, line 27: I realize after some thinking (I know, that can be dangerous), maybe it would be better to use the terminology **Backoff**, since that is well understood in resilience lingo.

e.g. `BackoffDelayInSeconds`

Another question that also popped into my head; the "InSeconds" part, is that needed, or is it sufficient just to keep it documented in the XML comment and in general documentation?

Then it would be `BackoffDelay`

Source/Kernel/Concepts/Configuration/Observers.cs
```diff
@@ -0,0 +1,38 @@
...
+
+    /// <summary>
+    /// Gets the retry delay exponential factor.
+    /// </summary>
+    public int RetryDelayExponentialFactor { get; init; } = 2;
```
einari, line 32: Same as above;

`ExponentialBackoffDelayFactor`

Plus, we should make this a float. That way you can have 1.5 or whatever you see fit.

Source/Kernel/Concepts/Configuration/Observers.cs
```diff
@@ -0,0 +1,38 @@
...
+
+    /// <summary>
+    /// Gets the max delay time in seconds for retrying a failed partition.
+    /// </summary>
+    public int MaxRetryDelayInSeconds { get; init; } = 60 * 10;
```
einari, line 37: Same as above:

`MaximumBackoffDelay`

Source/Kernel/Concepts/Configuration/Observers.cs
```diff
@@ -0,0 +1,38 @@
...
+
+/// <summary>
+/// Represents the configuration for observers.
+/// </summary>
+public class Observers
+{
+    /// <summary>
+    /// Gets the timeout in seconds for observer calling its subscriber.
+    /// </summary>
+    public int SubscriberTimeoutInSeconds { get; init; } = 5;
```
einari, line 14: Perhaps drop the `InSeconds` in name and leave that it for XML comments and documentation.
(see below commend)

einari (changes requested):
Looks good. Have a couple of suggestions and thoughts related to naming of the configuration properties.

einari (approved):

Issue #1686, opened 2025-01-31
Jobs RequestType discriminator seems to not work every


einari:
We're getting exceptions like the following:

![Image](https://github.com/user-attachments/assets/071858c1-1080-46e8-ba9c-ec0a08de772d)

einari:
Suggestion to fix:
- Custom MongoDB Serializer
- Introduce an Enum Type for the different JobTypes that we store on the JobState and use this as a discriminator in the serializer

einari:
The lock in to the CLR type is actually the biggest problem with this. It would then potentially break a system from one version to another with Jobs being hydrated while upgrading if we change the namespace or name of the job. 

We would need to have something that holds a map of `JobType` string to the CLR Type.

```csharp
public interface IJobTypeManager
{
     JobType GetJobTypeFor(Type type);
     Type GetClrTypeFor(JobType type);
     Type GetRequestClrTypeFor(JobType type);
}
```

This could discover anything that implements `IJob` and also the `IJob<TRequest>` interface.
By default it would take the CLR Type.Name as the `JobType` while discovering.

But we could easily support overriding this, which would prove handy when renaming and wanting
to have backwards compatibility.

By introducing a `JobTypeAttribute`:

```csharp
[....]
public class JobTypeAttribute(string jobType);
```

On a Job we would typically then add:

```csharp
[JobType]
public class CatchUpObserver
{
}
```

woksin:
Good idea. That makes sense

einari:
And probably even better name for it would be `IJobTypes` so we don't add those pesky postfixes we don't like :)

woksin:
Pr her #1693 however I'm having issues understanding how to implement the bson serializer

PR #1693, opened 2025-02-04
Fix/job request deiscriminator

woksin:
## Summary

needed.

### Added

- Describe the added features

### Changed

### Fixed

### Removed

### Security

### Deprecated


woksin:
I'm a bit stuck here on how to actually do the Bson serializer part... @einari

einari:
I've done something that might be useful for this. In fact, this is something that was accidentally removed from ApplicationModel and should be put back.

Fundamentals has the concept of derived types, which also stores a discriminator. This is done using a Guid, but our case would be the String representation.

It leverages the Convention system in the MongoDB driver:

https://github.com/einari/MongoDB/blob/main/Source/DerivedTypeDiscriminatorConvention.cs

This is the serializer:

https://github.com/einari/MongoDB/blob/main/Source/DerivedTypeSerializer.cs


This is the provider that provides the serializer:

https://github.com/einari/MongoDB/blob/main/Source/DerivedTypeSerializerProvider.cs

woksin:
The issue here is that we need to know the JobType when serializng the IJobRequest so we need a serializer for JobState as well? That's the part i don't understand, the problem is serializing the IJobRequest, we want to know it's concrete clr type consistently and we only know that through the JobType. As far as I understand that was the idea

einari:
The discriminator approach would do just that. The only downside is that the JobType would then be part of the Request BSON object.

An option here would be to stick the JobType on the IJobRequest interface. Then the discriminator convention approach would be fine and we wouldn't need a custom serializer for JobState.

I think this option is fine. If we want to expose JobType on the JobState level as well, we would need to keep the two in sync. But I don't think we would need it on the JobState level, as the Request is not optional (if I remember correctly)

woksin:
Would be nice if we could getting a reference to the parent bson document when we serialized the IJobRequest 😓

einari:
I agree, its not ideal and it would be the most natural to have in on JobState. 

I was kinda hoping the IBsonReader could access the parent, but it is a forward only type of thing. 

But, maybe we can make this work. 

Here's a thought;
What if we have a custom serializer for JobType. We keep the JobType on JobState. The only job the customer serializer has is to set a "context" for the discriminator to get. This could be a an AsyncLocal thing. The simplest thing that could possibly work would be that the discriminator convention has an internal static AsyncLocal of JobType that the JobType serializer sets. 
As for serialization, the JobType serializer would just read and write a string.

Could this work?

einari:
I slept on it, and realized that this will only work if the deserialization is deterministic in the sense that we can guarantee that the JobType property will be deserialized before the Request consistently.

An alternative could be to have a custom serializer for JobState that inherits from the base MongoDB serializer and just forwards to the base for implementation for both serialization and deserialization. The only thing for deserialization would be to use the Reader on the context before forwarding the call to determine the JobType. It can then use the bookmark functionality to get a bookmark, find the JobType element then get back to the bookmark. Just as the link to the DiscriminatorConvention thing I had earlier.
The JobType would then be something you set on the AsyncLocal<> of the JobTypeDiscriminatorConvention thing as mentioned in my previous comment.

woksin:
I don't necessarily like the static AsyncLocal, but I cannot see any other good solution to this.

Btw @einari I saw that the JobStepState also has this Request object, which causes the exact same problem I guess but just for JobStepState instead, which does not have the JobType on it. We probably don't need to store the Request on the JobStepState so I would prefer to remove it.

einari:
Agree. That is not needed there since a JobStep shouldn't be hanging around in the database without the Job being there. 

An alternative to the AsyncLocal and the whole Discriminator thing is to create a JobState serializer that deals with it on its own. 
Meaning that it will have to then run through all the properties and ignore the Request property. When it has all the other properties, including JobType, it can at the very end use the BsonReader to go back and find the Request object and deserialize it based on the type (remember to use the Bookmark functionality before deserializing all the other properties).

This would be a less "magical" way of doing things and way more robust, IMO. Since we're not going to have the Request on anything else, its a bit overkill with the discriminator approach.

woksin:
@einari BsonReaderBookmark did the trick for me :) 

I have one question though in relation to removing Request from JobStepState,
We have this class in Setup project which I have no idea what does:
```csharp

/// <summary>
/// Represents the <see cref="JsonConverter{T}"/> that can convert <see cref="JobStepState"/>.
/// </summary>
public class JobStepStateJsonConverter : TypeWithObjectPropertiesJsonConverter<JobStepState>
{
    /// <inheritdoc/>
    protected override IEnumerable<string> ObjectProperties => [nameof(JobStepState.Request)];
}

```

einari:
Great. 

That Json converter can probably be deleted. I believe it was there for the frontend APIs. But the APIs are now using the gRPC contracts and if we want to expose the request information in the APIs, we'll figure something out. Like we do :)

woksin:
And now we also have the same issue for Orleans serialization, cannot serialize an interface 😢 
Screenshot 2025-02-05 at 14 45 26

woksin:
Do we have to do kind of the same thing for the json serializer?

woksin:
This is also very annoying, this is just the same problem as with the MongoDB serializer, where we cannot create a converter just for IJobRequest because we don't have JobType and no way to reference parent object. And it's even more anoying with JsonConverter for JobType because there I don't have access to the service provider so I cannot use IJobTypes either....

einari:
Well, depends. 
What we've done with for instance the DerivedTypes and a few other things in Fundamentals is that we have a singleton instance we can get access to. This could be an internal field. And we would still access it normally for those things that can have it injected via the interface. 

E.g.:

https://github.com/Cratis/Fundamentals/blob/169f8c52cfc50a070f048b8f7f53c6a2c82f85ce/Source/DotNET/Fundamentals/Serialization/DerivedTypes.cs#L24

In globals we use it:

https://github.com/Cratis/Fundamentals/blob/169f8c52cfc50a070f048b8f7f53c6a2c82f85ce/Source/DotNET/Fundamentals/Json/Globals.cs#L26


Furthermore, we could have the JsonConverter have 2 constructors, one that takes the IJobTypes as a dependency (making it testable with mocks) and a default one that forwards to this passing the `JobTypes.Instance`.

woksin:
If this is the way we have to go then I guess we don't have much better alternatives 😄

einari:
There are always alternatives :) 
Like the one we have done in ApplicationModel:

https://github.com/Cratis/ApplicationModel/blob/main/Source/DotNET/Applications/Internals.cs

Its basically exposing the ServiceProvider internally. 

This could then be set like we do in AppModel:

https://github.com/Cratis/ApplicationModel/blob/3f4d341941303c11219aa0d3014197cb83c37886/Source/DotNET/Applications/ApplicationBuilderExtensions.cs#L20


Either way, its not optimal and perfect. They're all trade-offs. Pragmatically speaking, its fine.

einari:
But for this use case, its probably better with my previous suggestion, as the ServiceProvider is probably not ready at the time of setting up the serializer options.

woksin:
fixes #1686

PR #1697, opened 2025-02-06
Fix/event constraint append many

woksin:
## Summary

Fixed an issue in the Kernel EventSequence which could cause problems in terms of appending multiple events.

### Fixed

- When appending many events all event constraints are checked before any events are appended


woksin:
@einari I noticed we have introduced a problem with our specs some weeks ago when we made our AppendedEventsQueue handle partitioned events in parallel. The tests fails at random because of this :)

woksin:
@einari Don't merge this, we have a mega bug in AppendedEventsQueue StartQueueHandler. It seems to continuous start queue handlers for some reason due to the _queueTask.ContinueWith

woksin:
@einari This pr fixes that issue I mentioned : https://github.com/Cratis/Chronicle/pull/1698

PR #1750, opened 2025-03-08
Fix/minor-fixes

einari:
### Fixed

- The internal `JobTypes` system is used inconsistently with places asking for `JobType` based on Grain concrete and others by the interface. The base `Job` using just `GetType()`, which is the information it has, while other places the interface - which is the correct type to use. Changed this internally to work with the interface type while supporting getting `JobType` from a Clr type that can be either.
- Adding name of event store to the `NamespaceAdded` internal message.
- Introducing a `ProjectionsService` which is a `GrainService` that runs on every silo. Its responsibility is to populate an internal in-memory representation of projections and their definitions per event store and namespace. This was completely left incomplete, leading replay to not call `BeginReplay` on the projection pipeline.
- Making sure the in-memory projection definitions are updated when a `NamespaceAdded` message occurs.


woksin:
I see that I also fixed the issue you fixed here with JobType in my stupid long lived branch :)

PR #1866, opened 2025-04-14
Feature/correlation and tracing

woksin:
## Summary

When using an Orleans based client the CorrelationId of the appended events would always be the empty Guid due to the nature of Orleans message call contexts not cooperating with AsyncLocal which ICorrelationIdAccessor used.

### Added

- Optional parameter for `CorrelationId` in `IEventSequence` append methods

### Fixed

- `UnitOfWork` uses its own `CorrelationId` when appending events when committing


einari:
Could we hold off on this PR a bit.
I would like to explore options for getting the CorrelationId truly cross cuttingly and readily available without performance implications.

I'm very hesitant to adding CorrelationId to public APIs, as I consider it sort of internals.

PR #1900, opened 2025-05-22
Fix CreateClient in OrleansFixture integrationt test package

woksin:
### Fixed

- An issue with `OrleansFixture.CreateClient` method that would throw an exception when called with `WebApplicationFactoryClientOptions`


Source/Clients/XUnit.Integration/OrleansFixture.cs
```diff
@@ -65,7 +67,9 @@ public OrleansFixture(ChronicleFixture chronicleFixture)
         var configureServices = ConfigureServices;
         _webApplicationFactory = (Activator.CreateInstance(_webApplicationFactoryType, [this, configureServices, _contentRoot]) as IAsyncDisposable)!;
     }
-
+    
+    public virtual bool AutoDiscoverArtifacts { get; }
```
einari, line 71: Needs an XML comment

woksin:
It seems like we have issues with out integration tests. Let's fix this once and for all 😄

Issue #2644, opened 2026-02-14
Add the ability to add events directly in the Event Sequence editor / page in Workbench


einari:
In the Workbench in the EventSequence view, we want to able to do "inline" adding of events. 
This should work by having a "+ Add Event" button at the top. The added event should show up as a row at the top of the list in the view with a save Icon button to the left. The event type should be a drop down and the content needs to then be a "..." and a separate dialog for filling it out. You should not be allowed to save unless all properties that are not nullable in the event content has been filled out.

To be able to do this we need to be able to then append through the REST API with a command and it hooked up with the gRPC service.

Once the event is OK and saved, it will get an event sequence number at the tail of the sequence. We then need to refresh the data and if the event is beyond the page we're on, we should navigate to the last page and refresh there.

Issue #2642, opened 2026-02-14
Implement compensation support


einari:
In the EventSequence Grain API we have support for doing `Compensate` on an event. The idea is to be able to correct values for an event. We don't correct the original values directly, but instead we should leverage the `Compensations` part of the stored event and then keep track of any compensations that has happened to an event. Compensation is for the same event type that the original. 
The `EventCompensation` needs to keep the `EventTypeGeneration` in storage that was for the compensation.

When reading events, we would return the compensated events. 

As with Redact, we should rewind affected observers (`RewindPartitionForAffectedObservers`).

Tasks:

- Implement Compensate method in MongoDB + add new property on EventCompensation for the generation
- Implement Compensate method in EventSequence Grain - need to pull the event so that one can validate that the event type is the same - not necessarily the generation.
- When reading events using any of the Get* methods for getting events should then return the event with the compensation applied.
- Add support for Compensate in the gRPC Contracts and implement it in the EventSequence Service implementation
- Add support in the API (Command - Model Bound) that can do compensation
- Add a button in the EventSequence page in the Workbench for doing compensation

Issue #2640, opened 2026-02-14
Hook up redaction from Workbench


einari:
An EventSequence has the capability of redacting events, which basically means swapping out an existing event to a different type called EventRedacted. We need to support this from the Workbench frontend as well. 

- Add a button in the EventSequence page that is enabled when an event is selected called `Redact`.
- Clicking it should show a dialog where one needs to fill out the reason for the redaction
- Add necessary API in the form of a Command (Model Bound) that then calls the gRPC API for redacting an event.

Issue #2638, opened 2026-02-14
Compatibility checks for clients when connecting to server


einari:
Goal is to validate that the connecting client is compatible with the server for the DotNET Client. This can be done in the Connections project, as we use that for other things as well. This should be done with the server exposing an endpoint in the `IConnectionService` for getting the descriptor for the entire schema exposed for all services. The client needs to then do the validation for checking if it is compatible and can connect. If it is not compatible, it should log and error and not try to connect.

## The pattern

	1.	Server generates a FileDescriptorSet from its code-first service contracts at startup.
	2.	Server exposes it via a normal RPC, e.g. GetDescriptorSet() returning bytes.
	3.	Client generates its own FileDescriptorSet from its local contracts.
	4.	Client runs a protobuf wire-compatibility check:

## Fail if (hard breaks):

	•	A method the client needs is missing on the server (service/method name mismatch).
	•	Same field number exists on both sides but wire type differs (varint vs length-delimited vs fixed32 vs fixed64).
	•	Same field number exists but one side is repeated packed vs non-packed in a way that changes wire encoding expectations.
	•	Oneof/tag reuse conflicts.

## Allow (non-breaking):

	•	Server has extra fields or extra methods.
	•	Client has extra fields in requests (server ignores unknowns).
	•	Server returns extra fields (client ignores unknowns).
	•	Missing response fields on server (client sees default) — unless you choose to treat “presence required” as incompatible (schema alone can’t know your app semantics).

## How to generate descriptor sets (code-first)

Use the protobuf-net.Grpc schema/descriptor generator (the “reflection” package is just a generator; you’re not enabling the standard gRPC reflection service).
	•	Server: generate descriptor set from the service interfaces you host.
	•	Client: generate descriptor set from the service interfaces it was compiled with.

Then compare.

## Practical notes (protobuf-net specifics)
	•	Your compatibility hinges on stable field numbers ([ProtoMember(n)]) and stable service/method names.
	•	If you rely on implicit numbering anywhere, you’ll get false incompatibilities (or real breaks) — make numbering explicit.
	•	Prefer comparing descriptor-set bytes (normalized) rather than .proto text. The pattern
	1.	Server generates a FileDescriptorSet from its code-first service contracts at startup.
	2.	Server exposes it via a normal RPC, e.g. GetDescriptorSet() returning bytes.
	3.	Client generates its own FileDescriptorSet from its local contracts.
	4.	Client runs a protobuf wire-compatibility check:

[...]

PR #2647, opened 2026-02-14
15.4.0

einari:
### Added

- IServer.GetVersionInfo() endpoint exposing server version and commit SHA.
- disableWhen prop on DataPage menu items for conditionally disabling Workbench actions based on item state.
- Added a "raw" TypeScript client as a starting point (#2437).
- Adding support for Microsoft Aspire (#2798).
- Allowing to specify `Admin` password for development image, to avoid having to set this every time one resets the system (#2794).
- Support for integration between event stores through outbox -> inbox pattern (#2831).
- Making it possible to do ad-hoc queries using the projection engine, inferring types automatically (#2764).
- Exposing the PROTO files for the gRPC API (#2437).
- Adding the ability to specify seed data specific to a namespace (#2433).
- Adding a viewer for seeding seed data, both the global data for an event store and the data per namespace (#2433).
- Adding redaction of events in Workbench (#2640).
- Adding the ability to add events from the Workbench (#2644).
- Automatic cleanup of `dead in the water` jobs (#1685).
- It is now possible to specify `Occurred` when appending events - handle with care. Default behavior is still the same that the Kernel will assign this (#2729).
- Support for event compensation (#2642).
- Adding strict validation of event schema when appending (#287).
- Add button for testing WebHooks before adding them, validation will also happen when adding (#2731).
- Support for event migrations (#1859).
- Build/pre-publish time gRPC contract binary compatibility check (#2651).
- Runtime gRPC contract compatibility check when client connects (#2638).
- Support for constant (literal) keys in projections (#2737).
- **C# declarative projections**: `UsingConstantKey(string value)` on `IKeyBuilder`/`KeyBuilder`/`IReadModelPropertiesBuilder` and `UsingConstantParentKey(string value)` on `IParentKeyBuilder`/`ParentKeyBuilder`/`KeyAndParentKeyBuilder`/`IReadModelPropertiesBuilder`

  ```csharp
  public class Projection : IProjectionFor<MyReadModel>
  {
      public void Define(IProjectionBuilderFor<MyReadModel> builder) => builder
          .From<MyEvent>(e => e
              .UsingConstantKey("some value"));
  }
  ```

- **ModelBound projections**: `ConstantKey` property on `FromEventAttribute`, `CountAttribute`, `IncrementAttribute`, and `DecrementAttribute`; also exposed on their respective interfaces (`IFromEventAttribute`, `ICountAttribute`, `IIncrementAttribute`, `IDecrementAttribute`)

  ```csharp
  [FromEvent<MyEvent>(ConstantKey = "some value")]
  public record MyReadModel(
      [Increment<MyEvent>(ConstantKey = "some value")] int Counter);
  ```

- **Projection Declaration Language**: `literal` keyword for constant key expressions; compiles to `$value(...)` internally

  ```
  projection MyProjection => MyReadModel
    from MyEvent
      key literal "some value"
  ```

### Fixed

- Fixing a raise condition when connecting a client that registers projections and seed events simultaneously (#2741).
- Bundling the Monaco editor used in Workbench for offline support and faster load time on slower connections (#2739).
- Fixing `Decrement` builder - it was translated to an `Increment` for projections (#2735).
- WebHooks weren't being removed as state when removing them from the Workbench (#2727).
- Improved consistency in logging levels in line with our own guidelines (https://github.com/Cratis/.github/blob/main/logging.md).
- Kernel event types that belongs to every event store didn't register automatically when a new event store was added (#2725).
- Showing "Are you sure" dialog when ignoring a recommendation (#1439),
- TLS certificate validation now also accepts localhost name mismatch errors for development environments.
- TLS disabled flag is now correctly propagated to the OAuth token provider.
- Read model pagination now uses 0-based page numbers and guards against negative skip values.
- ReadModels contracts GetInstances and GetAllInstances responses now use IList instead of IEnumerable, preventing null returns.
- IdentityStorage.GetAll now uses ToListAsync instead of ToEnumerable to avoid a dispose bug.

### DIFF
--- a/Source/Kernel/Contracts/Contracts.csproj
+++ b/Source/Kernel/Contracts/Contracts.csproj
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <Project Sdk="Microsoft.NET.Sdk">
     <PropertyGroup>
-        <IsPackable>false</IsPackable>
+        <IsPackable>true</IsPackable>
         <ProduceReferenceAssembly>false</ProduceReferenceAssembly>
         <TargetsForTfmSpecificContentInPackage></TargetsForTfmSpecificContentInPackage>
         <AssemblyName>Cratis.Chronicle.Contracts</AssemblyName>
--- a/Source/Kernel/Directory.Build.props
+++ b/Source/Kernel/Directory.Build.props
@@ -5,8 +5,9 @@
         <!-- Override global multi-targeting for Kernel projects (except Contracts which needs to multi-target) -->
         <TargetFramework Condition="'$(MSBuildProjectName)' != 'Contracts'">net10.0</TargetFramework>
         <TargetFrameworks Condition="'$(MSBuildProjectName)' != 'Contracts'"></TargetFrameworks>
-        <!-- Kernel projects should not be packaged -->
-        <IsPackable>false</IsPackable>
+        <!-- Only Contracts is published as a NuGet package from Kernel -->
+        <IsPackable Condition="'$(MSBuildProjectName)' != 'Contracts'">false</IsPackable>
+        <IsPackable Condition="'$(MSBuildProjectName)' == 'Contracts'">true</IsPackable>
         <InvariantGlobalization>true</InvariantGlobalization>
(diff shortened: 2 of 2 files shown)


einari:
@copilot pull main in and rebase and fix any conflicts

PR #3107, opened 2026-04-16
feat: configurable admin user bootstrap with credentials

### LINKED ISSUE #3106: Configurable Admin user with credentials
As we've done with the applications (cliendId/clientSecret) we want to be able to spin up an instance of Chronicle with an admin user set up. This means having the ability to specify the username and password. As with client credentials, we need to treat this with care and don't keep the credentials in memory. It's only for feeding the database once. If there already is an admin user, we don't attempt this. 

If this is not in the config we revert to the current strategy. 
We should also have the option in the config of forcing a change of the password on first login. 

We need to update the documentation with all of this and point out the security aspects of this

einari:
Allows Chronicle to be configured with an admin user (username, password, email) at startup via configuration or secrets management, similar to how bootstrap clients work.

## Changes
- New `adminUser` section under `authentication` config
- Password is hashed immediately and plaintext is never retained in memory
- `requirePasswordChangeOnFirstLogin` option for production deployments
- Falls back to current strategy when not configured
- Updated documentation with Key Vault, Kubernetes, and Docker Compose examples

Closes #3106

Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/Documentation/hosting/configuration/authentication.md
+++ b/Documentation/hosting/configuration/authentication.md
@@ -16,13 +16,141 @@ Authentication is always enabled. When `authority` is not configured, Chronicle 
 | Property | Type | Default | Description |
 | --- | --- | --- | --- |
 | authority | string | null | External OAuth authority URL |
-| defaultAdminUsername | string | "admin" | Default admin username created on first startup |
+| defaultAdminUsername | string | "admin" | Default admin username created on first startup when `adminUser` is not configured |
+
+## Admin user bootstrap
+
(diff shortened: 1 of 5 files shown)


einari:
@claude we don’t need to say that you shouldn’t do this in production. It’s fine, the user just need to know - and that should source it from a secret store. As for the hashing, docs don’t need to say how we hash it - just that if gets hashed. And for the implementation, it obviously has to be hashed with the same hash we use to authenticate the user.

einari:
@claude The build fails, remember warnings are errors

Issue #3120, opened 2026-04-18
Compliance - PII


einari:
In the .NET client we have a way of describing properties as PII by using an attribute or by using a PIIConcept. 

We have a lot of infrastructure in the backend (Kernel) as well. Just not hooked up entirely as it should. 

What is important is that we add metadata from the client for properties that should be PII in the JSON Schema we create for EventTypes and also ReadModels. 

In the kernel we need to keep a certificate per event source id when persisting an event, if an event type has PII data on it.

We have logic for this in the EventSequence. We just need to make sure it's all hooked up and that we have the specs we need for it. This also mean decrypting when we need events. All of this should be from the grain perspective - Event Sequence. 

Every event source id that gets associated with PII needs a unique certificate that we store in the Certificate Store, implementations for both MongoDB and EntityFramework. 

We will need specs all the way and integration specs in DotNET.InProcess.

PR #3121, opened 2026-04-19
PII compliance: hook up encryption, add Subject concept

einari:
## Added

- New `Subject` concept on both the client (`Cratis.Chronicle.Subject`) and the kernel (`Cratis.Chronicle.Concepts.Events.Subject`) that identifies the target an event is about for compliance concerns. Implicit conversions from `string`, `Guid`, and `EventSourceId` (#3120).
- Optional `Subject` parameter on `IEventLog.Append` / `IEventSequence.Append`, threaded through the gRPC `AppendRequest` and `EventToAppend` contracts into the kernel grain. The grain uses the Subject — falling back to `EventSourceId` when not provided — as the identity key for PII encryption, so PII encryption can be keyed per person even when a single person spans multiple aggregates (#3120).
- SQL / Entity Framework implementation of `IEncryptionKeyStorage`, giving PII compliance feature parity with MongoDB. Previously the SQL backend fell back to in-memory storage, which lost encryption keys on process restart (#3120).
- `Compliance` documentation area covering the `[PII]` attribute, `PIIConceptAs<T>`, per-subject encryption keys and right-to-erasure, plus a `Subject` page under Concepts (#3120).

## Fixed

- Appending an event with a `[PII]`-decorated property no longer fails with `MissingEncryptionKey`. A unique encryption key is now generated and persisted for the subject on first use, so PII events can be encrypted and decrypted end-to-end (#3120).


einari:
@copilot Merge from main and rebase.
Finish your todo list. 

Then, also introduce a SubjectAttribute for the client that can be applied to properties and record properties (parameters)

then we need comprehensive documentation. Do this by creating a folder in the root of the docs for Compliance and a page explaining it and we do specifically with PII

then add a page for what you can do from the client and how practically works

einari:
@copilot Fix the PII and Subject attribute and any other attributes that only work on properties to work on record "properties" so we don't have to use the [property:<attribute>] syntax. Update documentation that is affected by this and anything in the repo that is using the attributes like this.

einari:
@copilot Implement the missing persistence that is described in the Pull Request description. And also the .Decrypt functionality in IReadModels from the client, but let's call it .Release(). Lets make a couple of overloads for this: 

- .Release<TReadModel>(readModelInstance)
- .Release<TReadModel>([readModelInstanceCollection] - IEnumerable<TReadModel>)
- .Release<TReadModel>(Subject subject, readModelInstance)

When using the Release overloads without subject, it needs to derive the subject from the type itself. This means looking for the same things we do on events - Subject attribute, and ultimately fall back to the Id, key of the ReadModel. The backend through gRPC contracts should return a discriminated union of the decrypted version - if error, return an error object. For instance, if the Subject does not exist, it means we can't decrypt it.

We need to log as error that it couldn't be decrypted, both on Kernel side and in the Client. But not being able to decrypt, should just return what was given to it for decryption - its not an Exceptional state that should cause a program halt.

If there are no Compliance related properties on the read model, it shouldn't call the server at all.

Remember specs on this and also documentation.

einari:
@copilot Lets represent Subject using a ConceptAs in the backend (Kernel) and storage as well, not a primitive type like string.

einari:
@copilot Fix the markdown errors - use the markdown verification inside Documentation to see whats going on. Then merge and rebase main into this branch.

PR #3165, opened 2026-05-02
Fix post-release regressions: PII schema, Subject round-trip, and projection deferred key

einari:
## Fixed

- PII compliance metadata now applied correctly to \`ConceptAs<T>\` types adorned with \`[PII]\` in the JSON schema generator — previously the concept path bypassed the compliance metadata check entirely
- \`EventContext.Subject\` defaults to \`EventSourceId\` when deserialising an \`AppendedEvent\` back from a contract that carries no \`Subject\` field, preventing a null-dereference on \`SubjectIsEventSourceId\`
- \`ShouldHaveAppendedEvent\` integration test helper now decrypts PII fields before asserting on event content, matching the architecture where decryption no longer happens in \`EventConverter\` (#3120)
- Restored deferred-future creation in \`TryFindParentByFallback\` so out-of-order child projection events (arriving before their parent is appended) are correctly deferred and re-processed once the parent materialises — the previous change incorrectly used \`UnresolvableKey\` for this legitimate scenario


einari:
@copilot Fix the regression in the for_EventSequence specs (https://github.com/Cratis/Chronicle/actions/runs/25250651015/job/74042253347?pr=3165)

PR #3244, opened 2026-05-13
Fix observer registration deadlock caused by inline replay triggering

einari:
## Fixed

- Fixed a deadlock that caused `ProjectionRegistrationFailed` (and equivalent) timeout errors when registering projections, reducers, reactors, or webhooks whose definitions had changed. `SetDefinition()` previously triggered `ReplayForAllNamespaces()` inline, which called `GetDefinition()` back on the same non-reentrant grain while `SetDefinition()` still held the grain's execution slot — neither call could proceed and Orleans timed out after 30 s.

## Changed

- Observer registration (`SetDefinition()`) is now always fast and non-blocking. Replay/catch-up is scheduled as a separate grain turn via `RegisterGrainTimer(dueTime: Zero, period: Infinite)`, which fires only after `SetDefinition()` returns and the grain is free to serve other calls.


einari:
@copilot merge main into here and rebase

einari:
@copilot There is now a regression in the integration-api : https://github.com/Cratis/Chronicle/actions/runs/25799106112/job/75789440335?pr=3244

einari:
@copilot Merge main in here and rebase

PR #3266, opened 2026-05-15
Fix SQL storage integration specs for PostgreSQL, MSSQL, and SQLite

einari:
## Fixed

- SQL Server integration specs now run: string primary key columns in all SQL migrations were missing a `maxLength`, causing SQL Server to generate `NVARCHAR(MAX)` which is not valid as a primary key. Added `maxLength: 200` to every string PK column across all migrations.
- All SQL storage integration specs (PostgreSQL, MSSQL, SQLite) now start up: `ClusterStorage.GetEventStores()` used a LINQ `Select` that cast an entity string property to `EventStoreName` inside the EF Core query, triggering a crash in the `ConceptAsExpressionRewriter` — the kernel could not initialize and all SQL tests failed. Fixed by projecting to `string` inside the query and converting to the concept in memory.
- `GetInstanceById(ReadModelKey.Unspecified)` no longer returns `null`: the sink-based fast path tried to look up a document with key `"*"`, which never exists, and returned `ReadModel = "null"`. The unspecified key now falls through to `ImmediateProjection` which replays all events and returns the last projected state.


einari:
Closing — fixes pushed directly to the original PR branch #2269 (copilot/enable-integration-specs-different-setups) instead.

PR #3314, opened 2026-05-23
Fix bugs related to enum handling and projection registration

woksin:
# Summary

Fixes three `ReadModelScenario<T>` bugs that surfaced read models as `null` or threw on `_scenario.Instance` for common, documented read-model patterns — enum properties, shared event types across class/property/nested mappings, and identifiers sourced from the event source ID.

## Added

- Spec coverage for enum properties round-tripping through `AsExpandoObject`.
- Spec coverage for projections that register a single event type through class-level `[FromEvent<T>]` + property-level `[SetFromContext<T>]` + a `[Nested]` child's `[SetFrom<T>]` simultaneously.
- Spec coverage for `ReadModelScenario<T>` populating `[Key]` and `[Subject]` identifier properties from the event source ID.

## Fixed

- `ReadModelScenario<T>.Instance` no longer throws `JsonException: Unexpected token StartObject when parsing an Enum` for read models with enum properties.
- `ProjectionFactory.ResolveEventsForProjection` no longer throws `ArgumentException: An item with the same key has already been added. Key: <EventType>+1` when one event type is registered through class-level `[FromEvent<T>]`, property-level `[SetFromContext<T>]`, and a `[Nested]` child's `[SetFrom<T>]` on the same projection.
- `ReadModelScenario<T>.Instance` now populates the read model's identifier property (`[Key]`, `[Subject]`, or conventionally-named `Id`) from the resolved projection key, matching the production MongoDB `_id`-to-identifier behavior.

### DIFF
--- a/Source/Clients/Testing.Specs/ReadModels/ModuleCustomerInfo.cs
+++ b/Source/Clients/Testing.Specs/ReadModels/ModuleCustomerInfo.cs
@@ -1,0 +1,17 @@
+// Copyright (c) Cratis. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Cratis.Chronicle.Projections.ModelBound;
+
+namespace Cratis.Chronicle.Testing.ReadModels;
+
+/// <summary>
+/// Nested record populated from <see cref="ModuleOpened"/> via property-level <see cref="SetFromAttribute{TEvent}"/>.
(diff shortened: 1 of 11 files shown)


einari:
@copilot Fix the compiler error: https://github.com/Cratis/Chronicle/actions/runs/26332171267/job/77519915626?pr=3314

Issue #1557, opened 2024-11-26
Show number of handled events for every observer


einari:
We want to show the number of events every observer has handled in the observers view. 

We can achieve this by counting number of events from the beginning of the event sequence to the last handled event sequence number.

As an optimization, we can have a Grain that holds this information in-memory and updates the numbers when asked. 
We can also do this on a timer on a regular cadence and update the in-memory stats. This could be better, since it could take some time to get all the numbers for a system with a lot of observers and events.

PR #3335, opened 2026-05-27
Add handled event count per observer

woksin:
## Added

- `HandledEventCount` and `HandledEventCountPerEventType` (per-event-type breakdown) to `ObserverState` — counts are incremented directly as live events are successfully handled (#1557)
- `HandledEventCountPerPartition` to `ObserverState` — tracks per-partition, per-event-type counts so that partition replay correctly adjusts aggregates (#1557)
- `HandledEventCount` on `ObserverInformation` (contract + TS proxy) — the number of events the observer has handled (#1557)
- "Handled events" column on the workbench observers table (#1557)
- `IEventComplianceHelper` interface — extracts shared event compliance/decryption logic into an injectable, testable service (#1557)

## Changed

- Observer service reads handled event counts directly from `ObserverState` in `GetObservers` / `ObserveObservers` / `GetObserverInformation` responses (#1557)
- Handled event counts are reset when a full observer replay starts, so counts reflect events handled since the last replay (#1557)
- When a partition replay starts, only that partition's contribution is subtracted from the aggregate and per-event-type counts, leaving other partitions' counts intact (#1557)
- `HandleEventsForPartition` now increments handled event counts during replay and catchup, so counts reflect all events handled regardless of how they were processed (#1557)
- `EventComplianceHelper` converted from a static helper to an injectable class implementing `IEventComplianceHelper`, with `IJsonComplianceManager` and `IExpandoObjectConverter` injected via constructor (#1557)
- `Observer` and `HandleEventsForPartition` now take `IEventComplianceHelper` instead of `IJsonComplianceManager` + `IExpandoObjectConverter` directly (#1557)
- `ConceptDictionaryJsonConverterFactory` now handles `IReadOnlyDictionary<,>` with concept keys in addition to `IDictionary<,>` and `Dictionary<,>` (#1557)
- `KeyJsonConverter` now supports `WriteAsPropertyName`/`ReadAsPropertyName` so `Key` can be used as a JSON object property name in dictionary serialization (#1557)

## Fixed

- JSON serialization of `IReadOnlyDictionary<EventTypeId, EventCount>` and `IReadOnlyDictionary<Key, IReadOnlyDictionary<EventTypeId, EventCount>>` in `ObserverState` — previously caused `InvalidOperationException` during Orleans grain state serialization, breaking integration specs (#1557)

## Removed

- `ObserverHandledEventCounts` Orleans grain — event counts are now tracked incrementally in state rather than computed on demand via a cached grain (#1557)


einari:
@copilot Lets improve on this, rather than having a grain that does this. This is state we actually can produce while handling events by just incrementing a number for each handled event. Also, in addition to doing an aggregate number for all handled events, we want to keep a dictionary per event type in the observer state that we also count per event type. This would simplify everything and not add unnecessary load on the Kernel.

einari:
@copilot We also need to take into consideration replay. When we do a replay of an observer, the number of handled events should be reset to 0 - both the aggregate and per event type. This means that we probably need to handle this in the HandleEventsForPartition as well. This gets a little bit more tricky when we are replaying partitions. When we do that, we need to subtract the amount of events for the partition from the values, also per event type - which means we might need additional support for this in storage.

While doing this. Do an assessment of duplication of logic. I have a sneaky suspicion we are now doing a lot of the same handling logic in 2 places (Observer.Handle & HandleEventsForPartition). That is no good. This could be improved by having a system we take dependency to that does this - be wary of nuances.

einari:
@copilot Instead of having the EventComplianceHelper as a static helper - lets actually make it something we can inject - extract it as an interface. A couple of the things could just be injected into it. This will create less friction and make it more testable. Add specs for the new implementation in isolation.

einari:
@copilot Specs and integration specs failing big time - see the builds for this PR

woksin:
@copilot+claude-sonnet-4.6 Fix build and test issues

PR #3346, opened 2026-06-01
Fix projection parent key and replay handling

woksin:
# Release notes

## Fixed

- Fixed `AppendMany` subject propagation for compliance and PII encryption.
  - `AppendMany(EventSourceId, IEnumerable<object>)` now resolves `[Subject]` from each event and sends it per event.
  - `AppendMany(IEnumerable<EventForEventSourceId>)` now preserves an explicit per-event `Subject` override, or derives it from `[Subject]` when no override is provided.
  - The gRPC contract, kernel append-many model, compliance application, SQL storage, in-memory storage, and the API append-many endpoint now carry the subject through consistently.
  - Batched events whose PII belongs to a person different from the event source are now encrypted under the person subject instead of falling back to the aggregate event source id.
- Fixed child projection parent-key resolution when the parent projection is identified by `$eventSourceId` and the child key is a different type.
- Fixed projection sink updates that combine parent collection property changes with child add/remove operations in the same update across MongoDB, SQL, and in-memory storage.
- Fixed repeated MongoDB redaction handling so already-redacted events preserve the original event type and do not trigger another observer rewind.
- Fixed redaction detection to compare against the event type id instead of the full event type value.
- Fixed compliance release handling for non-string subjects and multiple PII properties.
- Fixed observer event decryption when event type schemas with PII metadata are loaded lazily while handling events.

## Changed

- Registered the required Chronicle observation activity sources for reactor and reducer gRPC services.
- Extended append-many client/test harness support with per-event subject assertions and an in-process PII integration scenario.


einari:
@woksin In the PR #2269 we have an improvement similar to what you did for the build of the image - but we do it in the pipeline. So I think if we could hold off on this change, to avoid conflicts. Close to merging the #2269 one .

Issue #3381, opened 2026-06-10
Reactor side-effects needs to honor append result and fail if it fails


einari:
If an append/append many fails, it should then consider the reactor as failed and then end up as a failed partition with all the details - this will have to produce an exception, a specific custom exception.

Issue #3359, opened 2026-06-05
React to read model changes.


einari:
We want to be able to react to read models a they get added, modified or removed. 

The whole thing should be by convention with the method name telling what type of change you're looking for. 

The signature should then need either a single read models or a collection of read models. But allow for having other dependencies dynamically, where it can take event context as well as other dependencies. Much like reactors as described in #3358.

We also want to support side effects, like reactors. Allowing to return events in the same way as we for reactors and model bound commands. Like we do with Cratis Arc commands, it should be allowed to return tuples that includes the EventSourceId, Subject, EventStreamId. And also support the use of EventStreamType and EventSourceType attributes to set values that could be included when appending.

The signatures should support sync and asynchronous - with and without Task.

```csharp
public class MyReadModelWatcher : IReadModelWatcher
{
    public Task Added(SomeReadModel someReadModel)
    {
       // Do stuff
    }

    public Task Modified(SomeReadModel someReadModel)
    {
       // Do stuff
    }

    public Task Removed(SomeReadModel someReadModel)
    {
       // Do stuff
    }
}
```

This is a convenience layer on top of the IReadModels Watch APIs.

We need to track subscriptions and clean up on disposal of the client. 

Also, add an attribute called Materialized. When a watcher has this, it will use the Materialized API for watching. 

Important, we would typically use the multi instance watching APIs and then turn around and call each of the watchers individually. 

We probably need more info on the watcher to know which event context caused the changes. For materialized we would need to deduct that from the sequence number of the event.

woksin:
Done — this is `IReadModelReactor`. Methods are discovered by name (`Added`/`Modified`/`Removed` → `ReadModelChangeType`) in `ReadModelReactorMethods`, the watched read model type is derived from the first parameter (a single read model or an `IEnumerable<T>` of one), and `ReadModelReactorInvoker` resolves dependencies, supports sync and async, and handles side effects.

Worth noting the reliability follow-up in #3418 (Added vs Modified for reducer-backed and replayed changes) is a separate, still-open concern.

Issue #3358, opened 2026-06-05
Ability to take dependencies to read models in reactors


einari:
Reactors should support signatures that takes dependencies in general on methods but specifically for read models we should handle this. A read model is defined as something there is a reducer or a projection (declarative or model bound). 

For a regular IReactor - a flexible signature that allows for taking a dependency on the methods to read models, in addition to the event it reacts to.

```csharp
public class MyReactor : IReactor
{
   public Task On(SomethingHappened @event, EventContext context, SomeReadModel readModel)
   {
       // Do stuff... 
   }
}
```

The ReadModel would then be materialized using the reducer or the projection it has for it - making it strongly consistent.

In addition to this we need an interface in the reactors namespace that allows one to resolve the Read Model key to use. 'ICanResolveReadModelKey' - should have a resolve method that takes an event as object and the EventContext. The event should be the deserialized version. 

If this interface is there, that is the key for resolving. If not used, we'll assume the EventSourceId from the event context. 

We don't need to use the serviceprovider for read models, we have full control over that and can resolve it directly. 

But for other types you want to have as dependency we do need to resolve them from the service provider.

Look at how we've done this dynamically in Arc and the command handlers:

https://github.com/Cratis/Arc/tree/main/Source/DotNET/Arc.Core/Commands

And specifically the management around arguments and dependencies:

https://github.com/Cratis/Arc/blob/main/Source/DotNET/Arc.Core/Commands/ModelBound/ModelBoundCommandHandler.cs

woksin:
Done. `ReactorMethodArgumentsResolver.Resolve` fixes argument 0 to the event and resolves the rest through `ResolveArgument`: an `EventContext` parameter gets the context, a read model parameter is materialized directly by Chronicle rather than through the service provider, and anything else falls back to `serviceProvider.GetRequiredService`, throwing `CannotResolveReactorMethodArgument` when it cannot.

Where the read model key differs from the triggering event source, `ICanResolveReadModelKey` on the reactor supplies it.

PR #3402, opened 2026-06-16
React to read model changes with IReadModelReactor

woksin:
# Summary

An `IReadModelReactor` reacts to a read model being **added, modified or removed** using convention-based handler methods — a convenience layer over the Watch APIs so you write a class instead of subscribing to an observable.

One nuance worth knowing: projection-backed read models get a precise, server-provided change type and the causing event's context; reducer-backed and `[Materialized]` read models infer the change type client-side (first sighting of a key → `Added`, thereafter `Modified`) and carry only the model key. Read-model-reactor side effects are best-effort (a read model reactor has no partition to fail), so design handlers to be idempotent.

> Scope note: this PR previously sat at the top of a reactor stack. The reactor side-effect-append work (#3381) and reactor handler dependencies (#3358) shipped separately via **#3383** and **#3401** and are already in `main`. This PR is now scoped to the **read-model reactor** only.

## Added

- React to read model changes with `IReadModelReactor`: methods named `Added`, `Modified` or `Removed` are invoked by convention with the changed read model (a single instance or a collection), an optional `EventContext`, and services, and may return events as side effects. (#3359)
- `[Materialized]` opts a read model into materialized change tracking, deducing the change type by comparing successive materialized windows. (#3359)

## Changed

- Read model change notifications now carry the change type and the causing event's sequence number, occurred time and correlation id. (#3359)


woksin:
### Reacting to reducer-backed read models — fidelity concern (cc @einari)

This PR supports read-model reactors over **reducer-backed** read models as well as projection-backed ones. Flagging a semantic gap in the reducer path before we lean on it.

In `ReadModelReactors.Subscribe<TReadModel>`, projection-backed models get the precise change type (and the causing event's context) from the server, but reducer-backed models infer `Added` vs `Modified` from an in-memory set:

```csharp
// Projections report the change type from the server; reducers compute changesets locally without one,
// so first-seen keys are tracked client-side to distinguish an addition from a modification.
var seenKeys = eventStore.Reducers.HasFor<TReadModel>()
    ? new HashSet<string>(StringComparer.Ordinal)
    : null;
```

`seenKeys` is created fresh on every `Start()` — **including reconnects, which are routine** — and `Watch` is a live stream (no replay on subscribe). So after a reconnect, the **first change to a pre-existing instance is reported as `Added`** instead of `Modified`. Combined with read-model-reactor side effects being best-effort / fire-and-forget (a read model reactor "has no partition to fail"), the canonical *"do X once when an instance is created"* use case becomes unsafe for reducer-backed models — e.g. a welcome-email reactor would re-fire for existing instances after a reconnect.

`Removed` is reliable (server flag) and **projection-backed read models are unaffected** (exact change type from the server).

**Options:**
1. **Make it reliable** — have the sink surface insert-vs-update for reducer-materialized writes (it already tracks this via `__lastHandledEventSequenceNumber` / upsert) so the change type doesn't depend on client memory. Then reducers are as trustworthy as projections.
2. **Gate / warn** — opt-in or a startup warning for reducer-backed read-model reactors, beyond the doc note in `reacting-to-changes.md`.
3. **Restrict to projections** — fail fast at registration for reducer-backed models and point reducer users at **event** reactors (full context + `[OnceOnly]` for exactly-once), which is the canonical answer for a reducer's inputs anyway.

@einari — do we want reducer-backed read-model reactors at all, and if so which direction do you prefer? Happy to implement whichever.

einari:
@woksin Do you mind rebasing this on top of origin/main and fixing conflicts and then make sure there are no regressions.

woksin:
Folded into the consolidated non-Workbench PR #3507 (its `readmodels.proto` change was union-merged with main's `ReadModelChangeset`, and the `ReadModelWatcher` reconnect + reactor-body changes were combined). Closing in favor of #3507.

PR #3422, opened 2026-06-22
Fix RemovedWith with empty events using implicit EventSourceId resolution

einari:
## Fixed

- Fix DeclarativeCodeGenerator to properly handle well-known expressions in RemovedWith code generation


einari:
@copilot Fix the build errors

PR #3448, opened 2026-07-03
Trigger publish for Docker-only and publish-workflow changes

woksin:
# Summary
The Publish workflow triggered only on `Source/**` changes, so a Docker-only fix — like #3446 restoring the server apphost's executable bit — could not cut a release by itself. This adds `Docker/**` and the publish workflow file to the trigger paths so image and pipeline fixes can publish going forward.

_Internal CI change — no user-facing impact; nothing for the changelog._

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -38,6 +38,11 @@ on:
       - "**"
     paths:
       - "Source/**"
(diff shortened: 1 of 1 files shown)


einari:
@woksin Is this good to go?

Issue #3463, opened 2026-07-10
Non-passive model-bound [ChildrenFrom] child-collection projection does not materialize in the out-of-process kernel


woksin:
## Summary

A **non-passive**, **model-bound** read model with a `[ChildrenFrom<TEvent>]` child collection reliably fails to materialize its children when the projection runs in the **out-of-process** kernel (as used by the integration test matrix). `GetInstanceById<T>` returns the parent read model, but its child-collection property never reaches the expected count within the test timeout (150s), so the child collection appears empty/never-populated.

This surfaced while adding an end-to-end integration spec for a subject owning two PII read models (PR #3462). The child-collection variant timed out only in CI (out-of-process); the same test passed in-process and against a stock published kernel image locally, which is why it was hard to catch.

## What works vs. what fails

- ✅ **In-process** runtime: the child collection materializes and (when present) child `[PII]` releases correctly.
- ✅ **Out-of-process against a stock `*-development` image locally**: materializes.
- ❌ **Out-of-process in CI** (branch-built `-development` slim image, MongoDB backend): the child collection never materializes — the poll on `GetInstanceById<T>().Children` times out at 150s. A **simple** (non-child-collection) fluent projection in the exact same run materializes fine, so it is specific to the non-passive model-bound child-collection shape.
- Note: the existing passing child-collection integration test (`for_ReadModels/when_getting_instance_for_passive_projection/and_child_element_is_updated`) uses a **`[Passive]`** read model (materialized on read), so it does not exercise the async, persisted projection path that fails here.

No kernel-side exception, error, or observer quarantine was observed in the container logs — the projection simply does not produce the child rows.

## Minimal repro

Add to the client integration suite (out-of-process, MongoDB) and observe the poll time out:

```csharp
[EventType] public record ContractCreated(string Name);
[EventType] public record CandidateSubmitted(Guid CandidateId, string Name);

[FromEvent<ContractCreated>]
public record Contract(
    string Name,
    [ChildrenFrom<CandidateSubmitted>(key: nameof(CandidateSubmitted.CandidateId), identifiedBy: nameof(Candidate.CandidateId))]
    IReadOnlyList<Candidate> Candidates);

[FromEvent<CandidateSubmitted>]
public record Candidate([Key] Guid CandidateId, string Name);

// context: EventTypes => [ContractCreated, CandidateSubmitted]; ModelBoundProjections => [Contract]
await EventStore.EventLog.Append(id, new ContractCreated("MSA"));
[...]

PR #3464, opened 2026-07-10
Fix PII active projections stalling on the SQL sink (#3463)

woksin:
## Fixed

- Active (non-passive) read models with `[PII]` persisted to a SQL sink (PostgreSQL, SQL Server, SQLite) stalled and never materialized out of process — the query returned `null` indefinitely with no error or observer quarantine, and PII could fail to decrypt on read. MongoDB-backed read models were unaffected. (#3463)

### DIFF
--- a/Integration/Client/ChronicleConfigurableFixture.cs
+++ b/Integration/Client/ChronicleConfigurableFixture.cs
@@ -338,6 +338,31 @@ public class ChronicleConfigurableFixture : XUnit.Integration.ChronicleFixture
         ? _outOfProcessKernelContainer.GetMappedPublicPort(35000)
         : throw new InvalidOperationException("KernelGrpcHostPort is only valid in OutOfProcess mode after the kernel container has been built.");
 
+    /// <summary>
+    /// Gets the captured stdout/stderr of the out-of-process kernel container, or a short
+    /// placeholder when running in-process. Used by specs to surface kernel-side diagnostics
+    /// (observer subscription, routing, projection traces) into the test output when an
+    /// eventual-consistency poll times out in CI, where the container logs are otherwise lost.
+    /// </summary>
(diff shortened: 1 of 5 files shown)


einari:
@woksin Is this good to go?

woksin:
Not yet I think

woksin:
Update — this is no longer just a diagnostic PR. It now carries the actual fix for #3463 plus the out-of-process regression spec:

- **The fix** — active (non-passive) `[PII]` read models on a SQL sink (PostgreSQL, SQL Server, SQLite) stalled and never materialized out of process (`null` forever, no error, no quarantine). Root cause was three schema-round-trip drops in the compliance path (document key dropped on encrypt → nulled SQL primary key; `__subject` never persisted → double-encryption on the next event). Fixed in `ReadModelsCompliance.Apply`, `ProjectedColumns` (now emits a `__subject` column), and `Sink` (re-attaches the subject on materialize).
- **The spec** — drives materialization + PII release end to end for a subject owning a root-PII model alongside a sibling model-bound `[ChildrenFrom]` collection whose children carry `[PII]`.

I've updated the description to a proper release note (dropped the stale "diagnostic / not for merge" framing).

CI on the fix commit: **94 pass / 21 skipped / 1 fail**. The one red check is `coverage-merge-and-cache`, which is pre-existing plumbing unrelated to this change — the coverage-commit step runs `git pull --rebase origin <PR#>/merge`, and `<PR#>/merge` isn't a pushable branch ref on `pull_request` runs (`fatal: couldn't find remote ref …/merge`). It fails identically on other PRs (e.g. #3466) whenever coverage stats change, and it's not a required gate, so merge state stays `MERGEABLE`.

So: good to go from a CI + description standpoint. The coverage-merge workflow step should be fixed separately (it should only push on branch pushes, not `pull_request` runs) so it stops going red across PRs.

PR #3465, opened 2026-07-10
Upgrade TypeScript to 7

woksin:
## Changed

- Upgrade to TypeScript 7.0.2 (the native compiler) for faster build/typecheck.

## Package version changes (typescript)

| Package | File | Old | New |
|---|---|---|---|
| `@cratis/chronicle.contracts` | `Source/Clients/TypeScript/package.json` | `^6.0.2` | `^7.0.2` |
| `workbench` | `Source/Workbench/package.json` | `6.0.3` | `7.0.2` |

Range prefixes preserved. No `@types/*` or peer-dependency lower bounds were touched.

## Package manager

Both TypeScript-declaring packages use **Yarn**.

- `Source/Workbench` uses Yarn Berry with a committed `yarn.lock` — regenerated. Its **Yarn release was bumped `4.5.3 → 4.17.1`** (and `packageManager`/`yarnPath` updated) because Yarn 4.5.3's bundled `plugin-compat` unconditionally applies a builtin editor-SDK patch to the `typescript` package that targets `lib/_tsc.js` / `lib/tsserver.js`. The TS7 native package ships a different `lib/` layout (only a small `tsc.js` shim to the Go binary), so the patch fails with `ENOENT` and install cannot complete. Yarn 4.17.1's compat plugin handles the native layout, so this bump is required for the upgrade. `nodeLinker` is `node-modules`, so the patch is not needed at runtime.
- `Source/Clients/TypeScript` uses Yarn classic (its `yarn.lock` is gitignored) — no committed lockfile to update.

## Config migration for TS7

`Source/Workbench/.frontend/tsconfig.json`: TS7 removed `compilerOptions.baseUrl`. Removed it and rewrote every `paths` entry to be relative to the config file (prefixed with `../`, since `baseUrl` was `..`). Added an explicit `"Features": ["../Features/index"]` mapping that previously resolved implicitly through `baseUrl`.

## Verification

- `typescript` resolves to **7.0.2** in both packages (`require('typescript/package.json').version`); `tsc --version` → `Version 7.0.2` (native).
- **Workbench** (the main TS package): `tsc -b .frontend/tsconfig.json` and `tsc -p .frontend/tsconfig.specs.json` both exit `0` — this is the TS portion of the `yarn ci` gate.
- **Clients/TypeScript**: `tsc` runs under 7.0.2; a full `yarn build` there could not be exercised locally because it needs `protoc` codegen (`yarn generate`), which isn't installed in this environment — this is unrelated to the TS7 bump.

## Workflows

Grepped `.github/workflows/`: no explicit `typescript` version pins or direct `tsc`/typescript install steps. The TS jobs (`typescript-build.yml`, `workbench-build.yml`, `publish*.yml`) only run `yarn install` / `yarn build` / `yarn ci`, so **no workflow changes were needed**.

## Notes

- `@typescript-eslint/*` (8.59.1) declares a `typescript` peer range of `>=4.9.0 <6.1.0`, so Yarn emits a non-fatal peer-dependency warning under TS7. This is an upstream tooling range, not a compile error; the typecheck passes.


einari:
@woksin Could you have a look here and get it to pass so that we can get it merged?

PR #3478, opened 2026-07-16
Fix PII encryption size cap, root-level join partition freeze, and reducer GetInstanceById

woksin:
# Summary

Three independent runtime hardening fixes — for compliance encryption, root-level projection joins, and the client read-model API. Each addresses a case that previously threw or permanently froze an event-source partition. Also included is a fidelity fix to the in-process `EventScenario` testing helper so it no longer masks a rejected append in a multi-event action.

Known follow-ups for the root-level join work, intentionally not in this PR: fanning a shared join value out to *every* matching root (the current single-matching-root update), and eliminating the in-memory instance that the replay-based `GetInstances` — as opposed to materialized reads — still builds for an orphan join source.

## Changed

- `[PII]` values are now protected with hybrid AES-256-GCM envelope encryption (a fresh per-value data key wrapped by the per-subject RSA key) instead of raw RSA. Values written by earlier versions continue to decrypt transparently.

## Fixed

- `[PII]` values longer than ~245 bytes (long free-text fields, images) no longer throw a `CryptographicException` at append time — arbitrary-length PII now encrypts and decrypts.
- A root-level `[Join]` whose source stream key type differs from the read model's key (for example a string organization number joined onto a `Guid`-keyed root) no longer permanently freezes the observer partition with an "Unrecognized Guid format" error, and no longer materializes a phantom root document keyed by the join value; the join backfills when the real entity row appears.
- `IReadModels.GetInstanceById` now returns `null` for an unseeded reducer-backed read model instead of throwing, matching projection-backed read models — so a reactor's nullable pre-check no longer throws and freezes the partition.
- `EventScenario` (in `Cratis.Chronicle.Testing`) now short-circuits a multi-event `Events(...)` append on the first rejected event, returning that failed `AppendResult` instead of continuing and masking the rejection with a later event's success — so specs asserting a constraint or concurrency violation on a multi-event action report the failure, and later events no longer land.


woksin:
@einari — flagging the compliance-encryption change in this PR (commit `174cd730f`) for your eyes, since it changes how `[PII]` is encrypted at rest.

### What changed
`Source/Kernel/Core/Compliance/Encryption.cs` moves from **raw RSA-2048 over the whole value** to **hybrid (envelope) encryption**: the value is encrypted with a fresh **AES-256-GCM** data key, and only that 32-byte data key is RSA-wrapped. It's stored as a self-describing envelope — `CENV` magic + version + RSA-wrapped key + nonce + GCM tag + ciphertext.

### Why we switched
Raw RSA-2048 with PKCS#1 v1.5 caps plaintext at **~245 bytes** (256-byte modulus − 11 bytes padding). Any `[PII]` value longer than that threw `CryptographicException` ("data too large for key size") **at append time** — so `[PII]` was unusable for images and for any realistic free-text field (2000-char validators, bios, feedback, etc.); every such value 500'd the moment a real value was entered. RSA simply isn't meant to encrypt arbitrary-length data. Hybrid encryption removes the cap entirely (AES has no size limit) while keeping the per-subject RSA key as the root of trust, and gets authenticated encryption (the GCM tag) for free.

### Why this PR exists
It bundles three independent hardening fixes surfaced on a live run (Chronicle 16.1.1), each of which could throw or **permanently freeze an event-source partition**:
1. this PII encryption size cap;
2. a root-level `[Join]` freezing the observer on a differently-typed join value (a string org number joined onto a `Guid`-keyed root → "Unrecognized Guid format");
3. `IReadModels.GetInstanceById` throwing (instead of returning `null`) for an unseeded reducer-backed read model, which froze a reactor's nullable pre-check.

### Backward compatibility (verified)
- The `Compliance.IEncryption` interface is **unchanged**; only the implementation changed. The separate `Security.IEncryption` (webhooks) is untouched.
- `Decrypt` recognizes the `CENV` header and falls back to the **original raw-RSA path** for values written by earlier versions, so previously stored PII stays readable — **no data migration** (spec: `when_decrypting_a_legacy_raw_rsa_value`).
- `GenerateKey` still produces RSA keys, so existing per-subject keys keep working.
- **GDPR crypto-shredding is preserved**: only the AES data key is RSA-wrapped, so deleting a subject's RSA key still makes the ciphertext permanently unrecoverable.
- The only non-back-compat direction is the usual one — a *downgrade* to a pre-envelope build could not read new envelope values; not a concern for the normal upgrade path.

Happy to pull this commit into its own PR if you'd rather review the encryption change in isolation.

woksin:
Follow-up hardening on the encryption commit (`a0274967f`), given the security/GDPR sensitivity of this path:

- The AES data key is now wrapped with **RSA-OAEP (SHA-256)** rather than PKCS#1 v1.5 — modern padding for the new envelope format, and a smaller/random padding-oracle target.
- Switched from `RSACryptoServiceProvider` to **`RSA.Create()`** — the recommended, platform-appropriate RSA API.
- Documented the per-value fresh-AES-key invariant that makes AES-GCM `(key, nonce)` reuse impossible, so it can't be refactored away.

All backward-compatible and GDPR-preserving:
- The envelope format isn't in production yet, and legacy raw-RSA values are still read with PKCS#1 v1.5 (interoperable across providers — covered by `when_decrypting_a_legacy_raw_rsa_value`).
- **Crypto-shredding is unchanged**: the per-subject RSA key still gates the wrapped AES key, so deleting it makes the value permanently unrecoverable.

Rebased cleanly onto the branch after `main` was merged in; compliance suite green (99 specs, incl. the composite-key compliance-identifier work).

woksin:
### ⚠️ Operational note — the PII-at-rest format change is forward-only

Calling this out on its own (it was only a one-liner in my earlier comment) because it matters for release notes / the runbook:

- **Upgrade (old → new): safe.** The new build reads *both* formats — legacy raw-RSA (via header detection) and the new AES-GCM envelope.
- **Rollback (new → old): NOT safe once any PII has been written by the new build.** The old build cannot read envelope values — it runs raw RSA over the envelope bytes and fails. PII written *before* the upgrade stays readable; PII written *during* the new-code window becomes undecryptable if you downgrade the code.
- **Mixed-version clusters:** same reason — a still-old silo reading PII a new silo just encrypted will fail. Prefer draining/upgrading over long-lived mixed versions.

Practically, treat the deploy as a **one-way door** for the PII-at-rest format. There is no in-place downgrade of the ciphertext (it's encrypted — you can't re-encrypt it into the old format without the plaintext), so a rollback would have to tolerate or re-derive any PII written in the interim. This is inherent to any at-rest format change; flagging it so nobody discovers it during an incident.

(To make rollback safer in a future change, we could gate the envelope format behind a config flag defaulted off for one release, so a rollback window exists before any envelope value is written — happy to add that if you want a belt-and-suspenders rollout.)

woksin:
### Deployment & operations — checklist for the PII encryption change

Nothing is *required* to make it work — it's a transparent drop-in (same `IEncryption` interface, same key storage, no new config, no data migration). The items to plan for:

**Rollout — forward-only (see the rollback note above)**
- Upgrade old → new is safe: the new build reads both formats. Rollback new → old is not, once any envelope-format PII has been written.
- Clusters: **drain-and-upgrade**; avoid long-lived mixed versions — an old silo can't read PII a new silo just encrypted.
- No data migration, and no re-encryption of existing PII (not needed, and not possible without the plaintext — legacy raw-RSA values are simply read via the fallback forever).

**Keys — unchanged**
- Existing per-subject RSA keys keep working; no rotation needed. They remain the single root of decryptability and the GDPR-erasure lever — **crypto-shredding is unchanged**, so back them up with the same rigor as today.

**Capacity — only relevant if large PII is in play**
- Encrypted values now grow with the plaintext (long text / images encrypt where they previously threw). Mind MongoDB's **16 MB per-document limit** for genuinely large *binary* PII, and keep an eye on storage on PII-heavy read-model / event collections.

**Crypto policy / FIPS**
- AES-256-GCM and RSA-OAEP-SHA256 are FIPS-approved, and the switch to `RSA.Create()` (from `RSACryptoServiceProvider`) is friendlier under a FIPS OS policy. Validate once if you deploy into a FIPS-enforced environment.

**Documentation / audit**
- Update any DPIA, security description, or SOC2 answer: PII at rest is now **AES-256-GCM with an RSA-2048-OAEP-SHA256-wrapped data key** (previously raw RSA-2048 PKCS#1 v1.5).

einari:
I’ve reviewed and it all looks sound. My biggest concern was backwards compatibility and that seems to be in tact. If we were to break backwards compatibility we would’ve needed a migration thing that rotated keys and fixed it all. But this looks sound.

PR #3480, opened 2026-07-19
Add spec for SetFrom-bridged child collection projection

woksin:
# Summary

Adds a model-bound projection **builder spec** proving that a `[SetFrom<TEvent>(nameof(TEvent.OtherName))]` on a `[ChildrenFrom<E>]` child-collection property bridges a differently-named event **list** property into the child collection — no rename required. This closes a coverage gap: the existing analyzer specs only asserted the `CHR0030` (`ChildrenCollectionPropertyAutoMapsToNothing`) diagnostic, not that the bridged name actually threads through to the projection definition.

The spec asserts `childrenDef.From[TicketOpened].Properties["Comments"] == "Annotations"` (the bridge is wired) while the same-named `Title` stays AutoMapped.

Test-only — no user-facing or behavioral changes, so there are no release-note entries.


woksin:
Folded into the consolidated non-Workbench PR #3507. Closing in favor of #3507.

PR #3482, opened 2026-07-19
Regenerate stale Chronicle protobuf contracts

einari:
## Fixed

- The .proto files under Source/Kernel/Protobuf were last regenerated before several C# contract changes landed (ConnectRequest's ProcessId/ProcessPath/MachineName/ClientType, ConnectedClient's matching fields, and other unrelated drift). None of the client SDKs that generate code from these files could see the new fields until this ran. Regenerated via generate-protos.sh with no manual edits.

### DIFF
--- a/Source/Kernel/Protobuf/clients.proto
+++ b/Source/Kernel/Protobuf/clients.proto
@@ -6,6 +6,21 @@ message ConnectRequest {
    string ConnectionId = 1;
    string ClientVersion = 2;
    bool IsRunningWithDebugger = 3;
+   int32 ProcessId = 4;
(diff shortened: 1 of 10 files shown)


einari:
Rewrote this branch — it originally also contained connection-layer changes for the TypeScript/Kotlin/Elixir clients, but those belong in the dedicated \`Chronicle.TypeScript\`/\`Chronicle.Kotlin\`/\`Chronicle.Elixir\` repos (this repo's \`Source/Clients/*\` packages are contracts-only by design), not here. Dropped those commits; this PR is now just the protobuf regeneration.

Issue #3430, opened 2026-06-26
Constraint index does not need the actual data


einari:
To avoid putting PII data in the Constraint indexes, we should instead use the SHA of the string representation of the finished concatenated version.

This means we need to have the SHA of the unencrypted version of the property or properties the constraint is pointing at and then get the SHA for the finished concatenated value.

woksin:
Already the case. `UniqueConstraintDefinitionExtensions.GetValue` joins the property values, lower-cases them when `ignoreCasing` is set, and returns `ComputeSha256Hash(...)` — the index stores a lowercase hex SHA-256, never the value itself. The XML remarks state the intent verbatim: "hashed using SHA-256 to avoid storing PII data in the constraint indexes".

Consumers of the hashed value are `UniqueConstraintValidator` and `UniqueConstraintIndexUpdater`.

Issue #3400, opened 2026-06-16
Aspire: slim-image WithMongoDB silently requires a MongoDB replica set + directConnection — document and/or provide a provisioning helper


woksin:
## Summary

When using the **`Cratis.Chronicle.Aspire`** package with the *slim* image — i.e. `AddCratisChronicle(configure: c => c.WithMongoDB(mongo))` + an external MongoDB — there is no indication that the external MongoDB **must be a replica set** (and, for a single-node RS reached via a host-mapped port, that clients must connect with `directConnection=true`). The `WithMongoDB(IResourceBuilder<IResourceWithConnectionString>)` extension simply sets `Cratis__Chronicle__Storage__{Type,ConnectionDetails}` and leaves replica-set provisioning entirely to the caller, with nothing in the XML docs or the API surface to flag the requirement.

This is a quiet DX cliff: the embedded development image (`AddCratisChronicle()` with no `configure`) bundles a working replica set, so everything "just works". The moment you switch to the slim image + external Mongo (e.g. provisioned via Aspire's `AddMongoDB`, which runs `mongod` standalone by default), you silently lose that guarantee:

- Chronicle uses MongoDB transactions, and observers/projections — plus any Cratis Arc observable queries layered on top — rely on **change streams**. Both require a replica set.
- On a standalone `mongod` the change-stream watch never opens, so observable read-model queries return their seeded-empty value and look like a projection bug rather than a storage-topology problem.
- Any attempt to make Mongo a replica set while still using Aspire's `AddMongoDB` keeps that integration's `mongo_check` health check, whose connection string lacks `directConnection=true`; the driver then does topology discovery against the advertised member host (`localhost:27017`, only meaningful inside the container) and hangs (`OperationCanceledException` from `ListDatabaseNamesAsync`), so the resource never goes healthy and `WaitFor(mongo)` blocks.

## The knowledge already exists in this repo — just not on the Aspire surface

Chronicle's own integration harness already encodes exactly the correct recipe, on Testcontainers:

- `Source/Clients/XUnit.Integration/ChronicleInProcessFixture.cs` starts Mongo as a **self-initiating single-node replica set**:
  > `// Start mongod as a single-node replica set and initialize it so tests can use transactions.`
  ```
  .WithCommand("/bin/sh", "-c",
    "mongod --replSet rs0 --bind_ip_all > /proc/1/fd/1 2>/proc/1/fd/2 & " +
[...]

Issue #3168, opened 2026-05-02
Performance: blocking async in AppendedEventSerializer (Orleans serialization hot path)


einari:
## Summary

`AppendedEventSerializer` is the Orleans `IGeneralizedCodec` for `AppendedEvent`. It is called synchronously on every grain message that carries an `AppendedEvent`. Both the read and write paths block the calling thread with `.GetAwaiter().GetResult()`:

- `ReadValue` line 57: `JsonSchema.FromJsonAsync(...).GetAwaiter().GetResult()`
- `WriteField` line 68: `eventStore.EventTypes.GetFor(...).GetAwaiter().GetResult()`

**File:** `Source/Kernel/Core/Setup/Serialization/AppendedEventSerializer.cs`

## Impact

Orleans serialization runs on the thread-pool scheduler. Blocking thread-pool threads here reduces the concurrency available to all grain activations. Under high message volume this causes thread-pool exhaustion — leading to timeouts and cascading failures across the entire silo.

## Fix

The `IGeneralizedCodec` interface is synchronous so the calls cannot be made `async` directly. Options:

1. Cache schema lookups in a `ConcurrentDictionary` keyed by schema hash / event type id so subsequent calls are non-blocking after the first resolution.
2. Pre-warm the cache during startup / event type registration before message traffic arrives.
3. Evaluate whether the schema needs to travel with the serialized bytes at all, or whether it can be resolved from the local registry on the receiving side without async I/O.

woksin:
Fixed. `AppendedEventSerializer` now parses the schema synchronously — `_schemasBySchemaJson.GetOrAdd(appendedEventWithSchema.Schema, JsonSchema.FromJson)` against a `ConcurrentDictionary<string, JsonSchema>`. The `JsonSchema.FromJsonAsync(...).GetAwaiter().GetResult()` that blocked the Orleans serialization hot path is gone, and `FromJson` is a real synchronous overload rather than a wrapper.

The file contains no `async`, `Task`, `GetAwaiter` or `.Result` at all now.

Issue #1437, opened 2024-09-25
Event and EventType details views should have scrollbars when content overflows vertically


einari:

woksin:
Triage question — empty body, and this one is old enough (2024-09-25) that the first question is whether it still reproduces.

@einari — does it? The workbench has been through substantial change since, and a vertical-overflow fix is the kind of thing that gets fixed incidentally. If it still happens, could you note:

- **Which view** — the event details pane, the event *type* details pane, or both?
- **What overflows** — the JSON content of a large event, the schema of a wide event type, or the metadata block?
- **A screenshot** would settle it faster than a description.

If it no longer reproduces, this can be closed. I did not want to guess at a CSS change against a symptom I cannot see.

woksin:
## Correction — I should not have asked; this was already fixed

I asked above whether this still reproduces. It does not, and the answer was available in a merged PR — #3489:

> The Workbench event details view now shows a vertical scrollbar when the Context or Content payload overflows, instead of clipping (#1437)

That is this issue, referencing it by number. Apologies for the round-trip.

One thing the PR bullet does *not* cover, and which the title claims: it names the **event details** view, not the **event type** details view. If the latter still clips a wide schema, that is a separate remaining defect worth its own issue rather than keeping this one open.

@einari @woksin — safe to close.

einari:
Fixed by #3489 — the Workbench event details view shows a vertical scrollbar when Context or Content overflows.

PR #3489, opened 2026-07-22
Enforce PII unique constraints, add UoW tracing, speed up serialization

woksin:
A batch of independent fixes and improvements found while going through open issues.

## Added

- OpenTelemetry tracing spans for client unit-of-work commit and rollback, tagged with the correlation id (#1396)

## Changed

- Faster reactor, reducer and event-type dispatch on the .NET client (#1309)

## Fixed

- Removed a per-event blocking schema lookup on the Orleans serialization hot path, improving throughput when events cross silo boundaries (#3168)
- Process-level unhandled and unobserved-task exceptions in the server and API hosts now reach the logging pipeline and OpenTelemetry exporter instead of only the console (#1343)
- The Workbench event details view now shows a vertical scrollbar when the Context or Content payload overflows, instead of clipping (#1437)
- Corrected the Aspire hosting guide: Chronicle's MongoDB storage requires a replica set with `directConnection=true`; the previous guidance used a standalone `mongod` that silently breaks change streams (#3400)

## Security

- A unique constraint on a `[PII]` property is now enforced. The constraint index previously hashed the encrypted value, which differs on every append, so two subjects sharing the same value never collided and the constraint silently did nothing; values are now hashed from the decrypted content and existing indexes are rebuilt on upgrade (#3430)


woksin:
### Reviewer notes (not release notes)

**Internal changes bundled in this PR** (deliberately kept out of the release-notes body):
- Kernel concept data classes turned into `record`s — `FailedPartitionAttempt`, `InFlightEvent`, `ConnectedClient`, the `Job*` progress/status concepts, `ClosedStreamConstraintDefinition`. Non-positional conversion, serialized shape unchanged (#1537).
- A local read-through cache for projection definitions storage (`CachingProjectionDefinitionsStorage`), wired per event store on the Mongo/SQL storages. Documented as an intentionally eventually-consistent read-accelerator; no cross-silo invalidation (#1248).
- Out-of-process integration specs for constraints — violation details, a concurrent-append race, and a `[PII]` unique-constraint collision across two subjects, which is the real-engine regression proof for the security fix above (#1406).

**A regression this PR both introduced and fixed:** the unit-of-work tracing wiring (#1396) first resolved `IActivitySource<UnitOfWork>` with `GetRequiredKeyedService`, which throws in any client DI that doesn't register it (the integration host, and real apps). Unit specs missed it because `EventStoreForTesting` bypasses that path; the new #1406 integration specs caught it. Resolved with `GetKeyedService` + the default-source fallback — squashed into the #1396 work.

**Verification (local, since CI is blocked — see below):** every changed project builds Release clean; specs green per project — DotNET.Specs 1857, Core.Specs 1673, Storage.Specs 41, Storage.MongoDB.Specs 446, Storage.Sql.Specs 40, Storage.InMemory.Specs 46, unit-of-work tracing 51; the #1406 integration specs 15/15 (Testcontainers Mongo), including the PII collision proof. The Aspire snippet compiles through `validate-client-snippets.py`.

**⚠️ CI blocker — pre-existing, not caused by this PR.** `dotnet-build` / `dotnet-build-development` / `verify` fail on `IDX00001` in `Integration/Api` and `Server.csproj`: a fresh NuGet restore resolves `Microsoft.IdentityModel.JsonWebTokens 8.20.0` transitively while `Directory.Packages.props` pins `8.19.2`, skewing it against `Protocols`/`Protocols.OpenIdConnect`/`System.IdentityModel.Tokens.Jwt` at `8.19.2`; `IDX00001` isn't in `NoWarn`, so warnings-as-errors fails the build. It reproduces on any freshly-restored PR off current `main` (and now locally). Every changed project still produces its DLL. **Fix (its own PR — I did not touch manifests):** pin all four `Microsoft.IdentityModel.*` / `System.IdentityModel.Tokens.Jwt` packages to `8.20.0` in `Directory.Packages.props`.

**Deferred (need maintainer judgment, not autonomous wins):**
- #3434 (consolidate Testing doubles onto Storage.InMemory) — pure refactor with the widest blast radius across the in-process spec harness, no CI net.
- #1469 + #1470 (expose JsonConverters / override JsonSchema representation) — two halves of one serialization feature in a hot, regression-prone path, each carrying a design decision (`Globals.JsonSerializerOptions` migration, converter precedence).
- #3400 code helper (a `WithMongoDB` replica-set provisioning helper) — no Aspire test harness exists; the docs fix here is the low-risk half.
- #1417 (configurable timeouts — ~15 sites, sprawling) and #1433 (debugger observer timeout — maintainer noted wariness about auto-configuring Orleans).

**Recommend closing as already-resolved (with evidence):**
- #1395 (client reactor/reducer logging) — error logging + tracing already present (`ReactorsLogMessages`, `ReactorInvokerLogMessages`, `ReducersLogMessages`, `ReactorTraces`/`ReducerTraces`).
- #1791 (keyed Meter scoped to Chronicle) — already done via `AddNamedMeter(WellKnown.MeterName)`; all kernel consumers resolve `[FromKeyedServices]`.
- #3463 (ChildrenFrom OOP materialization) — fixed by PR #3464 (merged), with a regression spec.
- #3405 (XPlat Code Coverage collector) — not reproducible at HEAD; `coverlet.collector` is wired centrally and there are zero occurrences in current CI logs.

woksin:
### Additional justification for the #1309 change (found via a later bug-hunt)

The frozen-collections change to `Reactors._handlers` (this PR's #1309 commit) also fixes a confirmed **data race**, not just a micro-optimization: on origin/main `Reactors._handlers` is a plain `Dictionary` mutated in place (`Clear()` + rebuild) by `RecreateHandlersForReconnect` on a background disconnect thread, while readers (`GetHandlerById`/`GetHandlerFor`/`GetStateFor`/`Replay`/`GetFailedPartitionsFor`) and other mutators enumerate/index it without taking `_registerLock`. A reader coinciding with a reconnect rebuild throws `InvalidOperationException: Collection was modified` or reads a torn dictionary. Converting `_handlers` to a `FrozenDictionary` rebuilt-and-swapped atomically gives readers a consistent immutable snapshot — eliminating the race. (`Reducers` already used the swap pattern, which is why only Reactors was affected.)

PR #3490, opened 2026-07-23
Fix crash releasing an erased subject's PII collection

woksin:
## Fixed

- Releasing a coarse `[PII]` collection for a subject whose encryption key has been erased (right-to-erasure) no longer throws — it now reads as an empty collection. Previously the empty released value crashed re-parsing on the read side, and on the observer path failed the partition on every retry, which could quarantine the observer after an erasure.

### DIFF
--- a/Source/Kernel/Core.Specs/Compliance/for_JsonComplianceManager/when_releasing_a_compliant_list_and_the_key_has_been_deleted.cs
+++ b/Source/Kernel/Core.Specs/Compliance/for_JsonComplianceManager/when_releasing_a_compliant_list_and_the_key_has_been_deleted.cs
@@ -1,0 +1,26 @@
+// Copyright (c) Cratis. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Text.Json.Nodes;
+
+namespace Cratis.Chronicle.Compliance.for_JsonComplianceManager;
+
+public class when_releasing_a_compliant_list_and_the_key_has_been_deleted : given.a_value_handler_and_a_type_with_a_compliant_list
+{
+    const string Identifier = "9ae5067b-2920-4c97-a263-efe35bec2b43";
+    JsonObject _result;
+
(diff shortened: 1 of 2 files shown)


woksin:
### Reviewer notes

Found via a proactive bug-hunt (no prior GitHub issue). Severity: high — availability + compliance.

**Root cause:** in `JsonComplianceManager.RestoreReleasedContainerShape`, a coarse `[PII]` on a whole collection is blob-encrypted to one ciphertext string but keeps schema type `array`. After the subject's key is crypto-shredded (right-to-erasure), `PIICompliancePropertyValueHandler.Release` returns `JsonValue.Create(string.Empty)` (by design), and the array-restore path then ran `JsonNode.Parse("")`, which throws `JsonReaderException` → re-thrown as `InvalidOperationException`. Scalar `[PII]` returns `""` gracefully; only *collection* `[PII]` crashed — inconsistent with the erased-subject contract.

**Blast radius:** read side — queries/materialization of an erased-subject document crash. Observer side (worse) — `Observer.Handling` → `DecryptEvents` → `EventCompliance.Release` → same manager throws on every retry → `PartitionFailed` → the observer eventually **quarantines**. So erasing one subject could permanently poison a projection/reducer.

**Fix:** guard empty/whitespace before `JsonNode.Parse` and return an empty `JsonArray` (an erased collection reads as empty). Non-empty path unchanged.

**Verification (local; CI is blocked repo-wide by the pre-existing IdentityModel skew):** the new regression spec `for_JsonComplianceManager/when_releasing_a_compliant_list_and_the_key_has_been_deleted` FAILS before the fix with exactly `JsonReaderException: The input does not contain any JSON tokens` and PASSES after. Core + Core.Specs build 0/0; `for_JsonComplianceManager` 27 pass; full Compliance area 101 pass, 0 fail.

**Not verified:** the observer-quarantine path is reasoned from the shared code path, not exercised via an out-of-process observer integration test.

woksin:

PR #3492, opened 2026-07-23
Fix child-level [RemovedWith] routing in model-bound projections

woksin:
# Summary

Fixes a model-bound projection bug where removing a single keyed child from a `[ChildrenFrom]` collection removed the entire read-model document instead of just that child.

## Fixed

- A `[RemovedWith]` placed alongside a `[ChildrenFrom]` on the same collection member now removes only the targeted keyed child — previously it removed the whole read-model document at the root, or the whole parent item for a nested collection.
- Nested (grandchild) child collections now route their `[RemovedWith]` removal to the correct grandchild definition rather than the parent item.
- `[RemovedWithJoin]` declared on a child collection is routed the same way, removing the joined child rather than the containing document.


woksin:
### Reviewer notes (not release notes)

**Root cause.** A `[RemovedWith]`/`[RemovedWithJoin]` alongside a `[ChildrenFrom]` on the same member was appended to the root (or parent) removal definition, so the sink emitted a whole-document/whole-parent removal instead of a keyed child `$pull`. The fix guards that member (`!isChildrenCollection`) and re-homes the removal onto the child definition, with `RouteNestedRemovedWith` handling grandchildren.

**Coverage added** (`for_ModelBoundProjectionBuilder`):
- root, record-parameter collection
- nested/grandchild via record-parameter (`RootWithNestedRemovableGrandchildren`)
- nested/grandchild via class-property (`RootWithClassChildGrandchildren`)
- naming-policy key round-trip (camel-casing policy — the re-home looks the child up by the policy-converted name)
- `[RemovedWithJoin]` on a child collection

Each spec includes negative assertions that the root/parent no longer carries the removal — the actual regression guards.

**Verification:** full `DotNET.Specs` green — 1890 passed / 0 failed (1883 → 1890, +7 from the join spec). `for_ModelBoundProjectionBuilder` subset: 362 passed.

**Known limitation:** the root-level **class-property** path is not directly covered by its own model, but its routing is the mirror of the root record-parameter path (covered) and the nested class-property path (covered). Happy to add a dedicated root-class model if preferred.

woksin:
Added a second commit `Align Microsoft.IdentityModel package versions to 8.20.0` to unblock CI. It is **unrelated to the projection fix** — a fresh restore floated part of the `Microsoft.IdentityModel.*` family to 8.20.0 while `Protocols`/`Protocols.OpenIdConnect`/`System.IdentityModel.Tokens.Jwt` lagged at 8.19.2, tripping `IDX00001` (same-version check) as a build error under warnings-as-errors. This was already latent on `main` (a cached local build hid it). Pinned the whole family to 8.20.0 and referenced it as one unit from `Server`. Verified with a forced fresh restore: `dotnet list Server package` shows all seven at 8.20.0, and the full `Chronicle.slnx` Release build is 0 errors with `IDX00001` gone. Kept out of the release-notes body as an internal build change; happy to split it into its own PR if you'd prefer.

woksin:
Folded into the consolidated non-Workbench PR #3507 — kept the child-level `[RemovedWith]` routing fix and its specs; dropped the stale Microsoft.IdentityModel 8.20.0 pinning since main resolves the skew at 8.21.0. Closing in favor of #3507.

Issue #1684, opened 2025-01-31
We need an end-to-end API for changing names of an identity


einari:
Name of an identity could change. Therefor we need to expose an API all the way from the Kernel to the Client to be able to change the name of an identity based on the identifier of the identity.

PR #3493, opened 2026-07-23
Add the ability to rename an identity

woksin:
## Added

- The ability to rename an identity's display name, from the kernel storage layer through the gRPC `IIdentities` contract to the client SDK. The rename is keyed by subject and preserves the subject, username and on-behalf-of chain — so it never affects the PII encryption key, which is derived from the subject (#1684)

### DIFF
--- a/Source/Clients/Testing/EventSequences/InMemoryIdentityStorage.cs
+++ b/Source/Clients/Testing/EventSequences/InMemoryIdentityStorage.cs
@@ -47,6 +47,9 @@ internal sealed class InMemoryIdentityStorage : IIdentityStorage
         Task.FromResult(KernelIdentities::Identity.System);
 
     /// <inheritdoc/>
+    public Task Rename(string subject, string name) => Task.CompletedTask;
+
+    /// <inheritdoc/>
     public Task<IEnumerable<KernelIdentities::Identity>> GetAll() =>
         Task.FromResult(Enumerable.Empty<KernelIdentities::Identity>());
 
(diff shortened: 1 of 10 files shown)


woksin:
### Reviewer notes

Wired end-to-end: `IIdentityStorage.Rename(subject, name)` → MongoDB (`$set Name where Subject==subject` + in-memory cache refresh), SQL (EF), InMemory + Testing (both are documented no-op backends), and outward via an additive gRPC `IIdentities.RenameIdentity` operation + `RenameIdentityRequest` message, which is how the client SDK reaches identities (no separate client class needed). `identities.proto` was hand-synced to match generator output.

**Keyed by `Subject`, not `IdentityId` (deliberate):** the client-facing `Identity` contract exposes `Subject`/`Name`/`UserName`/`OnBehalfOf` but not the internal `IdentityId` (and `GetAll()` returns no id), so `IdentityId` is unreachable from the client without changing the existing message. `Subject` is documented as the identity's identifier and is the immutable value we preserve.

**Key-safety invariant:** Chronicle keys PII encryption on the compliance subject (`EncryptionKeyIdentifier`), which is separate from `Identity.Name`. Rename mutates only `Name`; the spec `should_preserve_the_subject_so_the_encryption_key_is_unaffected` asserts subject invariance directly. (A full encrypt→rename→decrypt round-trip isn't expressible at the storage layer since encryption lives above it.)

**Non-breaking:** the gRPC change is a new operation + new message only; existing signatures unchanged → minor.

**Verification (local; CI blocked repo-wide by the pre-existing IdentityModel skew):** all affected projects build 0/0. Storage.MongoDB.Specs 451 pass (incl. 5 new rename facts), Storage.InMemory.Specs 46, Storage.Sql.Specs 40, Core.Specs 1657 (+1 pre-existing skip), 0 failures.

**Left out (scope):** a Workbench UI/ASP.NET endpoint for rename — that's frontend and needs the Arc proxy generator (disabled in this build environment). The Kernel → gRPC → client-SDK path the issue asks for is complete.

woksin:

PR #3494, opened 2026-07-23
Fix read-model live queries freezing after a reconnect

woksin:
## Fixed

- Projection-backed observable read-model queries now keep emitting after a connection drop — previously the watch stream was never re-established after the first reconnect, so live queries silently froze on stale data

### DIFF
--- a/Source/Clients/DotNET.Specs/ReadModels/for_ReadModelWatcher/when_client_reconnects.cs
+++ b/Source/Clients/DotNET.Specs/ReadModels/for_ReadModelWatcher/when_client_reconnects.cs
@@ -1,0 +1,33 @@
+// Copyright (c) Cratis. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Cratis.Chronicle.Connections;
+using Cratis.Chronicle.Contracts.ReadModels;
+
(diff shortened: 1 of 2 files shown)


woksin:
### Reviewer notes
Found via a client-SDK bug-hunt (no prior issue). Severity: high. `ReadModelWatcher.Start()` had a one-shot `_started` latch that was never reset, so on reconnect `OnConnected → ClientConnected → Start()` no-opped and the gRPC watch stream (opened by the first `Services.ReadModels.Watch`) was never re-established → the read model's observable never emitted again for the process lifetime. Reducer-backed observable queries were unaffected (in-process `NotifyChange`), which masked it.
Fix: wire the existing `OnDisconnected` event to reset the latch and dispose the subscription; capture the `Subscribe` `IDisposable` (previously discarded → leak on any re-subscribe) and dispose it before re-subscribing; add onError/onCompleted routing through the same reset so a stream fault tears down cleanly. Observable contract unchanged.
Verification (local; CI blocked repo-wide by pre-existing IdentityModel skew): new `for_ReadModelWatcher/when_client_reconnects` fails 2/2 before the fix (only one `Watch` call; no dispose) and passes 2/2 after. DotNET client + Specs build 0/0; ReadModel area 112 pass, ConnectionLifecycle 7 pass.

woksin:

Issue #1925, opened 2025-06-29
Filtering of properties in Event Sequence table


einari:
It should be possible to simply apply a filter for any of the columns in the event sequence view.

woksin:
## Resolved on `main`

Landed in #3496:

> Filtering on the Workbench event-sequence table: a whole-sequence server-side filter by event source id, plus page-scoped column and global filters with an event-type dropdown

The same PR also fixed a defect this issue would otherwise have been blocked behind — the sequence-number and event-type columns rendered blank because their field paths named a non-existent `metadata` member instead of `context`.

@einari @woksin — safe to close.

woksin:
Closing — released in **v16.15.0**.

Delivered by #3496. Verification against `main` is in the comment above.

PR #3496, opened 2026-07-23
Add filtering to the Workbench event-sequence table

woksin:
## Added

- Filtering on the Workbench event-sequence table: a whole-sequence server-side filter by event source id, plus page-scoped column and global filters with an event-type dropdown (#1925)

## Fixed

- The event-sequence table's sequence-number and event-type columns no longer render blank — their field paths referenced a non-existent `metadata` member instead of `context` (#1925)


woksin:
### Reviewer notes
Frontend-only; edits `EventList.tsx` only (no backend/proxy). Two honest scopes: (1) a whole-sequence **server-side** filter wired to the one filterable query param, `AppendedEventsParameters.eventSourceId` (debounced); (2) **page-scoped** column/global filters + event-type dropdown via `DataTableForQuery`'s `clientFiltering`, which only filters the loaded 20-row page — labeled as such in help text so users aren't misled.
Two necessary deviations: the event-type dropdown sources options from the `AllEventTypes` registry (mirroring the working sibling `Sequences.tsx`) because `DataTableForQuery` doesn't expose loaded rows; and the column `field`/`dataKey` paths were corrected from a non-existent `metadata.*` to `context.*` (`AppendedEvent` has only `context`), which is the "Fixed" bullet — those columns were rendering blank before.
Verification (Source/Workbench): `tsc -b` app + specs 0 errors, vitest 20/20 (incl. 7 new helper specs), lint:ci 0 errors, build succeeds. Not verified: rendered UI / debounce feel.

PR #3497, opened 2026-07-23
Fix AppendMany livelock and sequence gaps on storage failure

woksin:
## Fixed

- Appending a batch of events (`AppendMany`) no longer risks an infinite retry loop that permanently wedges an event sequence when storage reports a duplicate sequence number (for example after a crash between a storage commit and state persistence) — the batch is now re-numbered from the next available sequence number before each retry
- Appending a batch of events no longer leaves a permanent sequence-number gap and a tail pointing at an event that was never persisted when the storage append throws — the sequence number now advances only after a successful append

### DIFF
--- a/Source/Kernel/Core.Specs/EventSequences/for_EventSequence/given/SimulatedStorageError.cs
+++ b/Source/Kernel/Core.Specs/EventSequences/for_EventSequence/given/SimulatedStorageError.cs
@@ -1,0 +1,10 @@
+// Copyright (c) Cratis. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Cratis.Chronicle.EventSequences.for_EventSequence.given;
+
+/// <summary>
+/// The exception that is thrown by a substituted storage to simulate a non-duplicate (transient/aborted) failure
+/// while appending a batch of events.
+/// </summary>
+public class SimulatedStorageError() : Exception("Simulated non-duplicate storage error");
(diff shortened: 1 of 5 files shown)


woksin:
### Reviewer notes
Found via a jobs/replay bug-hunt (no prior issue). Both defects were in `EventSequence.AppendMany` and are the batch-path analogues of behavior the single-event `Append` path already had right.

- **Livelock (HIGH):** the batch's sequence numbers were baked once before the retry loop, and `HandleAppendedDuplicateEventForMany` advanced `State.SequenceNumber` but never re-numbered the batch, so on a `DuplicateEventSequenceNumber` the same colliding numbers resubmitted forever → the single-threaded grain wedged (no further append to that sequence). Fix re-numbers the batch from `NextAvailableSequenceNumber` before resubmission.
- **Gap/phantom tail (MED):** `State.SequenceNumber` was advanced in the build loop before the storage write, with no rollback on a thrown (non-duplicate) error → a permanent gap and a `GetTailSequenceNumber` pointing at an unpersisted event. Fix numbers against a local and commits `State.SequenceNumber` only after a successful append.

Extracted the batch build/append/commit region into an `internal AppendManyToStorage` seam; the success-path behavior is identical.

**Verification (local; CI blocked repo-wide by the pre-existing IdentityModel skew):** new specs drive the real grain via the Orleans test kit over a substituted `IEventSequenceStorage`. Livelock spec: with a 20-iteration safety cap, PASSES with the fix (terminates in 2 calls, resubmits non-colliding numbers) and FAILS BOUNDED (cap reached, no hang) with the renumber removed. Gap spec: `GetNextSequenceNumber`==5/tail==4 with the fix, 6/5 (gap) without. Core + Core.Specs build 0/0; for_EventSequence 10 pass, EventSequences area 145, full Core.Specs 1664 (+1 pre-existing skip), 0 failures.

Note: this touches only the `AppendMany` region; PR #3489 touches a different `EventSequence.cs` method — a later rebase should be conflict-free.

woksin:

PR #3498, opened 2026-07-23
Make SQL username and email lookups case-insensitive

woksin:
## Fixed

- Username and email lookups on the SQL backend are now case-insensitive, matching the in-memory and MongoDB backends. Previously SQL matched case-sensitively, so the same name in different casing could create duplicate accounts and cause login failures when the stored casing differed

### DIFF
--- a/Source/Kernel/Storage.Sql.Specs/Cluster/Security/for_UserStorage/given/a_user_storage.cs
+++ b/Source/Kernel/Storage.Sql.Specs/Cluster/Security/for_UserStorage/given/a_user_storage.cs
@@ -1,0 +1,70 @@
+// Copyright (c) Cratis. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Cratis.Arc.EntityFrameworkCore.Concepts;
+using Cratis.Chronicle.Concepts.Security;
+using Microsoft.Data.Sqlite;
+using Microsoft.EntityFrameworkCore;
+
+namespace Cratis.Chronicle.Storage.Sql.Cluster.Security.for_UserStorage.given;
+
+/// <summary>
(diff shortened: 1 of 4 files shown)


woksin:
### Reviewer notes
Found via a security-storage bug-hunt (no prior issue). `GetByUsername`/`GetByEmail` are used for login and duplicate detection at user creation; they were case-insensitive on InMemory (`OrdinalIgnoreCase`) and MongoDB (`"en"` primary collation) but case-sensitive on SQL (SQLite + PostgreSQL) — the classic "specs pass on the InMemory backend, real backend behaves differently" trap.

Fix compares `u.Username.ToLower() == value.ToLowerInvariant()`. `ToLower()` (not `ToLowerInvariant()`) is used on the column deliberately: EF's SQLite provider only translates the parameterless `string.ToLower()` to SQL `LOWER()`, and Npgsql translates it too, so it's the single form that runs server-side on both shipped providers instead of client-evaluating the table. CA1304/CA1311/CA1862 are false positives for a server-translated predicate and are suppressed with a scoped `#pragma` + justification (the input side uses `ToLowerInvariant()`, analyzer-clean). No schema migration.

Verification (local; CI blocked repo-wide by the pre-existing IdentityModel skew): new SQLite-in-memory specs under `for_UserStorage` — before the fix 4/8 fail (the case-differs cases return null), after 8/8 pass. Storage.Sql + Specs build 0/0; Storage.Sql.Specs 48 pass (40 pre-existing + 8 new). InMemory/Mongo untouched.

Residual nuance: this matches ASCII case-insensitivity only; MongoDB's collation is additionally accent-insensitive (not required per scope). PostgreSQL `LOWER()` behavior was reasoned from the Npgsql translator (verified locally on SQLite, which the spec harness uses).

woksin:

PR #3499, opened 2026-07-23
Fix token revocation no-op and pruning of valid tokens

woksin:
## Fixed

- Token pruning no longer deletes still-valid tokens — it now removes only inactive or expired tokens created before the pruning threshold, so long-lived valid sessions/refresh tokens are no longer force-signed-out

## Security

- Token revocation now actually revokes. Previously every bulk revocation operation (by application, by subject, by authorization, and by client/status/type) reported success without revoking anything, so rotating a leaked client secret, removing an application, or signing a user out everywhere left already-issued tokens valid until their natural expiry

### DIFF
--- a/Chronicle.slnx
+++ b/Chronicle.slnx
@@ -22,6 +22,7 @@
         <Project Path="Source/Kernel/Core/Core.csproj" />
         <Project Path="Source/Kernel/Core.Specs/Core.Specs.csproj" />
         <Project Path="Source/Kernel/Server/Server.csproj" />
+        <Project Path="Source/Kernel/Server.Specs/Server.Specs.csproj" />
         <Project Path="Source/Kernel/Storage/Storage.csproj" />
         <Project Path="Source/Kernel/Storage.InMemory/Storage.InMemory.csproj" />
         <Project Path="Source/Kernel/Storage.InMemory.Specs/Storage.InMemory.Specs.csproj" />
--- a/Source/Kernel/Server.Specs/Authentication/OpenIddict/for_TokenStore/given/a_token_store.cs
+++ b/Source/Kernel/Server.Specs/Authentication/OpenIddict/for_TokenStore/given/a_token_store.cs
@@ -1,0 +1,18 @@
+// Copyright (c) Cratis. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Cratis.Chronicle.Storage.InMemory.Security;
+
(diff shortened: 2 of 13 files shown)


woksin:
### Reviewer notes
Found via a security-storage bug-hunt (no prior issue). The token-revocation half is a **fail-open security control**.

**Revocation (Security):** all four `IOpenIddictTokenStore.Revoke*` methods in `Server/Authentication/OpenIddict/TokenStore.cs` returned `new(0L)` without touching storage — the sibling `AuthorizationStore` implements the analogous methods for real, so the token store was simply left stubbed. Now each loads matching tokens via the existing `ITokenStorage.FindBy*` lookups, sets `Status = OpenIddictConstants.Statuses.Revoked`, persists via `Update`, and returns the count (shared `Revoke(IEnumerable<Token>)` helper). No lookup gaps — every needed `FindBy*` already existed. `RevokeAsync` uses the same client/subject/status branch ladder as `FindAsync` and returns 0 when neither client nor subject is supplied (matches AuthorizationStore).

**Pruning (Fixed):** `PruneAsync(threshold)` deleted every token with `CreationDate < threshold` regardless of `Status`/`ExpirationDate`, so OpenIddict's ~14-day hosted pruner force-deleted still-valid long-lived tokens. Fixed in all three backends to `CreationDate < threshold AND (Status != Valid OR ExpirationDate < now)`. Added `Storage/Security/TokenStatuses.cs` (`Valid = "valid"`) so storage doesn't take an OpenIddict dependency.

**Structural note:** the kernel `Server` project had no spec project, so this adds a minimal `Source/Kernel/Server.Specs` (IsTestProject, references Server + Storage.InMemory + Storage) and registers it in `Chronicle.slnx`, to unit-test the OpenIddict `TokenStore` against the real `InMemoryTokenStorage`.

**Verification (local; CI blocked repo-wide by the pre-existing IdentityModel skew):** before/after captured empirically — old no-op `TokenStore`: 11/15 revocation facts fail (count 0, status stays valid); old prune: `should_report_two_tokens_removed` fails (3 vs 2, the valid token wrongly deleted). After: Server.Specs 15/15, Storage.InMemory.Specs 50/50. All touched projects build 0/0 (Server/Server.Specs 0/0 with the pre-existing `IDX00001` suppressed command-line only for verification — NOT committed). Not run: Mongo/SQL token-storage spec execution — the three Prune predicates are identical, but the Mongo/SQL LINQ translation of the new predicate was verified by compilation only.

woksin:

PR #3500, opened 2026-07-23
Fix premature observer completion and leaked read cursors

woksin:
## Fixed

- Waiting for observers to catch up no longer reports success against an observer that has never handled any events, so a caller waiting for an append to be processed no longer proceeds against a stale or empty read model on the first run
- Event and read-model read cursors are now disposed when an enumeration error occurs, instead of leaking the underlying database cursor

### DIFF
--- a/Source/Kernel/Core.Specs/Services/EventSequences/for_EventSequences/given/all_dependencies.cs
+++ b/Source/Kernel/Core.Specs/Services/EventSequences/for_EventSequences/given/all_dependencies.cs
@@ -1,0 +1,32 @@
+// Copyright (c) Cratis. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Text.Json;
+using Cratis.Chronicle.Events;
+using Cratis.Chronicle.Storage;
+
+namespace Cratis.Chronicle.Services.EventSequences.for_EventSequences.given;
+
+public class all_dependencies : Specification
+{
+    protected IGrainFactory _grainFactory;
+    protected IStorage _storage;
+    protected IEventStoreStorage _eventStoreStorage;
+    protected IEventStoreNamespaceStorage _namespaceStorage;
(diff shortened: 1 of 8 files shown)


woksin:
### Reviewer notes
Found via a gRPC-services bug-hunt (no prior issue). Likely a root cause of long-standing "observer hasn't caught up" flakiness.

**Premature completion (MED-HIGH):** on the gRPC contract, `ObserverInformation.LastHandledEventSequenceNumber` is a plain `ulong` (the converter flattens the domain `EventSequenceNumber`). A never-run observer resolves to `ObserverState.Empty` → `EventSequenceNumber.Unavailable` = `ulong.MaxValue`, so `MaxValue >= tail` was trivially true and `WaitForCompletion` returned success before the observer processed anything. Fix casts back to `EventSequenceNumber` to reuse the `.IsActualValue` guard the rest of the kernel already applies (`Observer.Handling`, `CatchingUpInFlight`, `Observer.Watchdog`), parenthesized so the failed-observer OR still applies. `IsActualValue` also excludes the `Max` sentinel.

**Cursor leak (LOW-MED):** two service read methods (`EventSequences.GetEventsFromEventSequenceNumber`, `ReadModels.GetAllInstances`) called a bare `cursor.Dispose()` after the read loop; if `MoveNext()` threw, the cursor (wrapping a DB cursor) leaked. Converted both to `using var cursor` (matching the sibling methods that already did).

**Verification (local; CI blocked repo-wide by the pre-existing IdentityModel skew):** before/after captured — Bug 1 negative spec fails 2/2 before (returns success on the first poll / `IsSuccess` true) and passes after; Bug 2's `should_dispose_the_cursor` facts fail before (Dispose never reached after a thrown `MoveNext`) and pass after. Core + Core.Specs build 0/0; affected service specs (Observers/ReadModels/EventSequences) 63 pass, 0 fail.

woksin:

PR #3501, opened 2026-07-23
Fix event seeding losing events on a transient failure

woksin:
## Fixed

- Seeding no longer permanently loses events when applying them to an event store's namespaces fails partway through — the global seed record is now committed only after every namespace has been seeded, so a transient failure is retried instead of leaving some namespaces without the events
- Seeding a namespace no longer silently skips events when the append fails — events are marked seeded only after they have been appended, so a retry re-attempts the append instead of treating them as already done

### DIFF
--- a/Source/Kernel/Core.Specs/Seeding/for_EventSeeding/given/a_global_event_seeding_grain.cs
+++ b/Source/Kernel/Core.Specs/Seeding/for_EventSeeding/given/a_global_event_seeding_grain.cs
@@ -1,0 +1,17 @@
+// Copyright (c) Cratis. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Cratis.Chronicle.Concepts.Seeding;
+
+namespace Cratis.Chronicle.Seeding.for_EventSeeding.given;
+
+public class a_global_event_seeding_grain : an_event_seeding_grain
+{
+    void Establish()
+    {
+        _key = EventSeedingKey.ForGlobal("TestEventStore");
+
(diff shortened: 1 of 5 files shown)


woksin:
### Reviewer notes
Found via a seeding/namespaces bug-hunt (no prior issue). Both defects were the same root pattern in `EventSeeding`: a durable/in-memory "seeded" marker written BEFORE the fallible operation it gates, turning a transient failure into permanent/silent seed loss.

- **Global branch (MED-HIGH):** `TrackSeededEvent` + `WriteStateAsync` ran before the per-namespace dispatch loop. If a namespace grain threw mid-loop, the retry found `IsAlreadySeeded == true` → dispatch skipped → the remaining namespaces never received the events (durable loss; `NamespacesReactor.Added` only backfills brand-new namespaces). Fix dispatches to all namespaces first (each has its own idempotency guard, so re-dispatch is append-idempotent), then commits the global tracking.
- **Namespace branch (MED):** `GetEventsToSeed` marked entries seeded in-memory before the append; on a throw, `WriteStateAsync` was skipped (persisted state clean) but the in-memory set was dirty → a same-activation retry silently skipped the events. Fix splits query from mutation (a new `SeedableEvent` record) and marks each batch seeded only after its append succeeds.

Injected `IGrainFactory` (the `HandleEventsForObserver` pattern) so the global dispatch is substitutable; the only constructor of `EventSeeding` is the spec base, so the change is contained.

**Verification (local; CI blocked repo-wide by the pre-existing IdentityModel skew):** before/after captured — with the buggy orderings reintroduced, the 5 discriminating facts fail (namespace: append received 1 not 2; global: no re-dispatch, tracking committed too early); after, Seeding specs 11/11 pass. Core + Core.Specs build 0/0.

woksin:

PR #3502, opened 2026-07-23
Fix SQL append livelock on a stale duplicate sequence number

woksin:
## Fixed

- The SQL event-store backend no longer risks an infinite retry loop when appending at a sequence number that is already taken (for example after a silo restart leaves the grain's counter behind the stored tail). Storage now reports the true next-available sequence number so the append recovers, matching the MongoDB and in-memory backends

### DIFF
--- a/Source/Kernel/Storage.Sql.Specs/EventStores/Namespaces/EventSequences/for_EventSequenceStorage/given/an_event_sequence_storage.cs
+++ b/Source/Kernel/Storage.Sql.Specs/EventStores/Namespaces/EventSequences/for_EventSequenceStorage/given/an_event_sequence_storage.cs
@@ -1,0 +1,120 @@
+// Copyright (c) Cratis. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Dynamic;
+using Cratis.Arc.EntityFrameworkCore.Concepts;
+using Cratis.Chronicle.Concepts;
+using Cratis.Chronicle.Concepts.Events;
+using Cratis.Chronicle.Concepts.EventSequences;
+using Cratis.Chronicle.Concepts.Identities;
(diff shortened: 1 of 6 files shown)


woksin:
### Reviewer notes
Found via an event-storage bug-hunt (no prior issue). `DuplicateEventSequenceNumber.NextAvailableSequenceNumber` is, by contract, the next FREE slot the append grain jumps to before retrying. The SQL backend returned the *just-attempted (occupied)* number at both `Append` and `AppendMany` return sites, so the grain reset its counter to the occupied slot, retried the same number, the pre-read found the same row again, and it looped forever — permanently stalling the sequence whenever the grain's persisted `State.SequenceNumber` lags the real tail (silo restart, reset grain state, replayed append). MongoDB (`highest.Next()`) and in-memory (`Max+1`) return the true next-available; SQL was the odd one out.

This complements #3497 (the grain-level `AppendMany` renumber): that fix renumbers the batch *from* `NextAvailableSequenceNumber`, so on SQL it can only recover once the storage returns the correct value — both are needed for the SQL path to self-heal.

Fix adds a `GetNextAvailableSequenceNumber` helper (`MaxAsync((ulong?) SequenceNumber).Next()`, the same server-translatable shape already used by `GetTailSequenceNumbers`, empty-case guarded) used at both return sites.

**Verification (local; CI blocked repo-wide by the pre-existing IdentityModel skew):** 5 new SQLite-in-memory specs (`for_EventSequenceStorage`). With the fix reverted, the 4 duplicate-path facts fail (`NextAvailableSequenceNumber` comes back as the occupied `1` instead of `3`); after, all 10 pass. Storage.Sql + Specs build 0/0; Storage.Sql.Specs 50 pass. Specs assert the storage return value directly (bounded) — they never loop the grain.

woksin:

PR #3503, opened 2026-07-23
Fix event-type generation handling in Mongo and SQL storage

woksin:
## Fixed

- Adding a new generation to an existing event type now works on the MongoDB backend. The storage's "does this generation exist" check ignored the generation and reported any existing type as already having every generation, so legitimate event-type migrations were rejected as an unexpected schema change
- Resolving an event type by a specific generation from the SQL database on a cold cache now returns that generation's schema instead of always the first generation's

### DIFF
--- a/Source/Kernel/Storage.MongoDB.Specs/EventTypes/for_EventTypesStorage/given/an_event_types_storage.cs
+++ b/Source/Kernel/Storage.MongoDB.Specs/EventTypes/for_EventTypesStorage/given/an_event_types_storage.cs
@@ -1,0 +1,42 @@
+// Copyright (c) Cratis. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Cratis.Chronicle.Concepts;
+using Cratis.Chronicle.Concepts.Events;
+using Microsoft.Extensions.Logging;
+using MongoDB.Driver;
+
+namespace Cratis.Chronicle.Storage.MongoDB.Events.EventTypes.for_EventTypesStorage.given;
+
+public class an_event_types_storage : Specification
+{
+    protected static readonly EventTypeId _eventTypeId = new("event-type-with-first-generation");
+    protected static readonly EventTypeGeneration _firstGeneration = EventTypeGeneration.First;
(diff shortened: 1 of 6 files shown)


woksin:
### Reviewer notes
Found via an event-storage bug-hunt (no prior issue). The MongoDB half is significant — it breaks event-type migrations on the default backend.

- **Mongo `HasFor` (HIGH):** returned `schemas.Count == 1` — pure document-existence, ignoring the `generation` argument (the SQL analog correctly checks `Schemas.ContainsKey(generation)`). `RegisterEventTypes` relies on `HasFor(id, gen)` semantics: new-generation detection (`if (!await HasFor(id, gen))`) never fired, so `EventTypeGenerationAdded` wasn't appended, and `ValidateSchemaNotChanged` then compared the real gen-2 schema to a gen-1 fallback and threw `EventTypeSchemaChanged`, rejecting the migration. The existing unit spec passed only because it *mocks* `HasFor` with the correct behavior (`generation.Value == 1`), so real Mongo storage went untested. Fix mirrors SQL: `Schemas.ContainsKey(generation.ToString())`, null-generation defaulting to `First` (an existing type always has gen-1, preserving today's behavior for first-generation callers).
- **SQL `GetFor` cold-cache (MED):** the DB-fallback path used parameterless `ToKernel()` (hard-codes `EventTypeGeneration.First` + `Schemas.First()`), discarding the requested generation — while the cache-hit branch and Mongo use `ToKernel(generation)`. On a cold cache (multi-silo where this silo never populated the generation) the wrong-generation schema was used on hot paths (append serialization, immediate projections). Fix: `ToKernel(generation)`.

**Verification (local; CI blocked repo-wide by the pre-existing IdentityModel skew):** before/after — Mongo `should_not_have_the_second_generation` fails (Expected False, Actual True) without the fix, passes after; SQL fails all 3 (returns gen-1) without, passes after. New specs: Mongo `for_EventTypesStorage/when_checking_has_for` (substituted `IMongoCollection`, no live Mongo needed), SQL `when_resolving_schema_for_a_generation` (SQLite in-memory). Builds 0/0; Mongo specs 2/2, SQL EventTypes storage specs 9/9.

woksin:

PR #3504, opened 2026-07-23
Surface jobs storage failures instead of returning empty

woksin:
## Fixed

- Job monitoring no longer hides backend failures as an empty result. When the jobs storage fails, observing jobs now surfaces the error to the caller instead of an empty stream, and both observing jobs and reading job steps log the failure — so the Workbench no longer silently shows an empty list on a real fault

### DIFF
--- a/Source/Kernel/Core.Specs/Services/Jobs/for_Jobs/given/RecordingLogger.cs
+++ b/Source/Kernel/Core.Specs/Services/Jobs/for_Jobs/given/RecordingLogger.cs
@@ -1,0 +1,35 @@
+// Copyright (c) Cratis. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.Extensions.Logging;
+
+namespace Cratis.Chronicle.Services.Jobs.for_Jobs.given;
+
+/// <summary>
+/// A recording <see cref="ILogger{TCategoryName}"/> test double that captures the levels it was asked to log at.
+/// </summary>
+/// <typeparam name="T">The category type the logger is for.</typeparam>
+/// <remarks>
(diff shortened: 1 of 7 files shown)


woksin:
### Reviewer notes
Found via a gRPC-services bug-hunt (no prior issue). The `Jobs` service took no `ILogger`, and on a failed storage `Catch`, `ObserveJobs` returned `Observable.Empty` and `GetJobSteps` returned `[]` — both indistinguishable from "genuinely no jobs/steps", and entirely unlogged (a Workbench showing "no jobs" on a real backend fault). The sibling `GetJob` surfaces its failure via `OneOf<Job, JobError>`, and `Recommendations.ObserveRecommendations` uses no `Catch` (so faults propagate) — those are the propagate-not-swallow exemplars.

Fix: inject `ILogger<Jobs>` (updated the one manual DI registration in `ChronicleServerSiloBuilderExtensions`); on failure, `TryGetException` + log at Error; `ObserveJobs` returns `Observable.Throw(exception)` (client can distinguish fault from empty), `GetJobSteps` logs then returns `[]` (its `IEnumerable<JobStep>` return has no error channel, so the contract is kept). Log messages follow the co-located `JobsLogMessages` `[LoggerMessage]` convention (mirroring `ReducersLogging`/`ReactorsLogging`).

**Verification (local; CI blocked repo-wide by the pre-existing IdentityModel skew):** new specs use a hand-written `RecordingLogger` double (`ILogger<Jobs>` can't be NSubstitute-proxied — `Jobs` is internal in a strong-named assembly). Before the fix: 4/5 facts fail (observe completes empty; neither branch logs); after: 5/5 pass. Core + Core.Specs build 0/0; Jobs-filtered specs 100 pass (1 pre-existing unrelated skip).

woksin:

PR #3505, opened 2026-07-23
Fix unit-of-work commit leak, concurrency scope, and current handling

woksin:
## Fixed

- A unit of work whose commit fails (for example a transient connection error during the append) no longer leaks — it is always removed from the manager and the current-unit reference is cleared, instead of accumulating and leaving a stale current unit
- Optimistic concurrency is no longer silently disabled when more than one event is added to the same event source and a later event carries no concurrency scope — an already-set scope is now kept
- Completing one unit of work no longer clears the current-unit reference when a different unit has since become current

### DIFF
--- a/Source/Clients/DotNET.Specs/Transcations/for_UnitOfWork/given/a_unit_of_work.cs
+++ b/Source/Clients/DotNET.Specs/Transcations/for_UnitOfWork/given/a_unit_of_work.cs
@@ -15,6 +15,7 @@ public class a_unit_of_work : Specification
     protected IEventSequence _eventSequence;
     protected bool _onCompletedCalled;
     protected IEnumerable<EventForEventSourceId> _eventsAppended;
+    protected IDictionary<EventSourceId, ConcurrencyScope> _concurrencyScopesAppended;
     protected AppendManyResult _appendResult;
 
     void Establish()
@@ -35,6 +36,7 @@ public class a_unit_of_work : Specification
             .Returns(callInfo =>
             {
                 _eventsAppended = callInfo.Arg<IEnumerable<EventForEventSourceId>>();
+                _concurrencyScopesAppended = callInfo.Arg<IDictionary<EventSourceId, ConcurrencyScope>>();
                 return _appendResult;
             });
     }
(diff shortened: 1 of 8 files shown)


woksin:
### Reviewer notes
Found via an AggregateRoot/unit-of-work bug-hunt (no prior issue). All three are in the client `UnitOfWork`/`UnitOfWorkManager`. NOTE: PR #3489 also edits `UnitOfWork.cs` (adds OpenTelemetry tracing spans in Commit/Rollback) — different lines, a trivial rebase when both land.

- **Commit leak (HIGH):** `Commit()`'s `_isCommitted = true; _onCompleted(this)` ran after the awaited `Perform()` (gRPC append + event-type resolution + serialization). On a throw those lines were skipped, so `UnitOfWorkManager.UnitOfWorkCompleted` never ran — the unit stayed in the process-wide `ConcurrentDictionary<CorrelationId, IUnitOfWork>` and the static AsyncLocal `Current` kept pointing at it (unbounded leak for any direct `Begin`+`Commit` not wrapped in `using`; the AspNetCore middleware disposes in its catch, but the public API allows direct use). Fix moves completion into a `finally`; the exception still propagates.
- **Concurrency scope clobber (MED/HIGH):** `WithConcurrencyScope` is a last-wins setter and `Perform` drops `NotSet`-scoped sources, so a second event on the same source with no scope reset an earlier real scope to `NotSet` → the append ran with no concurrency check (silent lost-update). Fix: a `NotSet` never overrides an already-set scope (a real scope still overrides an earlier one).
- **Current wipe (MED):** `UnitOfWorkCompleted` cleared the AsyncLocal `Current` unconditionally, wiping a *different* unit that became current while this one was live. Fix guards with `ReferenceEquals(_current.Value, unitOfWork)`; dictionary removal stays unconditional.

Left `Rollback().IsSuccess` (a design-ambiguous "success after rollback") untouched — flagged for your call.

**Verification (local; CI blocked repo-wide by the pre-existing IdentityModel skew):** before/after — with the 4 source files reverted, exactly the 6 discriminating facts fail (2 per bug); after, `for_UnitOfWork` 58 pass, 0 fail (3 new specs + an extended `given` capturing the `AppendMany` concurrency-scopes arg). DotNET client + Specs build 0/0.

woksin:

Issue #1417, opened 2024-09-19
There are lots of magic TimeSpan.From*** in the codebase, many of them should be configurable


woksin:

woksin:
Triage question — this has an empty body. The title states the observation but not the decision, and the decision is the work. @woksin @einari:

- **Which ones.** "Many of them should be configurable" is the judgment call. A `TimeSpan.From*` guarding an internal retry loop and one bounding a client-visible request are different things, and only the second is obviously configuration. Is there a rule we want here — e.g. anything a *consumer* can observe the effect of becomes an option, anything purely internal stays a constant?
- **Where the configuration lands.** `ChronicleOptions` / `chronicle.json`, per-observer configuration, or the existing `TimeSpanFactory`? Note `TimeSpanFactory.DefaultTimeout()` already exists and is used by the integration specs, so there may be a seam to build on rather than a new one to invent.
- **Scope.** Kernel only, or clients too?

Without at least the first of those this cannot be picked up — the mechanical part (finding them) is easy, deciding which ones to expose is the whole task. Happy to produce an inventory of the call sites grouped by what they guard, if that would help you make the call.

woksin:
## Correction — a first batch already landed

My comment above asked which timeouts should become configurable, without noting that some already have. #3507:

> Configuration options for previously-hardcoded operational timeouts — connected-client intervals, the webhook HTTP resilience pipeline, the SQL live-query polling interval, the event-queue depletion wait, and the subscription-ready timeout — each defaulting to its current value (#1417)

So the pattern question is partly answered by precedent: those five are all things a **consumer can observe the effect of**, each kept backward-compatible by defaulting to its current value. That is a usable rule, and it is the one I was asking for.

The question that remains is narrower than I put it: **is the issue done?** Either the remaining `TimeSpan.From*` sites are deliberately internal — in which case this closes and the rule above is worth writing down in `csharp.md` — or there is a known set still to expose, in which case listing them would make this finishable.

@einari @woksin — I'm happy to produce an inventory of the remaining call sites grouped by what they guard, so the close/keep call can be made against evidence rather than a grep count.

woksin:
The consumer-facing timeouts were made configurable in #3507, and the Orleans response timeout followed in 6ca157503. The remaining hard-coded `TimeSpan` values in the Kernel are internal.

PR #3506, opened 2026-07-23
Fix read-model paging returning a short page

woksin:
## Fixed

- Paging a read-model query with a skip that is not a multiple of the take now returns the full requested number of instances instead of a short page. It also no longer throws when the take is zero

PR #3495, opened 2026-07-23
Show failed partition details in the Workbench

woksin:
## Added

- The Workbench failed-partitions view now shows a per-partition error glimpse on hover and a details pane on click, listing each attempt's timestamp, sequence number, error messages and stack trace (#1436)

PR #3491, opened 2026-07-23
Make hardcoded operational timeouts configurable

woksin:
## Added

- Configuration options for operational timeouts that were previously hardcoded — connected-client revise/reservation-TTL/observe/keep-alive/stale-threshold intervals, the webhook HTTP resilience pipeline (retry delay, circuit-breaker sampling/break durations, request timeout, test timeout), the SQL live-query polling interval, the event-queue depletion wait, and the subscription-ready timeout — each defaulting to its current value (#1417)

PR #3507, opened 2026-07-23
Consolidated non-Workbench fixes and improvements

woksin:
## Added

- Configuration options for previously-hardcoded operational timeouts — connected-client intervals, the webhook HTTP resilience pipeline, the SQL live-query polling interval, the event-queue depletion wait, and the subscription-ready timeout — each defaulting to its current value (#1417)
- The ability to rename an identity's display name, from kernel storage through gRPC to the client SDK, keyed by subject so it never affects the PII encryption key (#1684)
- React to read model changes with `IReadModelReactor`: methods named `Added`, `Modified` or `Removed` are invoked by convention with the changed read model (a single instance or a collection), an optional `EventContext`, and services, and may return events as side effects (#3359)
- `[Materialized]` opts a read model into materialized change tracking, deducing the change type by comparing successive materialized windows (#3359)

## Changed

- Read model change notifications now carry the change type and the causing event's sequence number, occurred time and correlation id (#3359)

## Fixed

- Releasing a coarse `[PII]` collection for an erased subject no longer crashes — it reads as an empty collection instead of throwing and potentially quarantining an observer after right-to-erasure
- Projection-backed observable read-model queries keep emitting after a connection drop, instead of silently freezing on stale data after the first reconnect
- Appending a batch of events no longer risks an infinite retry loop that permanently wedges an event sequence, or a sequence-number gap and phantom tail when the storage append throws
- The SQL event-store backend no longer risks an infinite retry loop when appending at a sequence number that is already taken
- Username and email lookups on the SQL backend are now case-insensitive, matching the in-memory and MongoDB backends
- Waiting for observers to catch up no longer reports success against an observer that has never handled any events; event and read-model read cursors are now disposed when an enumeration error occurs
- Event seeding no longer permanently loses events when applying them to an event store's namespaces fails partway through, and no longer silently skips events when an append fails
- Adding a new generation to an existing event type now works on the MongoDB backend, and resolving an event type by a specific generation from the SQL database on a cold cache returns that generation's schema
- Job monitoring surfaces backend failures instead of returning an empty result
- A unit of work whose commit fails is no longer leaked; optimistic concurrency is no longer silently disabled when a later same-source event carries no scope; and completing one unit no longer clears the current-unit reference of a different unit
- Paging a read-model query with a skip that is not a multiple of the take now returns the full requested window instead of a short page, and no longer throws on a zero take
- Child-level `[RemovedWith]` (and `[RemovedWithJoin]`) on a `[ChildrenFrom]` collection now removes only the targeted keyed child — including nested grandchild collections — instead of removing the whole read-model document or the parent item

## Security

- Token revocation now actually revokes. Previously every bulk revocation operation returned success without revoking anything, so rotating a leaked client secret, removing an application, or signing a user out everywhere left issued tokens valid until natural expiry; token pruning also no longer deletes still-valid tokens


woksin:
### Reviewer notes

Consolidates the **14 non-Workbench** open PRs into one, superseding them: #3490, #3491, #3493, #3494, #3497, #3498, #3499, #3500, #3501, #3502, #3503, #3504, #3505, #3506. Each individual PR has its own reviewer comment explaining how the fix was found and verified; those are being closed in favour of this one. The two **Workbench** PRs (#3495 failed-partition details, #3496 event-table filtering) and the earlier consolidated PR #3489 (which also touches the Workbench) are intentionally left separate.

All 14 branches merged cleanly (disjoint file sets). The merge is behaviour-identical to the individual PRs, with one addition below.

**A cross-branch break the individual builds missed:** #3491 (configurable timeouts) added a required `IOptions<ChronicleOptions>` to the `Webhooks` / `EventStoreSubscriptions` / `ConnectionService` constructors, and #3504 (jobs logging) added a required `ILogger<Jobs>` to `Jobs`. Those services are also constructed manually in the `Testing` and `XUnit.Integration` in-process harnesses (`TestingServices.cs`, `ChronicleConnection.cs`), which neither PR's scoped build compiled — so **#3491 and #3504 individually would fail a full CI build**. This consolidation updates those call sites to pass default options / a null logger (one commit: "Update test-harness service construction for new dependencies").

**Verification (local; CI is blocked repo-wide by a pre-existing IdentityModel transitive version skew — `IDX00001` in `Integration/Api`/`Server`, unrelated to these changes: pin `Microsoft.IdentityModel.Protocols`, `Protocols.OpenIdConnect`, and `System.IdentityModel.Tokens.Jwt` to `8.20.0` in `Directory.Packages.props` to unblock):** full-solution single-node Release build **0 warnings / 0 errors**; combined specs green — Core.Specs 1712, DotNET.Specs 1879, Storage.Sql.Specs 61, Storage.MongoDB.Specs and Storage.InMemory.Specs and Server.Specs green per the individual PRs, and Testing.Specs **231** (confirms the harness change works at runtime, since the whole scenario suite runs through it).

woksin:
### Second consolidation pass — folded in three more non-Workbench PRs

This PR now also carries (originals closed in favor of this one):

- **#3402** — `IReadModelReactor` feature (kernel + client + docs + integration specs)
- **#3492** — child-level `[RemovedWith]` routing fix in model-bound projections (+ specs)
- **#3480** — SetFrom-bridged child-collection projection spec

**Conflict resolutions worth a look:**
- `Source/Kernel/Protobuf/readmodels.proto` — union-merged: kept main's `ReadModelChangeset` fields 1–5 plus #3402's `ReadModelChangeType` enum and fields 6–9 (generator-emitted; regenerated on Release).
- `Source/Clients/DotNET/ReadModels/ReadModelWatcher.cs` — combined #3494's reconnect callbacks (`onError`/`onCompleted` → `ResetForReconnect`) with #3402's enriched OnNext body (builds the change `EventContext`, emits change type).
- `Source/Kernel/Core/Setup/ChronicleServerSiloBuilderExtensions.cs` — union of #3491's `IOptions<ChronicleOptions>` on Webhooks/EventStoreSubscriptions and #3402's new `IProjectionChangesetMediator` + `ILocalSiloDetails` on ReadModels.
- Dropped #3492's stale `Microsoft.IdentityModel.*` 8.20.0 pinning (`Directory.Packages.props` + `Server.csproj`) — main resolves the IDX00001 skew at 8.21.0 transitively; keeping #3492's pins caused `NU1010`.

**Local verification (single-node Release):** build 0 errors / 0 warnings (only the two pre-existing `CHR0025` in `TestApps/SimpleConsole`). Specs: DotNET.Specs 1940, Core.Specs 1715 (+1 skip), Testing.Specs 231 — all 0 failures.

PR #3512, opened 2026-07-24
Fix ReactorScenario dependency resolution and add Services + seeding

woksin:
# Summary

`ReactorScenario<T>` now resolves reactor handler-method dependencies the way the production reactor pipeline does, and the in-process scenario helpers gain matching ergonomics — register dependencies through a `Services` collection, seed read models directly, and assert the side effects a reactor returns — so testing reactors and read models no longer requires hand-building a service provider or mocking the event store.

## Added

- `ReactorScenario<T>.Services` collection (with logging registered by default) and an `Action<IServiceCollection>` constructor for registering a reactor's dependencies.
- `Given.ForEventSourceId(id).ReadModel(...)` to seed read-model handler-method parameters in a reactor scenario.
- `ReactorScenario<T>.ShouldHaveProduced<T>()` / `ShouldNotHaveProduced<T>()` and a `Produced` list to assert the events and commands a reactor returns as side effects, with no mocked event store.
- `ReadModelScenario<T>.Services` collection for registering reducer and projection dependencies without supplying a full `IServiceProvider`.
- `CannotActivateReactorForScenario`, thrown with the name of the dependency that could not be resolved when a reactor cannot be constructed.

## Changed

- A `ReactorScenario<T>` created without an explicit `IServiceProvider` now builds one from `Services` (standard dependency injection with logging registered) instead of using `DefaultServiceProvider`. A reactor that relied on `DefaultServiceProvider` constructing an unregistered concrete dependency must now register it.

## Fixed

- `ReactorScenario<T>` now resolves reactor handler-method parameters — injected services and materialized read models — instead of failing to resolve them.


woksin:
### Reviewer context

**Why the fix:** production wires `ReactorInvoker` with the service provider and arguments resolver (`Reactors.cs:448`); `ReactorScenario` passed neither, so reactor handler-method parameters that are services (or materialized read models) threw `CannotResolveReactorMethodArgument` in specs. No spec covered that path — only side-effect-return specs existed.

**Behavior change to sign off on:** a `ReactorScenario<T>` created with no explicit `IServiceProvider` now builds one from `Services` (standard DI + `AddLogging()`), replacing `DefaultServiceProvider`. This is why a reactor injecting `ILogger<T>` now activates by default. I tried a composite (Services + `DefaultServiceProvider` fallback) to keep the old conjuring behavior and **reverted it** — `DefaultServiceProvider`'s "construct any type" semantics break `EventStoreForTesting`, which relies on standard "null-for-unregistered" DI. Net: a reactor that leaned on `DefaultServiceProvider` conjuring an *unregistered concrete* dependency must now register it.

**Verification:** `Cratis.Chronicle.Testing` + `Cratis.Chronicle.Testing.Specs` build clean in Debug **and** Release (0 warnings); full `Testing.Specs` suite 234/234, incl. the new `for_ReactorScenario` specs covering method-param DI, read-model seeding, and the activation-failure path.

**Docs:** the snippet compile-check and `npm run check` run in the Documentation aggregator repo, not here — I verified every API in the new snippet against source by hand but did not run the docs CI locally.

**Deliberately not in this PR:**
- Asserting a reactor's *returned side-effect events* still needs the manual `IEventStore`/handler boilerplate. The best-DevEx version (default the handlers + expose `AppendedEvents` / `ShouldHaveAppended<T>`) needs the kernel cursor read path and its own specs — a well-scoped follow-up.
- An analyzer (Arc side) flagging `IServiceProvider`/`IEventStore` injected into an `IReactor`.

woksin:
### Follow-up pushed — produced-side-effect assertions

Added `ShouldHaveProduced<T>` / `ShouldNotHaveProduced<T>` (+ `Produced`) to `ReactorScenario`, so a return-style reactor's side effects are asserted directly rather than through a mocked `IEventStore`. This is the bridge between the pure-function test and the scenario test — the "then" is now the same vocabulary in both.

**Design:** a default recording `IReactorSideEffectHandlers` captures whatever the reactor returns, so it works uniformly for **events and commands** (the Arc-side command return) with no kernel cursor reads and no Arc coupling. It engages only when no explicit handlers are supplied; explicit-handler specs are unaffected.

**Verification:** Testing + Testing.Specs clean Debug + Release (0 warnings); full Testing.Specs 237/237, incl. new specs for a produced command and a produced (unwrapped cross-stream) event.

woksin:
### Follow-up pushed — ReadModelScenario Services parity

Added a mutable `Services` collection to `ReadModelScenario`, matching `CommandScenario`/`ReactorScenario`, so a reducer/projection with a constructor dependency can have it registered without supplying a whole `IServiceProvider`. Provider and event-store construction are deferred (lazy) so `Services` can be configured after construction; read-model seeds are buffered and applied on first use.

**Backward compatibility:** deliberately conservative here (50 existing `for_ReadModelScenario` spec files). When `Services` is untouched and no explicit provider is given, the existing `DefaultServiceProvider` activation is preserved unchanged — so no existing spec changes behavior. Registering anything in `Services` opts into a standard provider (with logging). This is a smaller default change than `ReactorScenario` made, on purpose.

**Verification:** Testing + Testing.Specs clean Debug + Release (0 warnings); full Testing.Specs 238/238, incl. the 206 existing `for_ReadModelScenario` specs (unchanged) and a new spec for a reducer resolving a registered service.

Deferred (own PR): adding `Given` to `CommandScenario` — the remaining surface divergence — and the Arc analyzer for `IServiceProvider`/`IEventStore` injected into an `IReactor`.

PR #3513, opened 2026-07-25
Give Kotlin compile daemon enough heap to build generated contracts

woksin:
# Summary

The Kotlin client failed to publish because `:compileKotlin` ran out of heap on the large generated protobuf/gRPC sources; this raises the Gradle and Kotlin daemon heaps.

## Fixed

- Kotlin client now publishes to Maven Central without the compiler running out of memory on the generated contracts.

### DIFF
--- a/.github/workflows/kotlin-build.yml
+++ b/.github/workflows/kotlin-build.yml
@@ -1,0 +1,65 @@
+---
+name: Kotlin Build
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  DOTNET_VERSION: "10.0.x"
+
+"on":
+  workflow_dispatch:
+  pull_request:
+    branches:
+      - "**"
+    paths:
+      - "Source/Clients/Kotlin/**"
+      - "Source/Kernel/Contracts/**"
+      - "Source/Kernel/Protobuf/**"
+      - "Source/Tools/ProtoGenerator/**"
+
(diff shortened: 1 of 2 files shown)


woksin:
### Reviewer context

**Failure (post-merge Publish for #3512):**
```
> Task :compileKotlin FAILED
  Backend Internal error: Exception during IR lowering
  File being compiled: .../contracts/TokenProvider.kt
  Caused by: java.lang.OutOfMemoryError: GC overhead limit exceeded
```
The `publishAndReleaseToMavenCentral` step compiles first and never reached publishing — credentials/signing were not involved.

**Root cause:** `Source/Clients/Kotlin/` had no `gradle.properties`, so the Kotlin compile daemon ran on the default heap and couldn't lower the IR for the large generated contracts. Not caused by #3512 (that PR touched only `Cratis.Chronicle.Testing` + docs, not the Kotlin client, contracts, or proto generation) — consistent with the publish workflow's history of a different leg failing per release.

**Fix:** `org.gradle.jvmargs=-Xmx4g` + `kotlin.daemon.jvmargs=-Xmx4g` (the compile OOM is in the Kotlin daemon; both set for robustness across execution strategies). Comfortable on the 16 GB `ubuntu-latest` runner.

**Verification:** none locally — no JDK on my machine, and the OOM is runner-memory-dependent, so it can't be reproduced here. This must be verified by CI / the publish workflow. Two standard Gradle properties, syntactically trivial.

PR #3514, opened 2026-07-25
Fix coverage-merge-and-cache checkout race on branch deletion

woksin:
# Summary

Internal CI fix — no user-facing effect. Reviewer context in the PR comment below.


woksin:
## Root cause

\`coverage-merge-and-cache\` runs last in \`dotnet-build.yml\`, after \`specs\` + the full \`integration\` matrix + \`integration-api\` + \`mongodb\` (often 10-30+ min). Its checkout step resolved the ref by PR branch **name**. If the PR merged (and its branch was deleted) before this tail job ran, checkout failed:

\`\`\`
##[error]A branch or tag with the name '<branch>' could not be found
\`\`\`

turning an otherwise fully green run red — noise, not a real failure. Reproduced twice in the last day: run [30151923156](https://github.com/Cratis/Chronicle/actions/runs/30151923156) (this repo's own \`feat/reactorscenario-di-and-seeding\`, PR #3512) and run 30099173633 (\`fix/nuget-pack-ide0301\`). Not a required status check, so it wasn't blocking merges, but it made otherwise-green PRs look broken after the fact.

This is a different bug from the one fixed in \`b0379c541\` (the \`<PR#>/merge\` ref rebase issue) — same job, different race.

## Why remove the commit-back step instead of just guarding it

This workflow only triggers on \`pull_request\` / \`workflow_dispatch\` — it never runs on \`push\`. So there was never a stable branch context to safely commit generated \`coverage-data.js\` stats onto in the first place; every attempt targeted a transient, soon-to-vanish PR branch. Release's \`publish.yml\` already owns the authoritative \`coverage-data.js\` updates on merges to main. The merge/cache/log-summary steps are unaffected — only the checkout ref (now the immutable commit SHA) and the git commit+push step changed.

## Verified

- \`python3 -c "import yaml; yaml.safe_load(open('.github/workflows/dotnet-build.yml'))"\` — YAML parses cleanly.
- Diff reviewed: -18/+1 lines, scoped entirely to the \`coverage-merge-and-cache\` job.

PR #3515, opened 2026-07-25
Fix in-memory storage reads and inverted event stream id filtering

woksin:
# Summary

Reads against the in-memory storage provider returned nothing, and event stream id filtering was inverted in every storage provider.

## Fixed

- Reads against the in-memory storage provider returned no events — `GetFromSequenceNumber` came back empty, `GetTailSequenceNumber` reported `Unavailable` and `GetCount` reported zero, even though the events were stored and appends reported success.
- Event stream id filtering was inverted: reading a named stream returned events from every stream, while a general read returned only default-stream events.
- Redaction, revision and replacing generation content did nothing on the in-memory storage provider, reporting success without changing the stored event.
- Targeting a sequence number that holds no event now reports `NoEventAtSequenceNumber` instead of a bare `InvalidOperationException`.


woksin:

**One change reaches beyond the in-memory provider.** `EventStreamId.IsDefault` was declared `this != Default` while its documentation says it reports whether the id *is* the default. Its only call sites — three in the MongoDB provider, three in SQL — all read `IsDefault == false` to mean "not the default, so narrow on it", so correcting the property is what makes those six call sites behave as documented. This does change observable read behavior for named event streams in MongoDB and SQL. Labeled `patch` since it restores documented intent, but it is worth a second opinion.

**Internal changes not in the release notes:** the testing harness carried a fork of the kernel's in-memory event sequence storage; the two had drifted into the same read-filter defect. The fork is deleted and the harness now aliases the kernel provider, so fixes land once. Also excludes the container publish output from the Server project's content glob, which was making a Debug build followed by a Release build fail locally with MSB3030.

**Verification.** Full solution builds clean in Debug and Release with zero warnings. 4,714 specs pass across Concepts, Storage, Storage.InMemory, Storage.MongoDB, Storage.Sql, Core, Services, Testing and DotNET. Against real MongoDB: 162 pass in `for_EventSequence`, 238 pass / 6 skipped across `for_ReadModels`, `for_Reducers`, `for_Reactors` and `for_PIIManager`.

The read-filter guard specs were checked by reintroducing the defect — all five fail with it present and pass with it fixed.

Out-of-process integration was not run: the locally built server image fails to start on this machine with a missing `Cratis.Screenplay` assembly, which reproduces identically on unmodified specs and is unrelated to this change. CI covers that matrix.

PR #3518, opened 2026-07-26
Kernel performance and correctness release

woksin:
# Summary

Performance and correctness pass over the kernel's append, observer, and projection paths.

## Added

- Benchmarks for projection, reducer and reactor throughput, projection replay, and passive-versus-materialized read models.
- Benchmarks comparing a single silo against a two-silo cluster, including a split-role topology.
- Startup warning when a clustered deployment is configured against local-only storage.
- Clustering configuration documentation covering environment requirements, multi-node compose, and placement behavior.
- `Jobs.StepCheckpointFlushInterval`, bounding how long a job step may hold an unpersisted checkpoint.
- `ReadModels.GuardSinkWritesOnWatermark`, the kill switch for the crash-recovery write guard.
- `Events.ConstraintsVersionCheckInterval`, bounding how often an active event sequence checks for constraint changes.

## Changed

- Reduced storage round-trips and CPU on the append, observer, and projection paths through event-type schema caching, debounced observer-state persistence, memoized projection key resolution, striped projection locking, subscribed-queue routing, and coalesced observer dispatch.
- Appends no longer block on slow observer consumption; a full delivery queue spills to the observer catch-up path instead of applying backpressure.
- Event batches are routed only to queues whose subscribed observers care about the batch's event types, and queues prefer co-location with their event sequence.
- Projections whose key collapses several partitions onto one read-model document now handle all partitions through a single activation.
- Documentation no longer claims projections and reducers are idempotent across a crash; that guarantee covers a deliberate replay only.
- `SaveChanges`, `ProjectionPipelineManager` and `ReducerPipelineFactory` each take an additional constructor parameter, which is source-breaking for anyone constructing them directly.

## Fixed

- Projection read models could be corrupted in a multi-silo cluster when replay and catch-up ran the read-modify-write cycle concurrently for a partition-collapsing key.
- Accumulating projections and reducers no longer double-count events re-delivered by crash recovery, except where a resumed catch-up re-cuts the original batch boundaries.
- Every replay job reported `CompletedWithFailures`, because benign step-start outcomes were counted as failures and no failed-step record was written.
- Resuming a job reported a healthy running step as a failed resume, and reported genuine resume failures as successes.
- `[Unique]` could reject a value nothing held on in-memory storage, because changing a constrained value never released the previous one.
- Joined projection changes after the first were silently dropped, leaving a stale read model with no error.
- Concurrent resolution of event-store and namespace storage could serve a stale event-type schema or leave two sinks for one read model with divergent replay state.
- A cached event-type schema was never invalidated when the event type changed, so properties present on the wire but missing from the stale schema were silently dropped between silos.
- A job step that completed after it was no longer tracked turned a successful completion into an opaque job error.
- Job lookups by type and status always returned nothing against MongoDB, so running and stopped jobs were invisible and were restarted rather than resumed.
- An observer could stop processing events permanently when preparing a catch-up never produced a job.
- An exception while entering or leaving a state left an observer unable to transition again.
- A durable append could be reported as failed after the event was persisted, inviting a retry that appended it twice and skipping the unique-constraint index.
- Queue assignment used a per-process randomized hash, so a restart could re-map observers to different queues.
- Grain timers scheduled for replay-on-definition-change were retained for the lifetime of the activation.
- Constraint index resolution during `AppendMany` stamped each event's index with the wrong sequence number.
- Constraints registered after an event sequence or observer had activated were not enforced.
- Server crashed on startup against MongoDB due to unsupported LINQ in job index creation.
- Concurrent changes to the in-memory read-model sink could race.
- Duplicate-key failures when writing a unique-constraint index surfaced as a raw storage exception.

## Removed

- `IEventTypesStorage.Populate()`, obsolete now that schemas are read through the cache.
- `ObserverState.HandledEventCountPerPartition`, replaced by a dedicated store.
- `IServer.ReloadState()` and its `host.proto` RPC, non-functional since inception.

## Security

- Stop logging raw `Authorization` header values.


woksin:
## Reviewer context

Not release notes — verification evidence, deferred decisions, and the things a reviewer should push back on.

### Verification at `ef6f4c4d1`

Clean `dotnet clean` followed by full Debug **and** Release builds, both zero warnings / zero errors. 17 spec projects, **~6,056 specs passing**, 1 pre-existing skip:

| Suite | Passed | Suite | Passed |
|---|---|---|---|
| Infrastructure.Specs | 572 | Server.Specs | 15 |
| Concepts.Specs | 105 | Services.Specs | 41 |
| Core.Specs | 2002 (1 skip) | DotNET.Specs | 1957 |
| Storage.Specs | 56 | DotNET.CodeAnalysis.Specs | 186 |
| Storage.InMemory.Specs | 82 | AspNetCore.Specs | 28 |
| Storage.Sql.Specs | 70 | Connections.Specs | 57 |
| Storage.MongoDB.Specs | 566 | Testing.Specs | 265 |
| Storage.Vault.Specs | 10 | GrpcCodeGenerator.Specs | 17 |
| **CLUSTER-CI (Integration/Clustering)** | **27/27** | | |

`Services.Specs` is not in `Chronicle.slnx` and was invoked explicitly — a gap that has hidden a compile break in this repo before.

**One local failure, proven pre-existing:** `Storage.AzureKeyVault.Specs` fails 10/10 with `AuthenticationException: UntrustedRoot`. The identical failure reproduces at base commit `4c4cae7c5`, 222 commits before this work, and this PR changes zero Azure Key Vault files. It is TLS interception on the machine that ran the suite, not a regression. CI should exercise it in a clean environment.

### Still owed

- **The three-way benchmark (`BASE → WPS → FOLLOWUPS`) has not been re-run** with the follow-up fixes included. `BENCHMARK-kernel-perf.md` requires it, and by its own rule a follow-up that regresses a number beyond the measured spread is a merge blocker for that item. The benchmark fixture binds fixed host ports and the harness varied ~30% across sessions, so it needs a quiet machine and all three sides measured back-to-back; the machine was not quiet. **The append numbers quoted in the description are from the pre-follow-up measurement.** The watermark guard adds a conditional clause on the sink write path and a 5s timer per job-step activation, which is where a regression would show.
- **The out-of-process integration matrix was not run locally**, for the same reason. It includes a new crash oracle (`for_AccumulatorRedeliverySafety`) that has never been executed. It may fail for harness reasons rather than product reasons: if out-of-process observers run in the test-host silo rather than the kernel container, restarting the container creates no redelivery window. The spec carries an anti-vacuity assertion (`RedeliveredSequenceNumbers > 0`) specifically so that situation fails loudly instead of passing green.

### One decision this PR deliberately does not make

`ConstraintBatchClaims` has the batch-level analogue of the in-memory `[Unique]` bug — and unlike that one it is **shared kernel code, so it affects all three storage providers including MongoDB**. Within one batch: source X claims value A, X moves to B, and a later event from source Y claiming A is rejected — a violation on a value nobody will hold once the batch lands.

This PR ships the **characterization specs only**. The non-release is not intentional (`CompleteDurableAppend` applies index updates as a sequential fold in event order, so the post-batch index is deterministic and the rejection is provably false), but the fix needs a decision: the defect has two manifestations and only one is reachable from the accumulator. `UniqueConstraintValidator` consults batch claims only when storage already said yes, so where the holder's original claim was persisted by an *earlier* append, fixing it requires the accumulator to override a storage rejection. Post-append index updates run inside a swallow-and-log `try/catch`, so a failed release plus a successful claim would leave two owners on one value — and on MongoDB the second write would throw inside that swallow, turning a false rejection into a silently divergent index. Worse than the bug.

The specs pin current behavior with XML docs naming exactly what to flip. See `FOLLOWUP-PLAN.md` §1b NEW-1.

### Semver

Labeled `minor`. `ISink.ApplyChanges` gained a **default interface method** overload, so third-party sinks compile and behave unchanged. But three public kernel types — `SaveChanges`, `ProjectionPipelineManager`, `ReducerPipelineFactory` — each gained a required constructor parameter, which is source- and binary-breaking for anyone constructing them directly. They are DI-composed plumbing with no expected external construction site and the break is a compile error rather than silent behavior change, which is why `minor` is defensible, but it is a real break and is called out in the description rather than implied away. Separately, `ReducerPipeline`'s new parameter defaults to `true`, so a hand-constructed instance is silently opted into the guard.

### Known-weakest verification

Every item ran implementer → independent read-only adversarial verifier → gate → merge, with each behavior change carrying a spec proven to bite (revert the fix, confirm red with the recorded message, restore, confirm green). **One exception:** the `GetNamespace` commit was cleared by the orchestrator rather than a fresh adversarial pass, because its verifier died on a session limit. Its claims were re-verified by reading and its bite proof re-run (red 3/3, green 8/8), but that is confirmation of a known argument rather than an independent attack. It is the weakest verification in the set and worth a reviewer's attention.

The accumulator fix required a fix-forward round: the first version turned a bulk-queued delete followed by a re-create into **silent row loss**, because an unflushed delete made the read return a doomed document, the write was then guarded update-only, and the ordered bulk ran the delete before the update matched nothing. Every suite was green when that bug was present — no spec exercised a delete and re-create inside one bulk window.

woksin:
## Follow-up: the two items the first comment listed as owed

Both are now resolved, and one changed the contents of this PR.

### The crash oracle was executed, failed honestly, and has been reverted out

`for_AccumulatorRedeliverySafety` ran for the first time here and **failed on all five infrastructure configs**. The
failure is the anti-vacuity assertion:

```
should_have_redelivered_at_least_one_event    0 should be greater than 0
```

On out-of-process + MongoDB: 3 passed, 2 failed — and the 3 that passed did so **vacuously**. Nothing was ever
redelivered, so the guard was never exercised. Without `RedeliveredSequenceNumbers > 0` this would have been a green
run proving nothing, which is exactly what that assertion exists to prevent.

Cause: the scenario appends 500 events with awaited round-trips, so the observers reach the tail *during* the append
loop; by the time the kernel restarts there is nothing to resume. A longer timeout or a lower checkpoint interval
does not fix it — the restart has to land mid-catch-up. Separately, the MSSQL cell fails earlier and differently
(`Health check mssql ... Unhealthy` → `SqlException: ... The login failed`) — a second, independent harness gap.

Skipping the five facts was tried and rejected: `[Fact(Skip = ...)]` raises **xUnit1004** ×5, and this repo's
standard is zero warnings with no suppression. A test that never exercises its fault should not ship, so the oracle
is reverted (`d1d3b080e`, `fcc136e4d` — `git revert`, no history rewrite) and tracked as `FOLLOWUP-PLAN.md` NEW-8.
Recover it with `git show 517f94ca2`.

**Consequence, stated plainly: G2 remains open.** The watermark guard is proven by in-process sink specs — 13 facts
that go red when the fix is reverted, independently re-run by the reviewing agent — but **not** end to end against a
real kernel crash.

### Benchmark: no regression beyond the measured spread

One harness (this PR's tip), one session, serial, both sides free of `NA` rows:

| Benchmark | before P0 `35c74ac04` | with P0 `ef6f4c4d1` | Δ |
|---|---|---|---|
| AppendSingleEvent | 3.554 ms ± 0.267 | 3.762 ms ± 0.487 | +5.9% |
| AppendManyEvents / 10 | 4.560 ms ± 0.406 | 4.746 ms ± 0.493 | +4.1% |
| AppendManyEvents / 100 | 10.209 ms ± 0.996 | 10.246 ms ± 1.939 | +0.4% |
| AppendManyEvents / 1000 | 52.053 ms ± 9.171 | 59.112 ms ± 11.188 | +13.6% |

Every delta is inside the error bars, and allocation is unchanged. **Sensitivity limit:** the same image measured
`/1000` at 52.1, 55.3 and 62.7 ms across three runs — a ±10% noise floor on identical code, wider than any delta
above, with a co-tenant workload on the machine. This rules out a gross regression, not a ~10% one.

A full BASE→P0 comparison turns out to be **unobtainable with a single harness**: a newer harness is refused by the
BASE server (`IncompatibleServerException: RenameIdentity missing`), and the older harness is refused by this PR's
server (401 — auth changed in this release). BASE→WPS→`35c74ac04` remains valid in `BENCH-CHAIN-summary.txt`.

### CI

93/93 jobs green after one re-run. The `specs` failure on the first attempt was a `Storage.MongoDB.Specs` flake in
`for_EventSequenceStorage` — it passed at the parent commit, passes 5/5 locally three consecutive times at the exact
failing SHA, is untouchable by an `Integration/Client`-only revert, and passed on re-run.

`PR Prerelase` is red, but it is red on every recent branch in this repository including dependabot's, and its jobs
report as skipped at check level.

woksin:
## Follow-up batch: four more fixes, and two problems left open on purpose

Four additional lanes, each implementer → independent adversarial verifier → gate → merge. All accepted.

### What landed

**The appended-event schema cache is now reachable and evicted.** It was evicted only on lookup failure, so a
cached schema outlived any registration that changed it, while the invalidation seam reached storage only. On the
receive side the converter iterates *schema* properties, so content present on the wire but missing from a stale
schema was silently dropped. Reachability is narrow — cross-silo only, and `ValidateSchemaNotChanged` blocks the
mainline path; the bypasses are the DEVELOPMENT image's `DisableValidation`, `RegisterSingle` (no validation,
Workbench-reachable), and `Register` with empty generations.

Worth recording: **the planned fix would not have worked.** `AddCompleteSerializer<T>` registers four independent
singleton descriptors, so `GetRequiredService<AppendedEventSerializer>()` returns an instance Orleans never uses —
proven with an isolated DI probe (`distinct instances = 4`). Injecting the serializer into the grain service, as
planned, would have evicted an empty dictionary on a dead object and passed every spec. The cache is now its own
singleton that the grain service can address. `AddCompleteSerializer` has gained a doc note so the next stateful
codec does not fall into the same hole.

**Two more testing-harness storage duplicates removed** (closed-stream and unique-event-type constraints), replaced
by the kernel types. Neither had drifted — but a third copy of the same shape already had, which is what produced the
`[Unique]`-violation-against-nothing bug fixed earlier in this PR. Verified by exact diff, and both directions
proven: mutating the kernel storages makes the new specs fail, and the same specs pass identically against the old
copies and the kernel types.

**Four hardening fixes.** A completed job step no longer tracked threw `KeyNotFoundException` out of a `finally`,
which was swallowed and turned a *successful* completion into an opaque job error (reachable after `Resume()`, which
reloads only non-completed steps, or on a duplicate notification). A null-dictionary guard in the job step reload.
A doubled `Consolidate()` on every resolved join — the verifier proved the pass reaches a fixed point in one
application for all inputs, not just the exercised ones. And a child projection context built by passing the join
flag into the slot declared for the initial-state requirement, discarding both it and `JoinKey`.

### Left open deliberately

**Failed-step accounting** — investigated, **not fixed**, and the investigation overturned its own premise. The
claim that all three providers filter `CompletedWithFailure` out of `GetJobSteps(Failed)` is **wrong for SQL**: SQL's
`MoveToFailed` hardcodes `Status = Failed` and on read the column overwrites the deserialized status. **So the
providers disagree** — SQL already behaves one way, MongoDB and InMemory the other — and
`Integration/Kernel/.../and_job_step_fails.cs:57`, which asserts `CompletedWithFailure`, **would fail against SQL
today**. Picking a semantic is not safe without a decision: `CompletedWithFailure` is documented as distinct from
"failed internally", the `Start` state machine branches on the pair, and both are serialized wire contract values in
`jobs.proto`. The unambiguous part — make SQL persist the real status — is needed under either choice. See
`FOLLOWUP-PLAN.md` NEW-2.

**`ConstraintBatchClaims`** remains untouched, as described in the previous comment.

### Verification

Clean `dotnet clean` + Debug and Release rebuilds, 0 warnings / 0 errors. 17 suites green; `Core.Specs` 2002 → 2019
and `Testing.Specs` 265 → 272, reconciling exactly against the individual lanes. CLUSTER-CI 27/27. CI: **93/93 jobs
green**.

`Storage.AzureKeyVault.Specs` still fails locally with `UntrustedRoot`; that reproduces at base commit `4c4cae7c5`
and this branch changes no Azure Key Vault files.

### Semver note

The `minor` label is correct **only because this branch ships as a single release**.
`EventTypesCacheGrainService`'s public primary constructor gained a parameter, which is a public-signature break; it
is safe solely because the type is unreleased (`git tag --contains` on its introducing commit is empty, and it exists
only on this branch). Cherry-picked onto a released line, that change would be `major`.

PR #3522, opened 2026-07-28
Add CHR0038 cross-subject [PII] join analyzer and validate MongoDB database names

woksin:
# Summary

Cross-subject `[PII]` joins are now a build error, and an illegal MongoDB database name fails with a Chronicle diagnostic.

## Added

- `CHR0038`: a model-bound `[Join<TEvent>]` copying a `[PII]` value out of a stream that is not the read model's own compliance subject is now an error at build time, instead of freezing every partition of the projection at runtime with an OpenSSL `oaep decoding error`.
- `InvalidDatabaseName` when the event store name and namespace compose a name MongoDB rejects, naming the event store, the namespace, and the offending character instead of the driver's bare `Invalid namespace specified`.

## Changed

- Documented Chronicle's MongoDB database naming, including that read models of the default namespace use the bare event store name while event sequences suffix every namespace.


woksin:
Reviewer context.

**Related:** Cratis/Arc#2383 is the other half of the database-naming story. This PR makes an illegal namespace fail loudly and centralizes the naming rule; the Arc PR fixes the actual empty-read bug, where a tenant named `Default` resolved `<database>+Default` while Chronicle materializes the default namespace into the bare `<database>`.

**CHR0038 scope.** Covers the model-bound `[Join<TEvent>]` attribute only; the fluent `builder.Join<T>(_ => _.On(...))` form is not analyzed, which the docs page states. PII detection mirrors `PIIMetadataProvider` (attribute on the declaring type, the property, the positional-record parameter, or the property's own `ConceptAs<T>` type). Subject resolution mirrors Chronicle's order: `[Subject]`, `[Key]`, an `EventSourceId<T>`-derived value, then `Id`. It deliberately does not fire when the join has no `on`, or when `on` is the read model's own subject.

Severity is Error to match CHR0034/CHR0035 — the pattern throws at runtime and has no correct use.

**Internal refactor.** `DatabaseNames` replaces three inline string interpolations across `Database.cs`, `EventStoreDatabase.cs`, and `EventStoreNamespaceDatabase.cs`. Composed names are byte-identical to before; the only behavior change is the added validation.

**Verification.** Full solution Release and Debug: 0 errors, 0 warnings (46 projects each). Client-snippet validation compiles. Specs: CodeAnalysis 193, Storage.MongoDB 577, Storage 56, Storage.InMemory 82, Core 2021, DotNET 1957 — all pass.

PR #3523, opened 2026-07-28
Serve the Workbench UI from the web root when it is not embedded

woksin:
# Summary

The Workbench UI has been unreachable in every published container image since 16.3.0.

## Changed

- The Kernel serves the Workbench UI from the web root next to the binary as well as from the copy embedded in `Cratis.Chronicle.Workbench`, using whichever is present
- The warning logged when no Workbench UI can be found now names the web root that was checked

## Fixed

- The Workbench UI is reachable again in the published container images - `https://localhost:35000` serves the Workbench instead of returning 401

### DIFF
--- a/Source/Kernel/Server.Specs/for_WorkbenchUI/given/a_file_provider.cs
+++ b/Source/Kernel/Server.Specs/for_WorkbenchUI/given/a_file_provider.cs
@@ -1,0 +1,24 @@
+// Copyright (c) Cratis. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.Extensions.FileProviders;
+using Microsoft.Extensions.Primitives;
+
+namespace Cratis.Chronicle.Server.for_WorkbenchUI.given;
+
+/// <summary>
+/// An <see cref="IFileProvider"/> holding exactly the files it is given, so specs can express what each
+/// candidate location does and does not contain without touching the file system.
+/// </summary>
+/// <param name="files">The files the provider holds.</param>
+public class a_file_provider(params string[] files) : IFileProvider
+{
+    public static a_file_provider Empty() => new();
+
+    public IDirectoryContents GetDirectoryContents(string subpath) => NotFoundDirectoryContents.Singleton;
+
+    public IFileInfo GetFileInfo(string subpath) =>
+        files.Contains(subpath) ? new an_existing_file(subpath) : new NotFoundFileInfo(subpath);
+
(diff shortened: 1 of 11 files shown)


woksin:
## Reviewer context

**Root cause.** `2d7358bd5` (first released in 16.3.0) switched the Kernel to serve the Workbench exclusively from a `ManifestEmbeddedFileProvider` over `Cratis.Chronicle.Workbench`. That assembly only embeds the UI when `Source/Workbench/wwwroot` exists as it is compiled — and `Source/Workbench/wwwroot` is gitignored, so the `dotnet-x64`/`dotnet-arm64` publish jobs (which do a bare checkout and never download the `workbench` artifact) always produce an assembly without it. The Docker jobs *do* download that artifact, but they copy it to `/app/wwwroot`, which the Kernel had stopped reading.

With no static files and no `MapFallbackToFile`, `/` matches no endpoint, and `SetFallbackPolicy(RequireAuthenticatedUser)` applies to endpoint-less requests too — hence 401 rather than 404.

**Why not fix it in CI instead.** Making the kernel publish jobs depend on the `workbench` job would work, but it puts the frontend build on the release critical path and adds ~31 MB to every image (the size the embedded copy adds to `Cratis.Chronicle.Workbench.dll`) on top of the `/app/wwwroot` copy the images already carry. Serving from either location keeps both delivery shapes valid, which is what the NuGet package (embedded) and the images (web root) actually need.

**Not made a hard startup failure.** A Kernel-only deployment legitimately ships no Workbench — `Docker/Local` is exactly that — so absence stays a warning, now naming the path that was checked.

**Verification.** Built the development image from this branch the way CI does — kernel published with `Source/Workbench/wwwroot` absent (confirmed: resulting `Cratis.Chronicle.Workbench.dll` is 65 KB with no embedded manifest, identical to the shipped one), then the image built with the UI copied in from the published `workbench` artifact:

| | before (`cratis/chronicle:latest-development`) | after (this branch) |
|---|---|---|
| `GET /` | 401 | 200, Workbench `index.html` |
| SPA deep route | 401 | 200 |
| hashed JS asset | 401 | 200 |
| log | "Workbench UI is not embedded" | none |

Negative control — same image run with an empty `/app/wwwroot` — still starts, still answers 401, and logs `... neither embedded in this build nor present in the web root '/app/wwwroot' ...`.

Bisected the regression to the release boundary: `16.2.1-development` → 200, `16.3.0-development` → 401.

Solution builds clean in Debug and Release (46 projects, 0 warnings); `Server.Specs` 23/23 pass.

PR #3524, opened 2026-07-28
Cover fluent joins in CHR0038 and diagnose the subject mismatch at runtime

woksin:
# Summary

Closes the two gaps left by CHR0038: the fluent join spelling, and what happens when a cross-subject `[PII]` join reaches runtime anyway.

## Added

- `ComplianceMetadataActionFailed` replaces the bare `InvalidOperationException` when applying or releasing compliance metadata fails. On a failed release it names the property, the subject the value was released under, and explains that the value was encrypted under a different subject, pointing at CHR0038 — rather than surfacing only the underlying OpenSSL `oaep decoding error`.

## Changed

- `CHR0038` now also reports the fluent `builder.Join<TEvent>(_ => _.On(...))` form, not just the model-bound `[Join<TEvent>]` attribute. A `[PII]` value counts as reaching the read model through an explicit `.Set(...).To(...)` mapping or through AutoMap matching an identically named property.


woksin:
Reviewer context. Follow-up to #3522, which introduced CHR0038 and explicitly documented these two gaps.

**Fluent analyzer scope.** `FluentCrossSubjectPiiJoinAnalyzer` matches `Join<TEvent>` on `IProjectionBuilder<TReadModel, TBuilder>`, taking the event type from the method's type argument and the read model from the containing type's. It stays quiet when the join has no `On()` (a child join keys on the parent's `identifiedBy`, so there is no separate stream to cross into), when `On()` names the read model's own subject, and when the `.On(...)` / `.Set(...).To(...)` lambdas are anything other than a simple `p => p.Member` — the same conservatism `RedundantSetToWithMatchingNamesAnalyzer` uses.

**Internal refactor.** The descriptor plus the shared PII detection and subject resolution move from `CrossSubjectPiiJoinAnalyzer` into `CrossSubjectPiiJoin`, so both analyzers report one diagnostic from one implementation. Behavior of the model-bound analyzer is unchanged — its six existing specs pass untouched. The message format is now neutral about the spelling ("The join with 'X' on 'Y'") so it reads correctly from either analyzer.

**Runtime diagnostic.** The subject-mismatch hint is attached only when the action is `release` *and* the error chain contains a `CryptographicException`. A cryptographic failure on `apply` means encryption itself failed, which is a different problem, and there is a spec asserting the hint is absent there.

**Labeled `patch`** — no public API is removed or changed. `ComplianceMetadataActionFailed` is new, but it replaces an `InvalidOperationException` that was never part of a contract.

**Verification.** Full solution Release and Debug: 0 errors, 0 warnings (46 projects each). Client-snippet validation compiles. Specs: CodeAnalysis 199 (193 + 6 new fluent), Core 2038 (+9 new compliance), DotNET 1969, Storage.MongoDB 577 — all pass.

PR #3525, opened 2026-07-28
Fix PR prerelease workflow that never ran a job

woksin:
## Changed

- PR prerelease Docker images are built for `linux/amd64` only. Releases still ship both architectures.
- The PR prerelease pipeline no longer re-runs when a pull request title or description is edited.

## Fixed

- PR prerelease NuGet packages and Docker images are published again. Nothing had been published since 2026-04-19.
- gRPC contract compatibility is compared against the pull request's base branch instead of against the pull request itself, so breaking changes are reported again.

## Removed

- Automatic pull request description updates for detected gRPC breaking changes.

### DIFF
--- a/.github/workflows/grpc-compatibility.yml
+++ b/.github/workflows/grpc-compatibility.yml
@@ -1,7 +1,11 @@
 name: gRPC Contract Compatibility
 
+# Scoped to this workflow by name rather than `github.workflow`. In a called
+# reusable workflow `github.workflow` resolves to the *caller*, so the generic
+# form produced a group identical to the caller's and the two runs contended
+# for the same slot.
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
+  group: grpc-compatibility-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
 
 on:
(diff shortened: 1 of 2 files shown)


woksin:
Reviewer context.

**Evidence.** Over the last 30 days `pull-requests.yml` produced 319 failed runs, 0 successful, 1 cancelled. I checked the last 60 runs individually: not one executed a single job — every job in every run reports `skipped`, and the nested `grpc-compatibility` job never appears in the job list at all. Onset correlates with `0e8436113` (2026-04-19), which re-added a workflow-level `concurrency` block to `grpc-compatibility.yml` four minutes after `03d80ec54` had removed it.

**Mechanism.** The nested call never starts, so `release` (which needed it) skips and the whole downstream graph skips while the run reports failure. The concurrency group used `github.workflow`, which in a *called* reusable workflow resolves to the caller — producing a group identical to the caller's own. I have not reproduced this in isolation, so treat the collision as the most probable mechanism rather than a confirmed one. The fix does not depend on it being right: the nested call is removable on independent grounds, since under `pull_request_target` it checked out the base branch and compared `main` against `main`.

**Internal changes not in the PR body:** `actions/cache` v3 → v4 (v3 is being retired), and the deprecated `::set-output` replaced with `$GITHUB_OUTPUT`.

**Two judgment calls, both easy to reverse:**

1. `.github/scripts/update-pr-description.sh` is now unreferenced. I left it and its README entry rather than deleting them. Re-attaching it to the standalone gRPC workflow would restore the intended behavior, but that workflow runs on `pull_request` and would need `pull-requests: write` — a feature decision I did not want to fold into a bug fix.

2. This un-breaks a pipeline dormant for three months, so PR prereleases start running, and costing, again. The amd64-only change is what keeps that from landing as a large bill. If the dormancy was intentional, deleting the workflow is the cheaper option and I am happy to switch.

**Not verified:** I could not exercise this end to end — the failure only reproduces on a real `pull_request_target` event. The job graph parses and has no dangling `needs`, but the first real run on this PR is the actual test.

woksin:

PR #3526, opened 2026-07-28
Run standalone benchmarks on main instead of on every pull request

woksin:
## Changed

- Benchmarks run on `main` rather than on every pull request. Pull request benchmarking is unchanged and still runs as part of .NET Build & Integration.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -3,9 +3,13 @@ name: Benchmarks
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
+# Benchmarks on pull requests are covered by the `benchmarks` job in
+# dotnet-build.yml, which reuses that workflow's cached build instead of
+# compiling again from scratch. This workflow tracks the series on main,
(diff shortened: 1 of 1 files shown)


woksin:
Reviewer context.

The same benchmark suite already runs on pull requests as the `benchmarks` job in `dotnet-build.yml`, which reuses that workflow's cached build instead of compiling from scratch. This workflow was doing identical work a second time on every PR touching `Source/**` — 204 runs in the last 30 days at ~12.7 job-minutes each, roughly 2,600 job-minutes.

Moving it to `main` also lines its `benchmark-cache-${{ github.sha }}` key up with the one `publish.yml` restores, which it could never match from a pull request.

One non-obvious detail: `comment-on-alert` posts a *commit* comment on push events rather than a PR comment, which needs `contents: write` — so permissions changed from `contents: read`. Regression alerting on pull requests is unaffected and still comes from the `dotnet-build.yml` job.

**Not verified:** `push`-triggered behavior only exercises after merge to main. The workflow parses and the trigger/permissions are as intended, but the first real signal is the first push to main.

woksin:

PR #3527, opened 2026-07-28
Cover one storage backend per pull request, all of them nightly

woksin:
## Added

- A nightly scheduled run of .NET Build & Integration covering every storage backend.

## Changed

- Client integration specs cover MongoDB in-process and out-of-process on pull requests. SQLite, PostgreSQL and SQL Server coverage moves to the nightly run and to manual dispatch.

### DIFF
--- a/.github/scripts/generate-integration-matrix.py
+++ b/.github/scripts/generate-integration-matrix.py
@@ -2,14 +2,18 @@
 """Generate the GitHub Actions matrix for the client integration specs.
 
 Every client integration namespace -- a directory under ``Integration/Client``
-that contains ``[Fact]`` tests -- runs against every infrastructure
-configuration (runtime mode x storage backend). Large namespaces are split
+that contains ``[Fact]`` tests -- runs against a set of infrastructure
+configurations (runtime mode x storage backend). Large namespaces are split
 into several shards that run as separate, parallel jobs so the slowest single
 job no longer gates the whole workflow.
 
 Sharding only changes how the work is distributed: every shard of a namespace
 together covers exactly the same tests as the un-sharded namespace would, so
-coverage across all five infrastructure configurations is preserved.
+coverage across the selected infrastructure configurations is preserved.
+
+Which configurations are selected depends on ``--all-providers``. Pull requests
+run the default backend only; the nightly schedule and manual dispatch run all
+of them.
 
 Balancing is by ``[Fact]`` count across test classes. Each spec file declares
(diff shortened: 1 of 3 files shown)


woksin:
Reviewer context.

**What drove this.** `.NET Build & Integration` is ~91% of Chronicle's CI compute, and `client-integration-specs` is ~85% of that workflow — about 78% of all CI time in the repo. Measured over 5 completed runs: 1,067 job-minutes total, 903 of them in `client-integration-specs`. One run is 93 jobs / 42 minutes wall clock.

The matrix was 16 namespace-shards × 5 infrastructure configs = 80 jobs per PR. Each job pays checkout + SDK setup + cache restore + Docker Hub pre-pull before its first test; on a sampled shard that fixed cost was ~56s against 171s of tests.

**Verified locally:**

```
generate-integration-matrix.py                  -> 32 jobs (inprocess/mongodb 16, outofprocess/mongodb 16)
generate-integration-matrix.py --all-providers  -> 80 jobs (all five configs, 16 each)
```

I also simulated the workflow's env expansion: `ALL_PROVIDERS=""` yields 32 jobs, `ALL_PROVIDERS="true"` yields 80. Both workflow files parse and have no dangling `needs`.

**Why nightly and not on merge to main.** `main` took 195 merges in the last 30 days, so a `push` trigger would run the full 80-job matrix ~195 times and save nothing. A nightly run costs ~30 full matrices a month.

**Rough arithmetic:** 193 PR runs × ~181 job-min today ≈ 35,000 job-min. After: 193 × ~72 plus 30 nightly × ~181 ≈ 19,400 — about 45% less.

**Deliberate omissions:**

- `SHARD_COUNTS` is untouched. It is tuned against observed wall-clock, and consolidating shards would trade developer wait time for machine time. Worth revisiting separately — several shards run under a minute of tests for ~56s of setup.
- The per-shard `docker pull mongo` runs even on the in-process config, which may not need it. I left it rather than guess at the fixture's requirements.

**The tradeoff to weigh:** a storage-provider regression specific to SQLite, PostgreSQL or SQL Server would now surface on the nightly run rather than on the PR that introduced it. If a given PR touches storage-provider code, `workflow_dispatch` with all backends is the escape hatch. If you would rather have that automatic, a path filter or a label-triggered opt-in would be straightforward to add — say the word and I will wire it up.

**Not verified:** I could not execute the matrix on GitHub's runners; the generator output and expansion logic are verified locally, but the first real run on this PR is the actual test.

woksin:

PR #3528, opened 2026-07-28
Upgrade actions/cache from v3 to v4

woksin:
## Changed

- Continuous integration caching uses `actions/cache@v4`.


woksin:
`pull-requests.yml` also still uses `actions/cache@v3` — it is deliberately excluded here because #3525 already upgrades it as part of unbreaking that workflow. Merging either order is fine; this PR will not conflict with it.

woksin:

PR #3529, opened 2026-07-28
Reduce CI compute and fix the dead PR prerelease pipeline

woksin:
## Added

- A nightly scheduled run of .NET Build & Integration covering every storage backend.

## Changed

- Client integration specs cover MongoDB in-process and out-of-process on pull requests. SQLite, PostgreSQL and SQL Server coverage moves to the nightly run and to manual dispatch.
- Benchmarks run on `main` rather than on every pull request. Pull request benchmarking is unchanged and still runs as part of .NET Build & Integration.
- PR prerelease Docker images are built for `linux/amd64` only. Releases still ship both architectures.
- The PR prerelease pipeline no longer re-runs when a pull request title or description is edited.

## Fixed

- PR prerelease NuGet packages and Docker images are published again. Nothing had been published since 2026-04-19.
- gRPC contract compatibility is compared against the pull request's base branch instead of against the pull request itself, so breaking changes are reported again.

## Removed

- Automatic pull request description updates for detected gRPC breaking changes.

### DIFF
--- a/.github/workflows/dotnet-build.yml
+++ b/.github/workflows/dotnet-build.yml
@@ -15,6 +15,10 @@ env:
 
 on:
   workflow_dispatch:
+  # Nightly full-coverage run. Pull requests cover the default storage backend
+  # only, so this is where sqlite, postgresql and mssql get exercised.
+  schedule:
+    - cron: "0 2 * * *"
   pull_request:
     branches:
       - "**"
@@ -209,6 +213,9 @@ jobs:
       # the Docker image still comes from dotnet-build.
       dotnet-cache-key: ${{ needs.dotnet-build-development.outputs.dotnet-cache-key }}
       chronicle-image: ${{ needs.dotnet-build.outputs.chronicle-image }}
+      # Every storage backend on the nightly schedule and on manual dispatch;
+      # pull requests run the default backend only.
+      all-providers: ${{ github.event_name != 'pull_request' }}
 
   integration-api:
     runs-on: ubuntu-latest
--- a/.github/workflows/pull-requests.yml
+++ b/.github/workflows/pull-requests.yml
@@ -14,8 +14,10 @@ env:
   WORKBENCH_CACHE: "web-cache-${{ github.sha }}"
 
 on:
+  # `edited` is deliberately absent: it fires on every PR title/description
+  # edit, which re-ran this whole prerelease pipeline for a text change.
   pull_request_target:
-    types: [opened, edited, synchronize, reopened]
+    types: [opened, synchronize, reopened]
     branches:
       - "**"
     paths:
@@ -23,34 +25,8 @@ on:
       - ".github/workflows/**"
 
 jobs:
-  grpc-compatibility:
-    permissions:
-      contents: read
-    uses: ./.github/workflows/grpc-compatibility.yml
-    with:
-      base-ref: ${{ github.event.pull_request.base.ref }}
-
-  update-pr-with-breaking-changes:
-    runs-on: ubuntu-latest
-    needs: [grpc-compatibility]
-    if: needs.grpc-compatibility.outputs.has-breaking-changes == 'true'
-    permissions:
-      pull-requests: write
-    steps:
-      - name: Checkout code
-        uses: actions/checkout@v4
-
-      - name: Update PR description
-        env:
-          GH_TOKEN: ${{ github.token }}
-        run: |
-          .github/scripts/update-pr-description.sh \
-            "${{ github.event.pull_request.number }}" \
-            "${{ needs.grpc-compatibility.outputs.breaking-changes }}"
-
   release:
     runs-on: ubuntu-latest
-    needs: [grpc-compatibility]
     outputs:
       version: ${{ steps.release.outputs.version }}
       publish: ${{ steps.release.outputs.should-publish }}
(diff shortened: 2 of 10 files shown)


woksin:
Reviewer context. This combines #3525, #3526, #3527 and #3528, which are now closed. Each is preserved as its own commit behind a merge commit, so any one of them can be reverted independently.

## 1. The PR prerelease pipeline has been dead since April (was #3525)

Over the last 30 days `pull-requests.yml` produced **319 failed runs, 0 successful, 1 cancelled**. I checked the last 60 runs individually: not one executed a single job — every job in every run reports `skipped`, and the nested `grpc-compatibility` job never appears at all. No PR prerelease NuGet package or Docker image has been published since 2026-04-19.

Onset correlates with `0e8436113` (2026-04-19), which re-added a workflow-level `concurrency` block to `grpc-compatibility.yml` four minutes after `03d80ec54` had removed it.

The nested call never starts, so `release` (which needed it) skips and the whole downstream graph skips while the run reports failure. The group used `github.workflow`, which in a *called* reusable workflow resolves to the caller — producing a group identical to the caller's own. I have not reproduced that in isolation, so treat the collision as the most probable mechanism rather than a confirmed one. The fix does not depend on it: the nested call is removable on independent grounds, since under `pull_request_target` it checked out the base branch and compared `main` against `main`.

The standalone gRPC check had a separate defect — the `inputs` context is only populated for `workflow_call`, so on a plain `pull_request` the baseline checkout fell back to the triggering ref and compared each PR against itself.

## 2. Integration matrix scoping (was #3527)

`.NET Build & Integration` is ~91% of Chronicle's CI compute and `client-integration-specs` is ~85% of that — roughly 78% of all CI time in the repo. Measured over 5 runs: 1,067 job-minutes, 903 in `client-integration-specs`. One run was 93 jobs / 42 minutes wall clock.

Verified locally, then confirmed live on #3527's own CI run: **32 client-integration jobs instead of 80**, 44 total instead of 93, and the reduced set passes.

`main` took 195 merges in the last 30 days, so a `push` trigger would have run the full matrix ~195 times and saved nothing — hence nightly.

Rough arithmetic: 193 PR runs × ~181 job-min ≈ 35,000 job-min today; after, 193 × ~72 plus 30 nightly × ~181 ≈ 19,400 — about 45% less.

## 3. Benchmarks deduplication (was #3526)

The same suite already ran on PRs as the `benchmarks` job in `dotnet-build.yml`, which reuses that workflow's cached build. The standalone workflow was doing identical work a second time on every PR touching `Source/**` — 204 runs in 30 days at ~12.7 job-min each, ~2,600 job-minutes. Moving it to `main` also lines its cache key up with the one `publish.yml` restores.

Non-obvious detail: `comment-on-alert` posts a *commit* comment on push rather than a PR comment, so permissions changed from `contents: read` to `contents: write`.

## Internal changes, not in the PR body

`actions/cache` v3 → v4 across `pull-requests.yml`, `publish.yml` and `workbench-build.yml` (v3's backing service is being retired), and the deprecated `::set-output` replaced with `$GITHUB_OUTPUT`.

## Judgment calls, all easy to reverse

1. `.github/scripts/update-pr-description.sh` is now unreferenced. I left it and its README entry rather than deleting them. Re-attaching it to the standalone gRPC workflow would restore the intended behavior, but that workflow runs on `pull_request` and would need `pull-requests: write` — a feature decision I did not want to fold into a bug fix.
2. This un-breaks a pipeline dormant for three months, so PR prereleases start running, and costing, again. The amd64-only change is what keeps that from landing as a large bill. If the dormancy was intentional, deleting the workflow is cheaper and I am happy to switch.
3. `SHARD_COUNTS` is untouched. It is tuned against observed wall-clock, and consolidating shards would trade developer wait time for machine time. Worth revisiting separately — several shards run under a minute of tests against ~56s of fixed setup.
4. A storage-provider regression specific to SQLite, PostgreSQL or SQL Server will now surface on the nightly run rather than on the PR that introduced it. `workflow_dispatch` with all backends is the escape hatch; a label-triggered opt-in would be straightforward to add if you would rather it were automatic.

## Not verified

**The prerelease fix cannot be validated by this PR's own CI.** `pull_request_target` reads the workflow from the *base* branch, so main's broken copy is what runs here — the green checks are other workflows. It only proves out after merge. That is inherent to the bug, not something I can work around.

.github/workflows/dotnet-build.yml
```diff
@@ -209,6 +213,9 @@ jobs:
       # the Docker image still comes from dotnet-build.
       dotnet-cache-key: ${{ needs.dotnet-build-development.outputs.dotnet-cache-key }}
       chronicle-image: ${{ needs.dotnet-build.outputs.chronicle-image }}
+      # Every storage backend on the nightly schedule and on manual dispatch;
+      # pull requests run the default backend only.
+      all-providers: ${{ github.event_name != 'pull_request' }}
 
   integration-api:
     runs-on: ubuntu-latest
```
woksin, line 214: @einari This maybe needs your opinion

.github/workflows/pull-requests.yml
```diff
@@ -298,7 +269,10 @@ jobs:
           builder: ${{ steps.buildx.outputs.name }}
           context: .
           file: ./Docker/Development/Dockerfile
-          platforms: linux/amd64,linux/arm64
+          # amd64 only for PR prereleases. arm64 has to be emulated through
+          # QEMU on GitHub-hosted runners, which dominated this job's runtime;
+          # the release build in publish.yml still ships both architectures.
+          platforms: linux/amd64
```
woksin, line 275: Not sure if this is okay

PR #3530, opened 2026-07-28
Stop CHR0038 reporting fluent joins that AutoMap cannot carry

woksin:
# Summary

CHR0038 reported a fluent join that carries nothing, breaking correct builds.

## Fixed

- `CHR0038` no longer reports a fluent join when AutoMap cannot carry the value across — `[NoAutoMap]` on the read model, and `[NoAutoMap]` on the target property of a positional record, were both still reported. Since the rule is an error, each was a build break on correct code.


woksin:
Reviewer context. Follow-up to #3522 / #3524.

**This is live.** #3524 merged at 16:40:48Z and v16.9.2 published 25 seconds later, so both false positives are in the current Latest release. CHR0038 is error severity, so a consumer on 16.9.2 with a fluent join on a read model that turns AutoMap off will not build.

**Two distinct causes, both proven by spec.** Each new spec fails against the previous commit and passes here:

- `[NoAutoMap]` on the read model type was never checked at all — only the fluent `.NoAutoMap()` call was.
- `[NoAutoMap]` on the target property *looked* handled but was not. `GetMembers` surfaces each member of a positional record twice, once as the property and once as the constructor parameter, and an attribute written without an explicit target lands only on the parameter. Filtering out the entry that carried the attribute left the other entry behind, so the name stayed mappable. The fix excludes the name outright rather than the entry.

The second one is the interesting failure: it was written as if fixed, compiled, and had no spec, so nothing contradicted it. It was only caught by writing the spec and running it against the prior commit first.

**Not affected:** the model-bound `CrossSubjectPiiJoinAnalyzer`. A `[Join<TEvent>]` attribute declares its mapping explicitly via `JoinDefinitionExtensions`, so AutoMap plays no part and `[NoAutoMap]` does not suppress it.

**Verification.** Rebased on main after #3524 merged. Release build 0 errors, 0 warnings (46 projects). CodeAnalysis specs 202 pass, including the two new ones.

Issue #1395, opened 2024-09-11
Add logging messages for client reactors + reducers


einari:
Trace for the process, but when there are errors we should output the errors in logs

woksin:
## Resolved on `main` — for reducers

Landed in #3554:

> Reducer failures are written to the client log. A reducer method that throws was reported to the kernel as a failed observation without ever appearing in the client's own log.

That is the "when there are errors we should output the errors in logs" half of this issue, which was the part with real diagnostic cost — a throwing reducer previously left no trace client-side at all.

**Worth confirming before closing:** the issue title says "reactors **+** reducers", and the PR bullet names only reducers. If reactor failures already reach the client log by another path this is fully done; if not, the remaining half is narrow and worth restating rather than leaving this open as-is.

@einari @woksin — close if reactors are covered; otherwise I'd narrow the title to reactors and keep it.

einari:
Delivered by #3554 — reducer failures are now written to the client log.

PR #3534, opened 2026-07-29
Log reducer failures on the client

woksin:
## Fixed

- Reducer failures are written to the client log. A reducer method that throws was reported to the kernel as a failed observation without ever appearing in the client's own log, so the only way to see why a reducer stopped was to inspect kernel-side observer state. (#1395)


woksin:

Issue #690, opened 2023-01-19
Improve error messages from JSON Compliance Manager when releasing events


einari:
<img width="1294" alt="image" src="https://user-images.githubusercontent.com/134365/213467687-db011ea8-ba97-4ca4-b0e0-3ba919b5d791.png">

woksin:
## Resolved on `main`

Three changes in #3554 address this, and between them they replace the message in the screenshot:

> `CompliancePropertyNotFoundInSchema`, thrown when a property being applied or released has no counterpart in the schema. The message names the property path, the compliance subject, and the properties the schema does declare.

> Applying or releasing compliance metadata for a property the schema does not declare no longer fails with a bare `Sequence contains no matching element`.

> Applying or releasing compliance metadata no longer fails when a schema flattened across inheritance declares the same property name more than once.

The middle one is the specific defect this issue was filed against — a bare `Sequence contains no matching element` that named neither the property nor the subject.

@einari @woksin — safe to close.

woksin:
Closing — released in **v16.10.0**.

Delivered by #3554. Verification against `main` is in the comment above.

PR #3535, opened 2026-07-29
Improve the diagnostic when compliance handling finds no schema for a property

woksin:
## Added

- `CompliancePropertyNotFoundInSchema`, thrown when a property being applied or released has no counterpart in the schema. The message names the property path, the compliance subject, and the properties the schema does declare. (#690)

## Fixed

- Applying or releasing compliance metadata for a property the schema does not declare no longer fails with a bare `Sequence contains no matching element`. (#690)
- Applying or releasing compliance metadata no longer fails when a schema flattened across inheritance declares the same property name more than once. (#690)


woksin:

Issue #1009, opened 2023-11-10
IObserverMiddleware should have which observer it is for


einari:

woksin:
## The type this is about no longer exists

`IObserverMiddleware` is gone from the codebase. On `main`, `git grep ObserverMiddleware -- Source/` returns nothing — the only match anywhere in the repository is a line in `IssueAnalysis.md`, which is generated from this issue's own title.

So there is no interface left to add "which observer it is running for" to. Whatever replaced the middleware seam (`IObserverMiddleware` predates the current `Reactors` / `Reducers` client dispatch) either carries the observer already or is a different design question worth restating.

@einari @woksin — I'd close this as obsolete. If there is still a real ask underneath it — "an interception point in the client observer pipeline should know which observer it is for" — that is worth a fresh issue against whatever the current seam is.

woksin:
## Correction — this was implemented, not abandoned

My comment above said the type was gone and recommended closing as obsolete. The conclusion (close it) was right; **the reason was wrong**, and the right reason is better news.

The ask was implemented in #3554, under the renamed interface:

> `IReactorMiddleware.BeforeInvoke` and `AfterInvoke` overloads that receive the `ReactorId` of the reactor observing the event. Chronicle calls these; they default to the existing overloads, so a middleware written against the previous interface keeps working unchanged.

`Source/Clients/DotNET/Reactors/IReactorMiddleware.cs:40,53`:

```csharp
Task BeforeInvoke(ReactorId reactorId, EventContext eventContext, object @event) => BeforeInvoke(eventContext, @event);
Task AfterInvoke(ReactorId reactorId, EventContext eventContext, object @event) => AfterInvoke(eventContext, @event);
```

So "the middleware should know which observer it is running for" is exactly what shipped — `IObserverMiddleware` was renamed to `IReactorMiddleware` along the way, which is why my grep for the old name found nothing and I mistook a rename for a removal. No follow-up issue is needed.

@einari @woksin — close as **done**, not as obsolete.

woksin:
Closing — released in **v16.10.0**.

Delivered by #3554. Verification against `main` is in the comment above.

PR #3536, opened 2026-07-29
Tell reactor middlewares which reactor they are running for

woksin:
## Added

- `IReactorMiddleware.BeforeInvoke` and `AfterInvoke` overloads that receive the `ReactorId` of the reactor observing the event. Chronicle calls these; they default to the existing overloads, so a middleware written against the previous interface keeps working unchanged. (#1009)


woksin:

Issue #1277, opened 2024-07-28
Add an overview on Docker hub for Chronicle images


einari:

woksin:
Reopening the practical status of this: the automation shipped in 16.10.0 but the overview has **not** actually reached Docker Hub.

The `Update Docker Hub overview` step ran on the 16.10.0 release and its PATCH was rejected:

```
Sending PATCH request
##[error]Forbidden
```

`GET https://hub.docker.com/v2/repositories/cratis/chronicle` still returns an empty `description` and `full_description`. Because the step is `continue-on-error` — deliberately, so an overview problem can never fail a release — it reports success and nothing flags it.

`Docker/README.md` is in the repo and correct. What is needed is for `DOCKER_PASSWORD` to be a token with **write/admin** scope; the Docker Hub description API rejects a push-only token. Alternatively, drop the step and paste the overview into the Docker Hub UI once.

A separate small defect from the same run is fixed in #3556: the short description was 104 bytes and Docker Hub truncates at 100.

woksin:
## Re-verified today — the content shipped, the overview still has not

#3554 lists this as delivered:

> An overview on the `cratis/chronicle` Docker Hub repository covering the published tag variants, ports, quick start and the environment variables that configure the server. (#1277)

The overview exists in the repository and the workflow step runs. **Docker Hub still has neither.** Checked just now:

```
GET https://hub.docker.com/v2/repositories/cratis/chronicle/
  description      → ""
  full_description → "" (0 characters)
```

That confirms the diagnosis in my earlier comment rather than superseding it: the `Update Docker Hub overview` step is `continue-on-error`, its PATCH is rejected with `Forbidden`, and the release goes green regardless. Shipping the content did not change that, because the content was never the blocker — the credential is.

**This cannot be closed on the strength of the PR.** It needs one of:

- `DOCKER_PASSWORD` replaced with a token carrying **write/admin** scope (the description API rejects a push-only token), or
- the step dropped and `Docker/README.md` pasted into the Docker Hub UI once, which takes a minute and removes a permanently-red-but-ignored step.

@einari @woksin — this is the second release cycle where it has reported success while doing nothing. Worth deciding which of the two, since the current state is the worst of both.

PR #3537, opened 2026-07-29
Publish an overview for the Chronicle images to Docker Hub

woksin:
## Added

- An overview on the `cratis/chronicle` Docker Hub repository covering the published tag variants, ports, quick start and the environment variables that configure the server. It is published from `Docker/README.md` as part of a release. (#1277)

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -470,6 +470,20 @@ jobs:
           build-args: |
             VERSION=${{ needs.release.outputs.version }}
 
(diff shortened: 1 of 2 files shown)


woksin:
The Docker Hub description API needs a token with write scope. If DOCKER_PASSWORD is a read-only push token the step will 401 — it is marked continue-on-error so that can never fail a release, but the overview would then need pasting in the Docker Hub UI instead.

woksin:

PR #3538, opened 2026-07-29
Fix the benchmarks job failing on every pull request

woksin:
Restores the benchmarks job, which failed on every pull request. No user-facing change, so there are no release-note bullets.

### DIFF
--- a/.github/workflows/dotnet-build.yml
+++ b/.github/workflows/dotnet-build.yml
@@ -26,6 +26,10 @@ on:
       - "Source/**/*.cs"
       - "Source/**/*.csproj"
       - "Integration/**"
+      # The benchmarks job builds these, so a change here has to run it. Without this a broken benchmark
+      # project lands green and fails on the next unrelated pull request instead.
+      - "Benchmarks/**/*.cs"
+      - "Benchmarks/**/*.csproj"
       - ".github/workflows/dotnet-build.yml"
       - ".github/workflows/integration.yml"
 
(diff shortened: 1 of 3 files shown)


woksin:
Root cause: `Chronicle.Benchmarks.csproj` reached the client API transitively through `Server.csproj`, whose ProjectReference is conditioned on `ArtifactsPath` being empty. BenchmarkDotNet rebuilds the benchmark project with `ArtifactsPath` set, so that rebuild lost every `Cratis.Chronicle.*` client namespace and failed with CS0234. Nothing surfaced it: the compile errors landed inside the `Run Benchmarks` step, which still reported success, and the job only failed two steps later on `No benchmark result was found ... "Benchmarks": []`.

Reproduced locally with `dotnet build Benchmarks/Chronicle.Benchmarks -c Release /p:ArtifactsPath=...` — 20 CS0234 errors before, clean after. `--list flat` discovers all 8 benchmarks. The real end-to-end proof is this PR own benchmarks job.

Pre-existing on main and unrelated to any PR content — it also fails on #3533 (a docs link fix). Worth noting the same latent trap remains: a step that runs a build and a run in one `run:` block can report success while the build failed.

woksin:
Added the three things that let this land silently, so the next one fails where it breaks:

- **Split the step.** `Run Benchmarks` ran the build and the run in one `run:` block, so a failed build was reported as a **successful step** and only surfaced three steps later. Now `Build Benchmarks` and `Run Benchmarks` are separate.
- **`Benchmarks/**` was not in the path filter.** A change to the benchmark project never ran the job that builds it — which is how the break landed green and then failed on someone else PR. Added `Benchmarks/**/*.cs` and `Benchmarks/**/*.csproj`. This also means this PR now validates its own fix, which the first push could not.
- **Empty results now fail in the merge step**, naming the likely cause, instead of failing later as the much less obvious `no benchmark result was found`.

Verified the merge script both ways locally: exits 1 with the new message on an empty result set, exits 0 and merges normally otherwise.

woksin:

PR #3539, opened 2026-07-29
Take benchmark runs off the pull request critical path

woksin:
## Changed

- Benchmarks run on `main` rather than on every pull request. Pull requests now verify only that the benchmark projects build, which is what keeps a broken one from landing.
- Benchmark results are committed by the benchmark run on `main` instead of being carried from a pull request to the release through a cache.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -3,10 +3,9 @@ name: Benchmarks
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
-# Benchmarks on pull requests are covered by the `benchmarks` job in
-# dotnet-build.yml, which reuses that workflow's cached build instead of
-# compiling again from scratch. This workflow tracks the series on main,
-# where its cache key also lines up with the one publish.yml restores.
+# This is where the benchmark series lives. Pull requests only check that the benchmark projects still
+# build (the `benchmarks-build` job in dotnet-build.yml) - running them there added ~13 minutes to the
+# critical path for a result that could never fail the build.
 on:
   workflow_dispatch:
   push:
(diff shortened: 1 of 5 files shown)


woksin:
Measured on the run for #3538 before changing anything:

- total wall clock **19.0 min**, of which `dotnet-build` is 6.0 and `benchmarks` is 12.9 — the two are sequential (`benchmarks` needs `dotnet-build`), so 6 + 12.9 ≈ the whole run. The benchmarks job **was** the critical path.
- total compute 114.8 min across 45 jobs, of which benchmarks is ~11%.

Expected after: time-to-green drops from ~19 min to ~11 min, with ~2 min of benchmark building running in parallel off the critical path.

**What is given up:** per-PR regression comments. Detection moves to the merge onto main, where `benchmarks.yml` already runs and `comment-on-alert` posts a commit comment. Nothing is lost as a gate — `fail-on-alert` is false, so the PR job could never fail a build.

**What is deliberately kept:** the guard that a benchmark project still builds. It builds with `ArtifactsPath` set, which is the mode BenchmarkDotNet rebuilds in and precisely where the missing project reference in #3538 bit — a plain build does not catch it. Verified locally: both benchmark projects build clean that way.

**One thing worth a second pair of eyes.** Benchmark results reached the repo indirectly: the PR job wrote a cache keyed on the merge sha, and `publish.yml` restored it at release time. With no PR run there is no such cache, so `benchmarks.yml` now commits the series itself (same shape as coverage statistics, `[skip ci]`, rebase-before-push) and the release job that restored the cache is removed. That path needs `PAT_WORKFLOWS` to be available to `benchmarks.yml` — it is used the same way by `coverage-statistics` in publish.yml, but this is the first time that workflow pushes.

#3538 is merged in so the new guard can pass; the diff collapses to just the workflow changes once #3538 lands.

woksin:

PR #3540, opened 2026-07-29
Register the observer partition counts class map before any spec runs

woksin:
Removes the last reachable instance of the lazy class-map registration that caused the event-sequence spec failures. No user-facing change, so there are no release-note bullets.

### DIFF
--- a/Source/Kernel/Storage.MongoDB.Specs/Observation/for_ObserverHandledCountsStorage/when_incrementing_and_removing_counts.cs
+++ b/Source/Kernel/Storage.MongoDB.Specs/Observation/for_ObserverHandledCountsStorage/when_incrementing_and_removing_counts.cs
@@ -1,12 +1,10 @@
 // Copyright (c) Cratis. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using Cratis.Arc.MongoDB;
 using Cratis.Chronicle.Concepts.Events;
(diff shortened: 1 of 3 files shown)


woksin:
Closes the last reachable instance of the pattern behind the #3529 spec failures: a class map registered lazily behind `IsClassMapRegistered` from a spec base loses to whichever spec touches the type first, and the driver auto-registers a *default* map on first serializer lookup. `ObserverPartitionCounts` was registered that way from two separate spec classes; nothing wins that race today, but the observer-state specs serialize a graph containing these counts, so it is one spec away from behaving like the event one did.

Now uses the server own `ObserverPartitionCountsClassMap` rather than a hand-rolled copy, so it cannot drift from production.

**I first tried registering every `IBsonClassMapFor<T>` by discovery, the way `AddCratisMongoDB` does — it breaks 13 specs, and the reason is worth its own issue.** The `JobState` specs register a map that `MapMember`s `JobState.Request`, while the server `JobStateClassMap` deliberately **unmaps** it and lets `JobStateSerializer` own it. So `for_JobStateSerializer` currently exercises the serializer against a class map production never uses, and with the real one they fail on an NRE inside `JobStateSerializer.Deserialize`. That is a fidelity gap in those specs rather than something to paper over from the initializer, so I left their local registration alone and documented why. Happy to open an issue for it.

Verified: 577/577 pass, Debug and Release build clean with zero warnings.

woksin:

Issue #3541, opened 2026-07-29
for_JobStateSerializer specs run against a class map production never uses


woksin:
`Source/Kernel/Storage.MongoDB.Specs/Jobs/for_JobStateSerializer/given/a_job_state_serializer.cs` registers its own `JobState` class map that maps every member including `Request`:

```csharp
cm.AutoMap();
cm.MapMember(c => c.Id);
...
cm.MapMember(c => c.Request);
```

The server's map does the opposite. `Source/Kernel/Storage.MongoDB/Jobs/JobStateClassMap.cs` unmaps `Request` on purpose:

```csharp
// Unmap the Request property to prevent MongoDB from using discriminator-based serialization.
// The JobStateSerializer handles complete serialization/deserialization using IJobTypes.
classMap.UnmapMember(_ => _.Request);
```

So the 13 specs under `for_JobStateSerializer` exercise `JobStateSerializer` against a document shape it never sees in the silo. Swapping in the real `JobStateClassMap` makes them fail with a `NullReferenceException` from `BsonDocumentReader.ReturnToBookmark` inside `JobStateSerializer.Deserialize` — which is either a real defect in the serializer or specs written around the wrong map.

Found while removing the lazy class-map registrations in #3540, where registering every `IBsonClassMapFor<T>` by discovery (the way `AddCratisMongoDB` does on the server) was the natural fix but is blocked by this. The spec-local registration was deliberately left in place there.

Worth noting the same specs are also exposed to the registration race #3540 closes elsewhere: `Indexing/for_get_jobs_raw` and `Indexing/for_get_jobs_of_type` touch `JobState` from a different xUnit collection, so which map wins is currently down to scheduling.

woksin:
Half of this turned out to be a real defect rather than a spec artifact, fixed in #3542: a job state persisted before its request was assigned could not be read back at all — `Deserialize` returned to a bookmark that was never taken. The existing specs could not see it precisely because of the divergence this issue is about.

What remains here is the divergence itself: `for_JobStateSerializer` serializes with `_original.ToBsonDocument()` (the class map) and deserializes with the serializer, while production uses the serializer for both. They should drive the serializer in both directions and run against the real `JobStateClassMap`. Once they do, registering every `IBsonClassMapFor<T>` by discovery in the specs — the way `AddCratisMongoDB` does on the server — becomes possible, which is what #3540 wanted.

woksin:
Resolved on `release/consolidated-improvements` (#3554).

[...]

PR #3542, opened 2026-07-29
Read back a job state that has no request

woksin:
## Fixed

- Reading back a job whose request was not yet assigned no longer fails with a `NullReferenceException`, losing the status and progress the job state was written to record. (#3541)

### DIFF
--- a/Source/Kernel/Storage.MongoDB.Specs/Jobs/for_JobStateSerializer/when_round_tripping_a_job_state_without_a_request.cs
+++ b/Source/Kernel/Storage.MongoDB.Specs/Jobs/for_JobStateSerializer/when_round_tripping_a_job_state_without_a_request.cs
@@ -1,0 +1,50 @@
+// Copyright (c) Cratis. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Cratis.Chronicle.Concepts.Jobs;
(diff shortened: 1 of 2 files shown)


woksin:
Found while triaging #3541, and it turned out to be a real defect rather than the spec artifact I filed it as.

`JobState.Request` is declared `= default!`, and `JobStateSerializer.Serialize` writes the request element only `if (value.Request is not null)`. `DeserializeJobStateExceptRequest` only takes a bookmark when it *encounters* that element, so for a state persisted before its request was assigned the bookmark stays null and `Deserialize` calls `ReturnToBookmark(null)` — NRE, and the status and progress the document was written to record are lost.

The existing `for_JobStateSerializer` specs could not catch this because they serialize with `_original.ToBsonDocument()` (the class map) and deserialize with the serializer. Production uses the serializer for both. The new spec round-trips through the serializer in both directions, which is what made it visible.

This is also why registering the server `JobStateClassMap` in the specs (attempted in #3540) fails 13 specs: that map unmaps `Request`, so `ToBsonDocument()` omits it and every one of them hits this same null bookmark. #3541 stays open for the remaining half — those specs should drive the serializer rather than the class map.

Verified: the new spec fails on main with the NRE and passes after; 581/581 pass; Debug and Release clean.

woksin:

Issue #2290, opened 2025-11-23
Investigate ReadModels.Register() behavior for reactor-only applications


einari:
## Summary

The `ReadModels.Register()` method in the client library could have an early return guard that skips registration when there are no read models (projections/reducers). This was attempted in commit 2b5315f28 to prevent overwriting existing registrations during hot reload scenarios.

However, this guard causes issues for reactor-only applications (apps with only reactors, no projections or reducers) because it prevents necessary initialization of the `ReadModelsManager` grain state.

## Current Behavior

In `Source/Clients/DotNET/ReadModels/ReadModels.cs`:
```csharp
public async Task Register()
{
    var readModels = new List<IHaveReadModel>();
    readModels.AddRange(projections.GetAllHandlers());
    readModels.AddRange(reducers.GetAllHandlers());
    readModels.AddRange(additionalReadModels);
    
    // Guard was removed as a quick fix, but needs proper investigation
    // Previously: if (readModels.Count == 0) return;
    
    await _chronicleServicesAccessor.Services.ReadModels.Register(...);
}
```

### 1. Reactor-Only Applications (Current Issue)
- Applications with only reactors but no projections/reducers
- The early return prevented `ReadModelsManager.Register()` from being called
- This left the grain in an uninitialized state
- Manifested as failures in retry scenarios (3+ retries with transient services)
- **Current fix**: Removed the guard entirely

### 2. Hot Reload with Empty Registrations (Potential Issue)
- The guard's original purpose was to prevent re-registration with empty lists
- Without the guard, calling `Register()` with an empty list **will overwrite** existing read model definitions
- This could be problematic during hot reload scenarios where not all artifacts are re-registered at once

## Investigation Needed

1. **Are reactor-only applications a valid use case?**
   - If yes, we need proper initialization support
   - If no, we should document that projections/reducers are required

2. **Hot reload behavior**
   - How are artifacts re-registered during hot reload?
   - Is it all-at-once or incremental?
   - Could an incremental reload accidentally wipe read models?

3. **Better solution options**
   - Option A: Smart guard that checks if there are existing registrations before skipping
     ```csharp
     if (readModels.Count == 0)
     {
         var existing = await GetDefinitions(...);
         if (existing.Any()) return; // Skip to avoid overwrite
         // Otherwise fall through for initialization
     }
     ```
   - Option B: Separate initialization vs re-registration APIs
   - Option C: Document that projections/reducers are required

4. **ReadModelsManager grain behavior**
[...]

Issue #896, opened 2023-06-21
Add tracing for operations in Client + Kernel


einari:
We need to be able to observe calls, what calls goes to what and also measure time spent in each call.

woksin:
Done. \`AddCratisChronicleInstrumentation()\` in \`Source/Clients/AspNetCore/OpenTelemetry/TracerProviderBuilderExtensions.cs\` registers the client activity source, and the per-area spans exist as \`*Traces.cs\` companions: \`EventSequenceTraces\`, \`ReactorTraces\`, \`ReducerTraces\`, \`UnitOfWorkTraces\`, wired through keyed \`IActivitySource<T>\` registrations.

Two follow-ups worth their own issues if wanted: nothing asserts that \`AddCratisChronicleInstrumentation\` registers the source, and there is no client-tracing page in the documentation.

PR #3543, opened 2026-07-29
Cover two pieces of public API that had no specs

woksin:
Adds specs for `ReadModelsManager` and `AddCratisChronicleInstrumentation`, neither of which had any. Spec-only, so there are no release-note bullets.


woksin:
Both were surfaced while triaging the backlog, against the framework rule that public surface with zero coverage is a defect in itself.

**`ReadModelsManager`** (#2290) had no specs at all, yet it is where the answer lives to whether a client registering zero read models wipes the ones already registered — the exact question that issue was opened to settle. Two specs pin it: registering an empty set leaves existing definitions alone, and registering a known identifier replaces its definition rather than accumulating a second one.

**`AddCratisChronicleInstrumentation`** (#896) is the entire public surface for getting client traces into OpenTelemetry and nothing exercised it. An `ActivitySource` produces nothing until something listens by name, so the spec asserts the source records nothing before the extension is added and records against itself afterwards — which is what makes it non-vacuous rather than just asserting a non-null builder. Both halves are in one class deliberately: an activity listener is process-wide, so separate classes would observe each other tracer provider under parallel xUnit.

Verified: 2100 Core.Specs and 31 AspNetCore.Specs pass, Debug and Release clean.

woksin:

PR #3544, opened 2026-07-29
Correct the stale CA1852 entry in the issue analysis

woksin:
Corrects a factually wrong entry and flags the hand-written sections as stale. Repository documentation only, so there are no release-note bullets.


woksin:

Issue #845, opened 2023-04-19
Support separate event handler methods for events during replay


einari:
It should be possible to have methods for handling replay by marking the methods that are only to be triggered when replay.
If no specific replay method exists, it will call the regular one. While if the replay method exist it will not call the regular one if the `ObservationState` is flagged with `Replay`.

```csharp
[Observer("fa2f35cc-40b8-435a-a1ef-50664be03a19")]
public class MyObserver
{
    public Task HandleSomething(Something @event, EventContext context) => Task.CompletedTask;

    [Replay]
    public Task HandleSomethingReplay(Something @event, EventContext context) => Task.CompletedTask;
}
```

Torgeir-Hansen:
Is it clear enough that `HandleSomething()` will not be called during a replay with this solution?

If this is implemented, isn't it logical to also implement the opposite? A way of tagging methods not to be run during replay?

einari:
I see your point on clarity.
Question is if we want to force the developer to be explicit for both scenarios when one tags something with `[Replay]`.

```csharp
[Observer("fa2f35cc-40b8-435a-a1ef-50664be03a19")]
public class MyObserver
{
    [Forward]
    public Task HandleSomething(Something @event, EventContext context) => Task.CompletedTask;

    [Replay]
    public Task HandleSomethingReplay(Something @event, EventContext context) => Task.CompletedTask;
}
```

Another alternative could be to say you implement your replay logic in a separate class and mark the observer as a replay observer:

```csharp
[Observer("fa2f35cc-40b8-435a-a1ef-50664be03a19")]
public class MyObserver
{
    public Task HandleSomething(Something @event, EventContext context) => Task.CompletedTask;
}

[ReplayObserver("fa2f35cc-40b8-435a-a1ef-50664be03a19")]
public class MyObserverForReplay
{
    public Task HandleSomething(Something @event, EventContext context) => Task.CompletedTask;
}
```

woksin:
## Resolved on `main`

Landed in #3554:

> `[Replay]` on a reactor handler method, marking it as the one to run while the observer is replaying. Only the marked handler runs during a replay; an event type without one behaves exactly as before.

`Source/Clients/DotNET/Reactors/ReplayAttribute.cs`.

[...]

PR #3548, opened 2026-07-29
Let a reactor handle an event differently while it is replaying

woksin:
## Added

- `[Replay]` on a reactor handler method, marking it as the one to run while the observer is replaying. Only the marked handler runs during a replay; an event type without one behaves exactly as before. (#845)

### DIFF
--- a/.ai/rules/reactors.md
+++ b/.ai/rules/reactors.md
@@ -37,6 +37,25 @@ public Task MethodName(TEvent @event, EventContext context)
 - **Return type** — `Task` or `void`, or a side-effect type (`TEvent`, `ReactorSideEffect`, or a collection of either) returned directly (sync) or wrapped in `Task<...>` (async). Prefer `Task`/async for real side effects, but synchronous returns are fully supported — there is no "always async" requirement.
 - **Method name** — can be anything descriptive. The name is for readability, not dispatch.
 
+## Handling replay differently — `[Replay]`
+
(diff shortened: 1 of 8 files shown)


woksin:
Everything except the dispatch was already in place: the kernel tags replayed events and `EventContext.ObservationState` carries `Replay` all the way to the client. `ReactorInvoker` just built one map of handlers and dispatched from it regardless.

Now it builds two — live and replay — and picks by observation state. Design points worth checking in review:

- **Non-breaking by construction.** An event type with no `[Replay]` handler falls through to the live handler, which is what every reactor written until now relies on during a replay. Specced in `and_only_a_live_handler_exists_during_replay`.
- **The replay handler replaces rather than adds.** Marking one redirects the work instead of doubling it, so the live handler does not also run.
- **`GetEventTypesFor` returns the union of both maps.** A type handled only by a `[Replay]` handler still has to be subscribed to, or the replay it exists for would never deliver it — this drives subscription from `Reactors.cs` and `EventSequence.cs`. Specced separately.
- **Side benefit:** splitting the maps removes a silent last-writer-wins when two methods handled the same event type.
- `[Replay]` vs `[OnceOnly]`: use `[OnceOnly]` when the side effect should not happen again, `[Replay]` when replay needs to do something *different*. Documented in `.ai/rules/reactors.md`.

Deliberately out of scope: reducers and projections are not method-dispatched, so this does not apply to them.

Verified: 1977 client specs pass, Debug and Release clean with zero warnings. Worth noting nothing under `Integration/Client/for_Reactors` exercises replay today, so the gRPC hop for `EventObservationState.Replay` is uncovered end to end — an out-of-process spec there would be a good follow-up.

woksin:

Issue #1601, opened 2024-12-06
IObserver should not implement IStateMachine, it's an internal concern


woksin:
Remove IStateMachine from IObserver interface

woksin:
## Resolved on `main`

`IObserver` no longer implements `IStateMachine` — `Source/Kernel/Core/Observation/IObserver.cs:22`:

```csharp
public interface IObserver : IGrainWithStringKey
```

and the `<remarks>` above it records the reasoning this issue asked for, in the issue's own terms:

> The observer is implemented as a state machine internally, but that is an implementation concern and deliberately not part of this grain contract — state transitions are driven from within the observer implementation, never across a grain reference.

`git grep IStateMachine` across `Source/Kernel/Core/Observation/` returns nothing, so the concern is fully internal now.

@einari @woksin — safe to close.

woksin:
Closing — released in **v16.10.0**.

Delivered by commit `26b0be74` ("Remove state machine surface from the IObserver grain contract"). Verification against `main` is in the comment above.

PR #3549, opened 2026-07-29
Take the state machine surface off the observer grain contract

woksin:
Removes `IStateMachine<ObserverState>` from the `IObserver` grain contract. No user-facing change — no caller ever invoked those members through a grain reference — so there are no release-note bullets. (#1601)

### DIFF
--- a/Source/Kernel/Core.Specs/Observation/States/for_CatchingUpInFlight/given/a_catching_up_in_flight_state.cs
+++ b/Source/Kernel/Core.Specs/Observation/States/for_CatchingUpInFlight/given/a_catching_up_in_flight_state.cs
@@ -13,6 +13,7 @@ namespace Cratis.Chronicle.Observation.States.for_CatchingUpInFlight.given;
 public class a_catching_up_in_flight_state : Specification
 {
     protected IObserver _observer;
+    protected IStateMachine<ObserverState> _stateMachine;
     protected CatchingUpInFlight _state;
     protected ObserverState _storedState;
     protected ObserverState _resultingStoredState;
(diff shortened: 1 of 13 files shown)


woksin:
Implemented from the #1601 triage plan, but it departed from that plan in two places and both are worth a look.

**The plan wanted an `IObserverStateMachine` composite interface. It should not exist.** `IObserver : IGrainWithStringKey`, so anything deriving from it is a grain interface as far as Orleans is concerned — Orleans would have regenerated proxies and invokers for `TransitionTo`/`CanTransitionTo`/`GetStates`/`GetCurrentState` under a different name, moving the smell rather than removing it. `internal` does not work either: the spec `given` classes are public with `protected` fields, so it fails CS0052. Instead the four spec contexts create one substitute implementing both interfaces — `Substitute.For<IObserver, IStateMachine<ObserverState>>()` — exposed through `_observer` and `_stateMachine`. No production type added, no second grain interface.

**The plan's third step was wrong and would have NRE'd.** It suggested retyping `BaseObserverState.Observer` to `(StateMachine as IObserverStateMachine)!`. `StateMachine` holds the concrete `Observer` grain, which does not declare that interface, so the cast would evaluate to null. Left unchanged.

**Label.** Marked `patch` because `IObserver` is a kernel grain contract rather than client SDK surface, and no caller ever invoked those members through a grain reference. If you consider the kernel assembly public API, this is a member removal and wants `major` — your call.

Verified independently in a clean checkout (not the agent worktree): Debug and Release both 0 errors 0 warnings, Core.Specs 2093 passed / 1 pre-existing skip.

woksin:

PR #3550, opened 2026-07-29
Expose identity renaming on the .NET client event store

woksin:
## Added

- `IEventStore.Identities` with `Rename(subject, name)`, exposing the identity renaming the kernel already supported to the .NET client. (#1684)

### DIFF
--- a/Source/Clients/DotNET.Specs/Identities/for_IdentityManager/given/an_identity_manager.cs
+++ b/Source/Clients/DotNET.Specs/Identities/for_IdentityManager/given/an_identity_manager.cs
@@ -1,0 +1,30 @@
+// Copyright (c) Cratis. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Cratis.Chronicle.Connections;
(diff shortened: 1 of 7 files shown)


woksin:
The kernel half of #1684 already existed — `IIdentityStorage.Rename` across every backend and `IIdentities.RenameIdentity` over gRPC — so this is purely the missing client surface, modeled on the existing `PIIManager`.

Two things for review:

- **`IEventStore` gains a member.** Additive for consumers, but technically source-breaking for anyone implementing that interface. Labeled `minor`; say if you would rather it were `major`.
- **It keys on `Subject`, not `IdentityId`.** The issue says "based on the identifier of the identity", but the shipped kernel API keys on subject, so this follows the kernel rather than inventing a second lookup path.

The Api/Workbench surface (a POST endpoint plus rename UI) is deliberately **not** included — it is a natural follow-up, not part of the client capability this issue asked for.

Verified independently in a clean checkout: Debug and Release 0 errors 0 warnings, DotNET.Specs 1975 passed.

woksin:

Issue #1470, opened 2024-10-14
Add a way to adorn types with what they should be represented as in the Json Schema


einari:
With types having a custom Json serializer, the data type it serializes will most likely be different than what it represents itself. There will therefor today become a type mismatch between what the `JsonConverter` serializes to and what it is expecting when handling it on the server side. 

We should not rely on specific attributes of 3rd party libraries for this, especially since we're looking at changing the Json schema generator (#1459) and instead introduce our own attribute that can be used to adorn types that has a `JsonConverter` associated with it.

woksin:
## Resolved on `main`

Landed in #3554, and exactly as this issue specified — our own attribute rather than a third-party one:

> `[JsonSchemaType(typeof(T))]` on a type, overriding how it is represented in the JSON schema generated for events and read models. Compliance metadata and nullability are preserved.

`Source/Clients/DotNET/Schemas/JsonSchemaTypeAttribute.cs`.

@einari @woksin — safe to close.

woksin:
Closing — released in **v16.10.0**.

Delivered by #3554. Verification against `main` is in the comment above.

PR #3551, opened 2026-07-29
Add [JsonSchemaType] to override a type's JSON schema representation

woksin:
## Added

- `[JsonSchemaType(typeof(T))]` on a type, overriding how it is represented in the JSON schema generated for events and read models. Compliance metadata and nullability are preserved. (#1470)

### DIFF
--- a/Source/Clients/DotNET.Specs/Schemas/for_JsonSchemaGenerator/when_generating_schema_for_pii_type_with_json_schema_type_override.cs
+++ b/Source/Clients/DotNET.Specs/Schemas/for_JsonSchemaGenerator/when_generating_schema_for_pii_type_with_json_schema_type_override.cs
@@ -1,0 +1,46 @@
+// Copyright (c) Cratis. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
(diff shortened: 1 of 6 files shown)


woksin:
Generalizes the redirect that already existed in specialized form — the concept branch in `JsonSchemaGenerator.TransformNode` — into an attribute-driven one, placed ahead of that branch so an explicit override always beats inference, and after the `Nullable.GetUnderlyingType` unwrap so `MyType?` is handled.

**The name is yours to change.** Triage flagged it as unresolved and it was picked as the most defensible option rather than settled: `[JsonSchemaType(typeof(Guid))]` in `Cratis.Chronicle.Schemas`. Also deliberately **type-level only** — the issue says "adorn types", so property-level application was not added.

Compliance metadata and the nullable marker are preserved through the redirect, which the specs cover explicitly — including the `[PII]` combination, since silently losing compliance metadata on an overridden type would be the worst failure mode here.

Documentation under `Documentation/**` is **not** included and should follow, since this is public-facing developer behavior.

Verified independently in a clean checkout: Debug and Release 0 errors 0 warnings, DotNET.Specs 1981 passed.

woksin:

Issue #984, opened 2023-10-08
Optimize event type registrations on server with registering all event types instead of singles


einari:
Today the `ISchemaStore` interface exposes only a way to register single event types. For an application with a lot of event types, this would then result in a lot of round trips to the storage engine.

woksin:
## Resolved on `main`

Landed in #3554:

> Registering event types on connect issues one read and one bulk write for the whole batch instead of several round trips per event type, so startup no longer scales its database round trips with the number of event types

`Source/Kernel/Storage.MongoDB/EventTypes/EventTypesStorage.cs:165` documents the shape — a single unordered bulk write, each event type merging its generations into the stored document — and `when_registering_a_batch_of_event_types.cs` specifies it.

@einari @woksin — safe to close.

woksin:
Closing — released in **v16.10.0**.

Delivered by #3554. Verification against `main` is in the comment above.

PR #3552, opened 2026-07-29
Register every event type with one read and one write

woksin:
## Changed

- Registering event types on connect issues one read and one bulk write for the whole batch instead of several round trips per event type, so startup no longer scales its database round trips with the number of event types. (#984)

### DIFF
--- a/Source/Kernel/Concepts/Events/EventTypeToRegister.cs
+++ b/Source/Kernel/Concepts/Events/EventTypeToRegister.cs
@@ -1,0 +1,15 @@
+// Copyright (c) Cratis. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Cratis.Chronicle.Concepts.Events;
+
+/// <summary>
+/// Represents an <see cref="EventTypeDefinition"/> to register together with the <see cref="EventTypeSource"/> it came from.
+/// </summary>
+/// <param name="Definition">The <see cref="EventTypeDefinition"/> to register.</param>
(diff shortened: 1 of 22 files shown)


woksin:
The highest-risk of this batch, so worth saying exactly what was checked.

**What changed.** The service reads all definitions once and answers every existence and schema-equality question from that dictionary instead of `HasFor`/`GetFor` per type and per generation. Storage gains `Register(IEnumerable<EventTypeToRegister>)` with a **default interface implementation** that loops the existing single-type `Register`, so `Storage.InMemory`, `Storage.Sql` and `Clients/Testing` compile untouched; MongoDB overrides it with one `FindAsync` and one unordered `BulkWriteAsync` of only the documents that actually change.

**The contract that had to be preserved** is the set of ids reported back as mutated: it drives cache invalidation and the `EventTypeAdded`/`EventTypeGenerationAdded` system events. Reporting a type that did not change is as wrong as missing one that did.

**I added the spec that was missing.** The MongoDB bulk override — the riskiest code here — had no coverage; the agent was barred from Docker specs, so the branch shipped it unspecced. `when_registering_a_batch_of_event_types` now runs against a real MongoDB and pins that contract: a first registration reports every type, the same batch again reports **none**, and changing one schema reports **only** that one.

Verified independently in a clean checkout: Debug and Release 0 errors 0 warnings; Services.Specs 46, Storage.Specs 62, Storage.InMemory.Specs 85, and Storage.MongoDB.Specs 581 — the Docker-backed suite, run serially.

Public API names are yours to change: the record `EventTypeToRegister` and the new `IEventTypesStorage.Register` overload.

woksin:

PR #3553, opened 2026-07-29
Add an Aspire helper that provisions MongoDB as a replica set

woksin:
## Added

- `AddCratisChronicleMongoDB()` for Aspire, provisioning a MongoDB container that initiates itself as a single-node replica set and hands back a `directConnection` connection string — which is what Chronicle requires and what a plain MongoDB resource does not provide. (#3400)

### DIFF
--- a/Chronicle.slnx
+++ b/Chronicle.slnx
@@ -3,6 +3,7 @@
     <Folder Name="/Source/Clients/">
         <Project Path="Source/Clients/Api/Api.csproj" />
         <Project Path="Source/Clients/Aspire/Aspire.csproj" />
+        <Project Path="Source/Clients/Aspire.Specs/Aspire.Specs.csproj" />
         <Project Path="Source/Clients/AspNetCore/AspNetCore.csproj" />
         <Project Path="Source/Clients/AspNetCore.Specs/AspNetCore.Specs.csproj" />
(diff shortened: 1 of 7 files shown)


woksin:
Two things to flag.

**The agent went beyond its brief, and it was the right call.** Triage said there is no Aspire spec project and therefore no automated gate — compilation would be the only verifiable thing. Rather than ship unverifiable, it added `Source/Clients/Aspire.Specs` with 16 specs asserting the resource shape: the container image and tag, the entrypoint and replica-set arguments, the endpoint, and the `directConnection` connection string. That is a new project in `Chronicle.slnx`; the solution diff is one line, alphabetically placed, with no normalization damage.

**What is still not verified:** that the container actually reaches a healthy single-node replica set at runtime. The specs pin the resource definition, not Docker behavior. Someone should run an AppHost with `AddCratisChronicleMongoDB()` + `AddCratisChronicle(configure: c => c.WithMongoDB(mongo))` before this is relied on.

Note the documentation half of #3400 already landed in f874ee02e; `Documentation/hosting/aspire.mdx` and `mongo-replica-set.md` should be updated to lead with this helper and keep the manual recipe as the fallback — deliberately left out of this PR.

Verified independently in a clean checkout: Debug and Release 0 errors 0 warnings, Aspire.Specs 16 passed.

woksin:

PR #3554, opened 2026-07-29
Consolidated fixes and improvements

woksin:
## Added

- `[Replay]` on a reactor handler method, marking it as the one to run while the observer is replaying. Only the marked handler runs during a replay; an event type without one behaves exactly as before. (#845)
- `IEventStore.Identities` with `Rename(subject, name)`, exposing the identity renaming the kernel already supported to the .NET client. (#1684)
- `[JsonSchemaType(typeof(T))]` on a type, overriding how it is represented in the JSON schema generated for events and read models. Compliance metadata and nullability are preserved. (#1470)
- `AddCratisChronicleMongoDB()` for Aspire, provisioning a MongoDB container that initiates itself as a single-node replica set and hands back a `directConnection` connection string. (#3400)
- `IReactorMiddleware.BeforeInvoke` and `AfterInvoke` overloads that receive the `ReactorId` of the reactor observing the event. Chronicle calls these; they default to the existing overloads, so a middleware written against the previous interface keeps working unchanged. (#1009)
- `CompliancePropertyNotFoundInSchema`, thrown when a property being applied or released has no counterpart in the schema. The message names the property path, the compliance subject, and the properties the schema does declare. (#690)
- An overview on the `cratis/chronicle` Docker Hub repository covering the published tag variants, ports, quick start and the environment variables that configure the server. (#1277)

## Changed

- Registering event types on connect issues one read and one bulk write for the whole batch instead of several round trips per event type, so startup no longer scales its database round trips with the number of event types. (#984)

## Fixed

- Reading back a job whose request was not yet assigned no longer fails with a `NullReferenceException`, losing the status and progress the job state was written to record. (#3541)
- Reducer failures are written to the client log. A reducer method that throws was reported to the kernel as a failed observation without ever appearing in the client's own log. (#1395)
- Applying or releasing compliance metadata for a property the schema does not declare no longer fails with a bare `Sequence contains no matching element`. (#690)
- Applying or releasing compliance metadata no longer fails when a schema flattened across inheritance declares the same property name more than once. (#690)

### DIFF
--- a/.ai/rules/reactors.md
+++ b/.ai/rules/reactors.md
@@ -37,6 +37,25 @@ public Task MethodName(TEvent @event, EventContext context)
 - **Return type** — `Task` or `void`, or a side-effect type (`TEvent`, `ReactorSideEffect`, or a collection of either) returned directly (sync) or wrapped in `Task<...>` (async). Prefer `Task`/async for real side effects, but synchronous returns are fully supported — there is no "always async" requirement.
 - **Method name** — can be anything descriptive. The name is for readability, not dispatch.
 
+## Handling replay differently — `[Replay]`
+
+A reactor sees the same event twice for different reasons: as it happens, and again when its observer is replayed. When those call for different work, mark a second handler for the same event type with **`[Replay]`** and it takes over for the duration of the replay.
+
+```csharp
+public class OrderNotifications(INotificationService notifications) : IReactor
+{
+    public Task OrderPlaced(OrderPlaced @event) => notifications.Notify($"Order {@event.Number} placed.");
+
+    [Replay]
+    public Task OrderPlacedDuringReplay(OrderPlaced @event) => Task.CompletedTask;
+}
+```
+
+- With a `[Replay]` handler, **only** it runs during replay — the regular handler does not also run.
+- Without one, the regular handler runs during replay exactly as before, so this changes nothing for existing reactors.
+- An event type handled *only* by a `[Replay]` handler is still subscribed to.
+- Reach for **`[OnceOnly]`** instead when the side effect should simply not happen again; use `[Replay]` when replay needs to do something *different* rather than nothing.
+
 ## Taking Dependencies
 
 Beyond the event and `EventContext`, a handler method can take any number of additional parameters as dependencies. Only the first parameter is fixed (it is the event that drives dispatch); every parameter after it is resolved when the method is invoked.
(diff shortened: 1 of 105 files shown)


woksin:
Consolidates 16 branches into one merge so this cuts **one release instead of 14**. `publish.yml` triggers on `Source/**`, `Docker/**` and `publish.yml`, and 14 of the 16 touch one of those.

Merged rather than squashed, so every original commit and its reasoning survives and the PR can still be reviewed commit by commit.

**Internal changes not in the release notes**, since they have no user-facing effect:

- The `benchmarks` job was failing on *every* pull request — `Chronicle.Benchmarks.csproj` reached the client API transitively through `Server.csproj`, whose reference is dropped when `ArtifactsPath` is set, which is exactly how BenchmarkDotNet rebuilds. Fixed, plus three things that let it land silently: the build and run shared one step (a failed build reported as a *successful* step), `Benchmarks/**` was not in the workflow path filter, and an empty result set only failed later as "no benchmark result was found".
- Benchmark runs move to `main`. They were ~13 of the ~19 minute pull-request critical path and could never fail a build (`fail-on-alert: false`); pull requests keep a build-only guard that builds the way BenchmarkDotNet does. The series is now committed by the run on `main`, since the release previously restored it from a cache written on the pull request.
- Spec class maps are registered from the module initializer instead of lazily — the race that produced the `specs` failure on #3529.
- New specs for `ReadModelsManager`, `AddCratisChronicleInstrumentation` and the MongoDB batch registration, all of which had none.
- A corrected CA1852 entry in `IssueAnalysis.md`.

**Verified on the merged result, not on the branches individually:** full-solution Debug and Release builds at 0 errors / 0 warnings, and every in-process spec project green — Core 2104, DotNET 1995, MongoDB 585 (Docker-backed, run serially), Testing 272, Storage 62, InMemory 85, Services 46, AspNetCore 31, Aspire 16.

Worth checking in review: `ReactorInvoker.cs` was touched by both the middleware-reactor-id change and the `[Replay]` dispatch change. Git merged them without conflict and the result is correct — `TryGetHandlerFor(...)` and `BeforeInvoke(reactorId, ...)` are both present — but it is the one genuine semantic overlap in the set.

Decisions still open, each flagged on the original PR: the `[JsonSchemaType]` attribute name (#3551), `EventTypeToRegister` and the new `IEventTypesStorage.Register` overload name (#3552), whether removing four members from the `IObserver` grain contract should be `major` (#3549), and whether adding a member to `IEventStore` should be `major` (#3550). The label here is `minor` on the basis that nothing breaks a consuming application.

Not verified: that the Aspire helper's container actually reaches a healthy replica set at runtime — its 16 specs pin the resource definition, not Docker behavior.

PR #3556, opened 2026-07-29
Keep the Docker Hub short description within the limit

woksin:
Trims the short description to fit Docker Hub's 100-byte limit. Release pipeline only, so there are no release-note bullets.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -432,6 +432,7 @@ jobs:
       # The description API needs a token with write scope, which the push credentials do not necessarily
       # have - never fail a release over an overview that did not update.
       - name: Update Docker Hub overview
+        id: dockerhub-overview
         if: needs.release.outputs.prerelease != 'true'
         continue-on-error: true
         uses: peter-evans/dockerhub-description@v4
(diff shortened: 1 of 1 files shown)


woksin:
Found by reading the first release run of the step (30485909016), not from a failing check — `continue-on-error: true` means the step reports **success** while the log says otherwise.

Two separate things surfaced there:

```
##[warning]The short description exceeds DockerHub's limit and has been truncated to 100 bytes.
Sending PATCH request
##[error]Forbidden
```

This PR fixes only the first — 104 bytes down to 95.

**The second is a credentials matter and needs you, not code.** The PATCH was rejected with `Forbidden`, so the overview never reached Docker Hub: `GET https://hub.docker.com/v2/repositories/cratis/chronicle` still returns an empty `description` and an empty `full_description`. `Docker/README.md` is in the repo and correct, but #1277 is not achieved in practice until `DOCKER_PASSWORD` is a token with write/admin scope rather than push-only.

The `continue-on-error` guard did its job — the release published normally — but it also means this will keep failing silently every release until the secret is changed. Worth deciding between granting the scope or dropping the step and pasting the overview in the Docker Hub UI once.

PR #3563, opened 2026-07-30
Release PII per property instead of failing the whole read

woksin:
# Summary

Releasing PII now degrades a single unreadable property instead of failing the whole read.

## Added

- Several event types declared under the same constraint name form one constraint allowing at most one event from that set per event source, making mutually exclusive outcomes enforceable at append time.
- `CHR0039` warns when a `Task`-returning testing assertion is discarded rather than awaited, where it can never fail and `CS4014` does not fire.

## Changed

- A `[PII]` value that Chronicle never encrypted is passed through untouched when releasing, instead of being blanked.

## Fixed

- Releasing a `[PII]` value encrypted under a different subject no longer fails the whole read model and the query returning it. That property is surfaced as empty and the diagnostic goes to the Kernel log.
- Releasing a `[PII]` value for a subject that never had an encryption key no longer returns an empty value indistinguishable from crypto-shredded erasure.

### DIFF
--- a/Documentation/client-snippets/code-analysis/chr0039/fix.md
+++ b/Documentation/client-snippets/code-analysis/chr0039/fix.md
@@ -1,0 +1,25 @@
+```csharp
+using Cratis.Chronicle.Events;
+using Cratis.Chronicle.EventSequences;
+using Cratis.Chronicle.Testing.EventSequences;
+using Xunit;
+
+[EventType]
+public record Chr0039AuthorRegisteredFixed(string Name);
+
+public class Chr0039WhenRegisteringAnAuthorFixed
+{
+    readonly IEventLog _eventLog = default!;
+
+    // Declaring the fact 'async Task' and awaiting the assertion makes the exception it throws
+    // observable, so the assertion can actually fail.
+    [Fact]
+    async Task should_have_appended_the_event() =>
+        await _eventLog.ShouldHaveAppendedEvent<Chr0039AuthorRegisteredFixed>(e => e.Name == "Jane Austen");
+
+    // Returning the Task works too — the test runner awaits it.
+    [Fact]
+    Task should_have_appended_exactly_one_event() =>
+        _eventLog.ShouldHaveTailSequenceNumber(EventSequenceNumber.First);
+}
+```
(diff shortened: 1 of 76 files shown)


woksin:
Reviewer context — not release notes.

**Origin.** All three items came from defect reports raised while building an application on Chronicle. Each was re-verified against `main` at 16.11.1 before any code was written; two further reported items were investigated and deliberately **not** acted on:

- An empty `[ChildrenFrom]` collection serializing as absent rather than `[]` is **intentional**. `ChangesetExtensions.AddPropertiesFrom` filters children paths out of the root projection's `PropertiesChanged` so a parent's `Features=[]` cannot race ahead of a sibling's `ChildAdded` during parallel replay and wipe children already written. The suggested fix — serialize `[]` — reintroduces that race. The consumer-visible half is real but belongs in the Arc TypeScript proxy generator, which still types collection fields as non-optional `T[]` with no `= []` default (52 fields across 209 generated proxies).
- A zero-valued enum reading back as `null` could not be reproduced. `BsonValueExtensions.ToBsonValue` converts an enum `0` to `BsonInt32(0)` with no elision, so the reported "sink elides defaults" mechanism does not hold. Needs a live-kernel repro before it is worth chasing.

**Internal changes not in the release notes.** `IUniqueEventTypesConstraintsStorage.IsAllowed` now takes the set of covered event types instead of one; `UniqueEventTypeConstraintDefinition` carries `EventTypeIds`; the contract's `UniqueEventTypeConstraintDefinition` moves from a single `EventTypeId` (field 1, now reserved) to a repeated `EventTypeIds` (field 2). Labelled `minor` rather than `major`: `builder.Unique<TEvent>(...)` is unchanged, and Arc references none of these types — it uses only `ConstraintViolation` and friends.

Grouping happens when the builder builds rather than in the kernel, because `Constraints.GetFor` resolves with `.Single()` and the kernel joins definitions by name — duplicate names reaching either would break them. The definition compares its event types by content, since registration decides "changed?" via `Equals` and generated record equality would compare the collection by reference and report a change on every startup.

**Verification.** Debug and Release both 0 errors / 0 warnings; each of the four commits builds independently. 4594 specs pass across Core, DotNET, Testing and CodeAnalysis (1 pre-existing skip). Client snippet compile validation and markdownlint pass.

**Not verified locally.** `Storage.MongoDB.Specs` needs a container, so its updated compliance specs compile but were not run. The gRPC contract change is the thing most likely to draw a CI opinion — worth watching `grpc-compatibility`.

PR #3565, opened 2026-08-01
Fix failed-partition recovery, method-level OnceOnly, and pre-16.12 constraint upgrade

woksin:
# Summary

Four cases where Chronicle reported success while quietly doing nothing.

## Changed

- `[OnceOnly]` on a reactor handler method now takes effect — the handler is skipped for replayed events while the rest of the reactor still replays. Only class-level placement was honored before.
- The kernel now warns when it skips a concurrency check because the append declared a concurrency scope with no expected sequence number.

## Fixed

- A 16.12 or newer kernel can now read constraint definitions written by an older kernel, so an in-place upgrade no longer breaks constraint registration and client startup.
- Recovering a failed partition no longer clears it without re-running the handler, silently losing the missed side effect.
- The observer management API's `try-recover-failed-partition` and `replay` routes were swapped, so recovery replayed and replay retried.
- A job step that handled no events no longer overwrites the sequence number recorded by an earlier step that did.


woksin:
Reviewer context — these came in as four upstream defect reports (CHR-30 … CHR-33) from a consumer running 16.10.0 and upgrading to 16.13.1.

**CHR-32 — pre-16.12 store** (`4b09223be`). `UniqueEventTypeConstraintDefinition` widened from `EventTypeId` to `IEnumerable<EventTypeId>` in 16.12.0 (the constraint-grouping feature) and nothing mapped the stored shape onto it, so `Equals` dereferenced a null sequence and `Constraints.Register` died on the first legacy definition — taking every constraint with it. Fixed in two places: `ConstraintDefinitionSerializer` maps a legacy `eventTypeId` element onto a one-element `eventTypeIds` (preserving the constraint's meaning, which null-safety alone would not), and the record normalizes an absent sequence as the backstop. Note the upgrade builds a new document rather than editing in place — the one read back is a `RawBsonDocument` and rejects mutation; the spec caught that.

**CHR-30 — failed-partition recovery** (`7784ad3f6`, `aa17972b1`). Split into the behavioral fix and the route swap because they revert independently. The report noted it could not pin *why* `GetRange` returned zero events for a correctly-recorded failure; this fix does not depend on that answer — it makes the clear conditional on evidence from the event sequence either way, and logs both branches so the next occurrence is diagnosable from the kernel log alone.

**CHR-31 — method-level `[OnceOnly]`** (`44a171ebc`). The report offered honoring the placement or rejecting it in the analyzer. Honoring it: the invoker already receives `EventContext.ObservationState`, so per-handler replay suppression is a few lines at dispatch, needs no wire-contract change, and makes the attribute's existing XML docs true. Rejecting it would have forced consumers onto strictly coarser class-level placement after a per-reactor semantic review.

**CHR-33 — Chronicle half only** (`39d104e65`). The primary fix is in **Arc**, not here: `ConcurrencyScopeBuilder.BuildFromCommandContext` hardcodes `EventSequenceNumber.Unavailable` and `EventSourceId: null`, so `concurrency: true` produces a scope that displaces the client's default `OptimisticConcurrencyStrategy` and then validates nothing. Suggested fixes 1–3 (resolve the tail, fix scope granularity, per-target-stream scopes for cross-stream appends) belong to an Arc PR and are deliberately **not** attempted here — fixing the sequence number without the granularity would turn a silent no-op into a per-`EventSourceType` write lock. What is in scope for Chronicle is suggested fix 4: the state is now named (`ConcurrencyScope.IsIncomplete`, both kernel and client) and the kernel warns instead of failing open silently. Happy to open the Arc PR next if wanted.

**Verification.** `dotnet build Chronicle.slnx` clean in Debug and Release, zero warnings. Full spec runs green for every affected project: Kernel `Concepts.Specs` (112), `Core.Specs` (2187), `Storage.MongoDB.Specs` (590), `Storage.InMemory.Specs` (98), `Storage.Specs` (62), `Services.Specs` (46), `Server.Specs` (24); Clients `DotNET.Specs` (2002), `Testing.Specs` (272), `CodeAnalysis.Specs` (209), `AspNetCore.Specs` (31). Docs build + link check pass; the 16 broken links reported are pre-existing and in unrelated sections.

Not verified: none of this was exercised against a live kernel — the CHR-32 upgrade path in particular is covered by a serializer spec over the legacy document shape rather than by booting a new kernel on an old volume.

woksin:
Thanks — three of the four review comments were right and are fixed in the follow-up commit:

- **Null result in `JobStateWithLastHandledEvent`** — real. `TryGetFullResult` carries `[NotNullWhen(true)]`, but its implementation suppresses with `!` after `DeserializeIfNecessary`, so a step that succeeded without a result object (`JobStepResult.Succeeded()`) returns true with a null result. Guarded once at the top instead of relying on the annotation.
- **Dead `reader`/`ctx` in `ConstraintDefinitionSerializer`** — correct, pre-existing; removed.
- **"exactly once" on the OnceOnly page** — correct and worth more than wording. `OnceOnly` gates on the `Replay` observation state only. A failed-partition retry passes `EventObservationState.None`, so the event keeps its stored `Initial` state and the handler *does* run again — which is exactly what a retry is for. The page now says so.

**Declining the generic-catch flag on `RetryFailedPartition`.** It is deliberate and asymmetric on purpose: the catch decides whether a failure record is stale, and any failure to reach the event sequence must resolve to "assume events remain". Narrowing it to specific storage exceptions would let an unanticipated one escape and, on the next retry, clear a partition whose handler never ran — the exact data loss this PR exists to close. It logs the exception and returns the conservative answer rather than swallowing it.

woksin:
Companion PR for the Arc half of CHR-33: Cratis/Arc#2437 — that one resolves the expected sequence number, binds the scope to the event source being appended to, and builds a scope per target for cross-stream appends. This PR keeps the Chronicle-side half: naming the fail-open state (`ConcurrencyScope.IsIncomplete`) and warning when the kernel skips such a scope.

PR #3566, opened 2026-08-01
Fix event source type scoping in tail sequence number reads

woksin:
# Summary

Follow-ups from #3565: two scoping bugs that only became reachable now that a command can declare a real concurrency scope.

## Fixed

- The MongoDB event sequence storage narrows the tail sequence number by an actual event source type, and stops narrowing on the `Default`/`Unspecified` sentinel. It had the check inverted, so asking for the tail of one event source type answered across every type, and asking for the tail of everything answered with only the events carrying the sentinel.
- The optimistic concurrency strategy reads the expected tail with the same narrowing it puts on the scope. It read the tail of the event source alone while the kernel validates through the scope's stream type, stream id and event source type, so the two could disagree — reporting conflicts between independent appends, or missing real ones.


woksin:
Follow-up to #3565 / Cratis/Arc#2437. Both bugs are pre-existing but were unreachable until that pair landed, because no caller ever put a real (non-sentinel) event source type or stream metadata into a concurrency scope.

**The MongoDB sentinel inversion** is the more serious of the two, and it is not confined to concurrency — `GetTailSequenceNumber` is also the gRPC tail surface any client can call. The check read:

```csharp
if (eventSourceType?.IsDefaultOrUnspecified == true)   // narrows when it means "do not narrow"
```

Every sibling in the same method uses the negated form (`eventSourceId?.IsSpecified == true`, `eventStreamType?.IsAll == false`, `eventStreamId?.IsDefault == false`), and both `Storage.Sql` and `Storage.InMemory` already use `== false` here. MongoDB — the one the server runs on — was the outlier, which is exactly the divergence `framework.md` warns about: a signature match that hides a semantics mismatch, failing by returning wrong data rather than by throwing.

**The strategy asymmetry**: `OptimisticConcurrencyStrategy` attached stream type / stream id / event source type to the scope but read the expected tail from `GetTailSequenceNumber(eventSourceId)` alone, while `ConcurrencyValidator` reads it back through all of the scope's metadata. Same sentinel accident hid it — the strategy previously only ever received `EventStreamType.All` / `EventStreamId.Default` / `EventSourceType.Default`, which the kernel ignores anyway, so the two agreed by coincidence.

**Also here:** a `Cratis.Chronicle.Api.Specs` project. The swapped retry/replay routes in #3565 shipped without a spec because nothing could see the API controllers; that gap is now closed, with both directions of the pairing pinned. And a corpus rule in `.ai/rules/vertical-slices.md` claimed a command without `concurrency: true` appends with no check at all, which was never true — corrected to describe what actually happens.

**Verification.** `Chronicle.slnx` builds clean in Debug and Release, zero warnings, now 48 projects. Affected suites green: `Api.Specs` (4, new), `Concepts.Specs` (112), `Core.Specs` (2193), `Storage.MongoDB.Specs` (593), `Storage.InMemory.Specs` (98), `DotNET.Specs` (2010), `Testing.Specs` (272).

Not verified: no live concurrency race. The MongoDB fix is covered by a spec that appends events under two different event source types and asserts the tail for each, plus the sentinel case — that one runs against a real replica-set Mongo, so it does exercise the actual query.

Issue #1091, opened 2024-02-12
Add retry policies for client observers and reducers


einari:

einari:
Duplicate of #494

PR #3567, opened 2026-08-01
Update Cratis.Fundamentals to 7.16.8

woksin:
## Changed

- Updated `Cratis.Fundamentals` and `Cratis.Metrics.Roslyn` to 7.16.8.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -14,8 +14,8 @@
     <PackageVersion Include="Aspire.Hosting.AppHost" Version="13.4.6" />
     <PackageVersion Include="Aspire.Hosting.Yarp" Version="13.4.6" />
     <!-- Cratis -->
-    <PackageVersion Include="Cratis.Fundamentals" Version="7.16.7" />
-    <PackageVersion Include="Cratis.Metrics.Roslyn" Version="7.16.7" />
+    <PackageVersion Include="Cratis.Fundamentals" Version="7.16.8" />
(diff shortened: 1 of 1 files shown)


woksin:
First hop of the dependency chain after Cratis/Fundamentals#1090 and #1091 landed in Fundamentals 7.16.8.

**Scope note, so nobody expects more than this delivers.** The Fundamentals change in 7.16.8 is JavaScript-only (`JsonSerializer` giving an absent collection an empty array). Chronicle has no direct `@cratis/fundamentals` dependency — the Workbench receives it transitively through `@cratis/arc.*` — so this bump carries no functional change for Chronicle's C#. It keeps the version current so the chain can move; the consumer-visible effect arrives once Arc picks up the new Fundamentals.

`Cratis.Metrics.Roslyn` ships from the same repository on the same version and is bumped with it, rather than being left a version behind and risking the kind of transitive skew that has produced red CI here before.

**Verification.** `Chronicle.slnx` builds clean in Debug and Release, 48 projects, zero warnings. Specs green: `Concepts.Specs` 112, `Core.Specs` 2193, `Storage.MongoDB.Specs` 593, `DotNET.Specs` 2010, `Testing.Specs` 272, `Api.Specs` 4.

**Also worth a look, deliberately not included here:** `Cratis.Arc.*` is pinned at 20.69.0 while 20.69.1 is released and contains the `concurrency: true` fix from Cratis/Arc#2437. That felt like a separate decision from a Fundamentals bump, so I left it out — say the word and it is a one-line follow-up.

PR #3572, opened 2026-08-04
Fix 13 defects and land 8 improvements found building on Chronicle

woksin:
# Summary

Defects and missing seams found running Chronicle in a large production application.

## Added

- `IEventStore.Registration` reports what became of each declared artifact when registration ran, and what stopped the run if it failed, with `WaitForRegistration()` to await it
- `IProjections.GetStateForModel<TReadModel>()` and `GetFailedPartitionsForModel<TReadModel>()` reach a projection by the read model it maintains, which for a model-bound projection is the only handle there is
- `ReadModelScenario` exposes the client artifacts registry, so a spec can ask which artifacts were registered
- `[ReservedProtoFields]` keeps retired proto field numbers reserved across regeneration
- `CHR0039` now covers every awaitable and the integration testing surface

## Changed

- A unique constraint violation message no longer includes the offending value; it moved to the violation details, where the application chooses whether to render it
- Client `EventSequenceNumber.Max` is now `ulong.MaxValue - 1`, matching the kernel; code that captured the old value sees it as an ordinary sequence number
- Subscribing the same handler to `IConnectionLifecycle.OnConnected` or `OnDisconnected` twice now subscribes it once; removing is unchanged
- `IEventStore` gains a member, so any implementation outside this repository must add it

## Fixed

- The pre-16.12 constraint-store upgrade shipped but never ran, because its serializer was never registered
- One unregistered event type aborted projection discovery for every projection, taking the whole read side down
- A projection replay rewrote every stored child whole, defaulting every member the projection did not set
- Reading a read model turned an absent collection into `null` instead of an empty one
- The compliance release pass blanked a value that had never been encrypted
- A client concurrency scope meaning "no expected sequence number" arrived at the kernel as an ordinary number, so the kernel ran a check that could only pass
- A skipped concurrency check was attributed to the caller
- `GetStateFor<TProjection>()` looked in the read model dictionary with a projection type, so it could only ever throw
- A failed projection registration named every identifier in the batch instead of the definition that failed
- A registration that threw left its outcome looking unfinished, so a consumer waiting on it timed out instead of being told what failed
- Two projections maintaining the same read model resolved silently by declaration order
- Two dependency injection containers in one process sharing a namespace were handed the same event store
- Concurrent callers asking for the same event store each built their own and ran discovery twice
- An event store that could not be reached on its first resolution stayed unreachable for the lifetime of the process
- `CHR0002` demanded `[EventType]` on generic arguments that are not events
- `CHR0004` and `CHR0005` analysed a reactor's private helper methods as if they were handlers
- `CHR0010` counted the absence of `[EventStore]` as a distinct store, so a single-store projection reported as multi-store
- `CHR0025` reported a collision an event named by `[ChildrenFrom]` cannot cause
- The code analysis rule index was missing its `CHR0039` row
- `[NoAutoMap]` was ignored on `[ChildrenFrom]` children and `[Nested]` members, and a nested object used the root's exclusions instead of its own
- Two `[SetFromContext<T>]` for the same event type on one member silently discarded all but the last
- A `message:` on a property-level `[Unique]` was silently discarded
- Three attribute docstrings promised that projections filter on event metadata
- `Cratis.Chronicle.XUnit.Integration` omitted the embedded kernel's runtime dependencies from its package
- A unique constraint definition compared its covered events by reference, so every re-registration looked like a change and asked for a reindex


woksin:
## Reviewer context

A backlog of 22 items — 13 defect reports and 9 improvement proposals — found while building a large production application on Chronicle. Every cited `file:line` was re-verified at HEAD, every defect reproduced at the right tier before being fixed, and every fix mutation-proven: red before the fix, green after, red again with the fix deleted.

### Semver

**Minor.** The public surface only grows. Two things worth calling out in release notes:

- `IEventStore` gains a `Registration` member, so any implementer outside this repo must add it. Both in-tree implementers (`EventStore`, `EventStoreForTesting`) are updated.
- `EventSequenceNumber.Max` changes value on the client. That is the fix, not a side effect — the two copies disagreed and sentinels are wire values — but code that captured the old number (`18446744073709551613`) will now see it as an ordinary sequence number.

### Verification

- Release build: 0 warnings, 0 errors.
- Full spec gate: 6526 passed, 0 failed, across all 16 spec assemblies.
- Integration tiers were not run locally; out-of-process client integration needs a Docker image not present in this environment. A pre-existing pair of failures in `counting_events_per_key` was baselined against `main` in a separate worktree and is unrelated to these changes.
- Where a fix's own specs survived a mutation, the gap was closed rather than reported as covered. That happened once: the registration-attribution search recurses because the attribution arrives wrapped by `Task.WhenAll` and the grain call, but every spec handed it over bare, so removing the recursion left them all green.

### Refuted rather than fixed

Four report claims did not survive verification, and are recorded as refuted:

- **CHR-27** (the Mongo read model sink drops a zero-valued projected enum) did not reproduce. A spec now pins that the sink stores it.
- **IMP-12** (`CHR0025` recommends `[NoAutoMap]` where the attribute is inert) was closed by the `[NoAutoMap]` fix in this branch — the attribute is no longer inert there.
- **IMP-16**'s headline premise (a consumer must reach past Chronicle to the MongoDB driver for "absent means empty") was closed by the absent-collection fix.
- **IMP-7** axis 3 was refuted empirically, and **IMP-15**'s `IsConnected` claim was refuted — `ConnectionLifecycle.Connected()` sets it before invoking the `OnConnected` handlers, so it reads `true` while registration is still in flight. That is documented in three places so nobody wires the two together later.

### Deferred

**IMP-6 — `IReadModels` has no non-key query surface.** Filed as needing a design ruling rather than a patch: whether read models are queryable at all is the open question, and it touches contract, proto, kernel service, client and harness. Bringing the question, not an implementation.

### Scoping decisions

- `Registration` covers projection artifacts only. Reactors' and reducers' `Register` returns before any round trip to the kernel, so including them would report a hope as a fact; consumers are pointed at `WaitTillSubscribed` / `WaitTillActive` instead.
- The offending value in a unique constraint violation moved from the message to the details. Constraints are validated after compliance, so a property marked as PII carries its encrypted form there rather than the appended value — both key docstrings say so.
- One repository-level build change: nested checkouts were picking up a second `.globalconfig` and losing the analyzer baseline (~3250 spurious errors). `Directory.Build.targets` now uses only this repository's own.

woksin:
## Design review — open questions on the new public surface

I reviewed the design of what this branch adds rather than only its behaviour. Two defects found this way are fixed in `9f0d73062` and `b4e95fccc`. Four things remain that are design calls rather than defects, and they are worth a decision before this merges — the API is unreleased, so all four are cheapest to change now.

### 1. `WaitForRegistration` throws away the failure it exists to report

`Registration` is only assigned as the last statement of `RegisterAll`. If `RegisterAll` throws — which it can, and which the connection lifecycle rethrows — `Registration` stays `NotRun` forever. A consumer calling `WaitForRegistration()` then burns the full timeout and gets `TaskCanceledException`, with the actual cause nowhere in it.

The total-failure case is the one where the feature tells you least, and it is the case the feature was proposed for. Options: add an `Exception? Failure` to `RegistrationOutcome` and publish it from a `catch` before rethrowing (additive, and `WaitForRegistration` can rethrow it); or accept the timeout and say so in the docs. I lean to the first.

### 2. `Registration` is a breaking change to `IEventStore` that need not be

Adding a member to a public interface breaks every implementer outside this repo. A default implementation — `RegistrationOutcome Registration => RegistrationOutcome.NotRun;` — would make it additive, and both in-tree implementers override it anyway. The repo uses no default interface members today, so this is a convention question, not just a mechanical fix. If the answer is no, the semver caveat stands as written.

### 3. `GetStateFor<TProjection>` is now correct and still unusable for the common case

The fix pointed it at the right dictionary. But its `where TProjection : IProjection` constraint means model-bound projections — which have no projection type — cannot be named at all, so there is still no way to ask for their state. `GetFailedPartitionsFor(Type)` was widened to accept either handle and works for both. So a consumer can get failed partitions for a model-bound projection but not its state. The asymmetry is defensible per method and odd as a pair; a `GetStateFor(Type)` overload would close it.

### 4. A static cache keyed on less than it captures (pre-existing)

`_eventStores` in both client registration extensions is `static` and keyed only by `EventStoreNamespaceName`, but its factory captures the first resolving `IServiceProvider` and the event store name. Two DI containers in one process sharing a namespace — the normal shape of `WebApplicationFactory`-style tests — get one `IEventStore` built from whichever resolved first. This predates the branch; I only added the eviction. It is worth flagging because my own specification for that eviction had to mint a random namespace to avoid the collision, which is the tell.

### Duplication this branch grew rather than shrank

`EventSequenceNumber` exists twice and this branch contains a fix for a bug caused by those two copies drifting. `WellKnownConstraintDetailKeys` is byte-identical in kernel and client and I edited both by hand. The two `ChronicleClientServiceCollectionExtensions` differ in 36 of ~145 lines, so the `Lazy` fix and its specification were written twice. No change proposed here — but the branch is evidence for the cost.

### What the two committed fixes were

- **`9f0d73062`** — `DeclareReservedFields` skipped quietly on all three paths that could fail to produce a reservation, including a missing message. A reservation that does not happen frees the retired number for reuse, which is what the attribute exists to prevent. Worse, `ProtoGenerator` is not in the solution and had no specifications project, so none of it was ever compiled by the build gate or covered by the spec gate — the built assembly on disk predated the change by months. Both projects are in the solution now and the gate runs 17 assemblies instead of 16.
- **`b4e95fccc`** — `RegistrationOutcome` is a record whose only meaningful member is a list, so its generated equality compared by reference and two identical outcomes came out unequal. This is the same defect as the constraint-definition one fixed earlier on this branch, reached the same way.

Gate after both: 0 warnings, 0 errors; 6541 passed, 0 failed.

woksin:
Re-ran the failed jobs on `b4e95fc`. `integration-api` passed on the retry, so the earlier failure was a transient `actions/cache` restore timeout, not the image build. With the image available the integration specs ran for the first time on this branch and surfaced two separate things.

**Fixed in 23d905e — a stale assertion, deterministic.** `for_ReadModels.when_getting_instance_for_passive_projection.and_parent_event_exists.should_not_have_event_models_yet` asserted `ShouldBeNull()` and got `[]`. `PassiveApplication.EventModels` is declared as a non-nullable `IEnumerable<PassiveEventModel>`, and `Documentation/projections/model-bound/children.mdx` in this branch states the rule as "non-nullable materializes as an empty collection, nullable stays `null`". So `[]` is the intended behaviour and the assertion still encoded the bug this branch fixes. It failed on all three attempts in both `inprocess` and `outofprocess`. Changed to `ShouldBeEmpty()`; the test name still reads true. The other two `ShouldBeNull()` uses under `Integration/Client/for_ReadModels` are on nullable scalars (`CompletedAt`, `Value`) and are unaffected.

**Not fixed — looks flaky.** `when_getting_instance_by_key.and_projection_is_active` failed three assertions (`should_return_the_model`, `should_have_value_from_second_event`, `should_have_number_from_first_event`, all `Assert.NotNull() Failure: Value is null`) on the first `inprocess` attempt, then passed on retry — the `outofprocess` job settled at `Passed: 75 / Failed: 1` with only the assertion above outstanding. Reads like an eventual-consistency race on resolving the instance immediately after append rather than a defect in this branch, but worth a look since it is the on-demand immediate-projection path.

**Unrelated CI note.** The `integration-api` job sets `fail-on-cache-miss: false` and then copies unconditionally from `Source/Kernel/Server/bin/Release/net10.0/`, so a cache miss fails at `cp: cannot stat` rather than at the cache step. Either rebuild on a miss or fail fast there — as it stands a transient cache timeout reads as a build failure.

woksin:
## Design review round two — the four open calls, resolved

All four are now decided. Two changed my recommendation once I looked at the evidence.

### 1. A failed registration reports itself — `bfba71144`

**Correction to what I said earlier:** the outcome was not stuck "forever". `ConnectionLifecycle.Connected()` rolls `IsConnected` back when a handler throws, and `ConnectionWatchdog` calls `Connect()` again on session drop, which re-fires `RegisterAll`. Registration self-heals on reconnect.

That weakened the severity and sharpened the actual defect: `HasRun` was a latch that could not express *an attempt finished and failed*, so a waiter could not tell "still starting" from "already broken" and sat out its whole timeout in both cases. `HasRun` now means an attempt finished, `Failure` carries what stopped it, and `IsSuccess` separates the two. Artifacts that registered before the failure are reported alongside it, because a partial read side is what the consumer is holding. Re-run-on-reconnect makes this safe — each run replaces the outcome wholesale, so a stale failure cannot outlive a later success.

### 2. No default interface member — no change

I recommended one and then argued myself out of it. A default returning `NotRun` means an implementer who never overrides it reports "never registered" permanently and `WaitForRegistration` hangs to timeout against it — trading a compile error for a silent wrong answer, which is the exact anti-pattern the rest of this branch is about. `IEventStore` has 24 members and two implementers in-repo; it is consumed, not extended. The hard break stands.

### 3. Model-keyed projection lookups — `75cb30eaf`

**I was wrong to recommend narrowing `GetFailedPartitionsFor(Type)`.** On `origin/main` it resolved through the *read model* dictionary, so every caller it has passes a read model type. Narrowing it to what its parameter name says would break all of them. It stays lenient, and now says why.

The interface already had the convention I was reaching for — `GetProjectionIdForModel`, `HasFor(Type readModelType)` — so `GetStateForModel` and `GetFailedPartitionsForModel` follow it. The capability existed already; `_handlersByModelType` holds `IProjectionHandler` for model-bound projections. Only the surface was missing.

Also fixed the collision found alongside it: two projections maintaining the same read model were resolved silently by declaration order, so a model-bound projection could vanish from the index and asking for it by its read model answered about the fluent one instead. Both are still registered and both still write to that read model — now it is logged rather than silent.

### 4. One event store cache — `c3401b7c1`

This turned out bigger and better than "make the static one per-container". There were **two** caches: the client keys its own on name and namespace correctly; the registration extensions kept a second one in front of it, static and keyed on namespace alone. The second one is gone and the registration delegates, passing the namespace it resolved rather than letting the client resolve it again independently.

Removing it exposed that the client's own cache had no concurrency guarantee — it checked, then built, then assigned, so two concurrent callers each constructed an event store, each ran discovery, and the second overwrote the first, handing one caller a store that was no longer cached while the connection lifecycle held registration handlers for both. That is now a single construction per key, with a faulted one evicted so an unreachable kernel on the first call is not permanent.

**Not done:** the `OnConnected -= / +=` workaround in `ChronicleOrleansFixture` stays. I expected this to remove it, but it survives for a different reason — .NET's own scope cache holds the resolved `IEventStore` for the root scope, independent of anything Chronicle caches. Changing integration test infrastructure I cannot execute here would be guessing.

### Verification

0 warnings, 0 errors; 6564 passed, 0 failed across 17 spec assemblies. Every change mutation-proven, including restoring the old static namespace-keyed cache to confirm the new container-isolation spec catches it. Two mutation runs that reported green were discarded as stale — the mutated code had failed to build, so the assertions ran against an old binary.

woksin:
## Round three — the four follow-ups, and a release-blocking bug the gate could not see

### A net8.0 break that passed every check

`ConnectionLifecycle` used `System.Threading.Lock`, which is net9.0 and later. The client packages ship **net8.0, net9.0 and net10.0**, but that multi-target happens only under `IsPackaging=true` — and `IsPackaging` appeared in exactly one place in CI: `publish.yml`, which runs *after merge*.

So it compiled on the net10.0 the gate builds, passed 6569 specifications, passed every pull-request check, and would have failed while the release was being cut with the merge already done. It is an `object` now, with a comment saying why.

`dotnet-build.yml` now builds what `publish` packs, on the frameworks it packs them for. The workflow already installed all three SDKs — they were provisioned for exactly this and nothing used them. Verified locally: every client package plus Contracts builds clean on net8.0, net9.0 and net10.0.

### Read models are queryable by design — documented, not implemented

Ruled as by design: a read model is a shape decided in advance so one read case is answered by one lookup, and non-key querying is the storage engine's job rather than something Chronicle should take over. `Documentation/read-models/querying.md` states the boundary, what to do instead, and what going to storage directly costs — so the absence stops reading as an oversight.

Re-verified before ruling: zero `IQueryable` anywhere under `Source/Clients/`, twelve members on `IReadModels` and not one takes a filter, `GetAllInstancesRequest` carries five fields and none is a filter.

### The mirrored concepts: a guard, not a generator

The plan was to generate the client's concept copies from the kernel's. The ground work killed it. Pairing correctly — by declared namespace rather than by filename, which had been matching `Contracts` and `Api` types that merely share a name — gives **37 true duplicates out of 79 same-named pairs**. The other 42 are deliberately different types: `UniqueConstraintDefinition` at 0.195 similarity, `FailedPartition` at 0.182, the client carrying a slimmer consumer-facing shape. Generating those would have been actively wrong.

So the defect is not the duplication, it is that nothing distinguishes a copy that must stay in sync from a different type that shares a name. `EventSequenceNumber` sat at 0.977 — it looked like a copy, was treated as one, and drifted on a wire-critical sentinel with nothing to catch it.

`for_MirroredConcepts` compares them by **runtime value, not source text** — the ground work forced that too, since the copies routinely spell the same value differently (`Guid.Empty` against `new(Guid.Empty)`, a literal against a named constant). It asserts only on members both sides declare. Proven against both failure modes: reintroducing the `Max` drift and renumbering an enum member each name the offender exactly. No live drift today across 6 enum pairs and 42 constant-carrying types.

A Roslyn source generator was considered and rejected on evidence: this repository already pins `Microsoft.CodeAnalysis.CSharp` to 4.5.0 for net8.0 and 5.6.0 elsewhere, so a generator would be loaded by three different compilers for what is a text substitution.

### Subscribing a lifecycle handler is idempotent

`OnConnected` was a plain multicast event, so a caller that could not be sure whether it had already subscribed had no way to say "subscribe unless already subscribed". The integration fixture did it by removing first, with a comment recording that without the removal the handler count grows by one per test class and registration re-runs once per prior class on every reconnect. That is a hazard for anything that evicts event stores and re-resolves, not only for tests. The fixture no longer has to.

### Still open

The `ChronicleOrleansFixture` still re-attaches its handler after `EvictEventStores`. That half is .NET's own root-scope cache holding the resolved `IEventStore`, which nothing here touches.

### Verification

0 warnings, 0 errors; 6569 passed, 0 failed across 17 spec assemblies; all three shipped frameworks built clean; snippet validator green. The integration tier runs in CI only.

PR #3573, opened 2026-08-05
Release only from the branch that is released

woksin:
# Summary

Scopes the publish workflow's `pull_request` trigger to `main`, so a release can only come from the branch it is released from.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -34,8 +34,14 @@ on:
         - debug
   pull_request:
     types: [closed]
+    # Only a pull request into main releases. On "**" every base branch released, so merging one pull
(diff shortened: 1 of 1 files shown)


woksin:
Release-pipeline change only — no user-facing effect, so it carries no release-note bullets and no semver label, and publishes nothing itself.

Context: the trigger was `branches: ["**"]`, so the workflow released whenever a pull request closed into *any* base branch. Merging one pull request into another one's branch — the ordinary way to stack work — therefore cut and published a version from a branch still in review. That is how `Cratis.Fundamentals` v7.17.0 came to be published from the head of an open pull request, carrying every unmerged change on that branch under release notes describing only the one that had just merged.

Same fix as Cratis/Fundamentals#1092, applied to the remaining repositories.

PR #3574, opened 2026-08-05
Reference Cratis.Fundamentals 7.17.1

woksin:
## Changed

- References `Cratis.Fundamentals` 7.17.1, up from 7.16.8. A concept held in a collection or a value map now serializes as its underlying value rather than as `{"value": ...}`, matching how a concept serializes everywhere else

### DIFF
--- a/.github/workflows/dotnet-build.yml
+++ b/.github/workflows/dotnet-build.yml
@@ -25,6 +25,12 @@ on:
     paths:
       - "Source/**/*.cs"
       - "Source/**/*.csproj"
+      # Central package management decides what every project in the repository compiles against, so a version
+      # bump here reaches further than any single project file - and reached it without building, because a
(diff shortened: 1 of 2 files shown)


woksin:
Also adds `Directory.Packages.props` and `Directory.Build.props` to the `dotnet-build` path filters — no user-facing effect, so it is not in the release notes.

It is here because this pull request exposed the hole: a change confined to `Directory.Packages.props` matched none of the existing patterns, so the bump reported all checks green having compiled nothing. Central package management decides what every project compiles against, which makes it the widest-reaching change in the repository and, until now, the only one nothing built.

With the filter fixed the pull request builds itself — the `.NET Build & Integration` run on this branch is the real verification that `Cratis.Fundamentals` 7.17.1 is compatible.

PR #3575, opened 2026-08-05
Release when central package versions change

woksin:
## Changed

- The published packages reference `Cratis.Fundamentals` 7.17.1, up from 7.16.8. A concept held in a collection or a value map now serializes as its underlying value rather than as `{"value": ...}`, matching how a concept serializes everywhere else

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -53,6 +53,12 @@ on:
       - "Docker/**"
       # Changes to the publish pipeline itself should be exercised by a release.
       - ".github/workflows/publish.yml"
+      # Central package management decides what the published packages depend on, so a version bump here
+      # changes what a consumer restores even though no file under Source changed. Without this the bump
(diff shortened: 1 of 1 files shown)


woksin:
Adds `Directory.Packages.props` and `Directory.Build.props` to the publish path filters — no user-facing effect of its own, so it is not in the release notes.

It exists because #3574 exposed the hole. That pull request bumped `Cratis.Fundamentals` to 7.17.1 and merged green, but matched none of the publish filters, so the Publish workflow never ran and no version was cut. `main` referenced 7.17.1 while the published packages still referenced 7.16.8, and nothing reported the difference — the `verify-published` job added in #3572 catches a run that publishes nothing, but cannot catch a run that never starts.

Because this pull request touches `publish.yml`, it does trigger Publish, so merging it cuts the release that carries the 7.17.1 bump. The release note above describes that bump, since this is the release that actually delivers it.

PR #3576, opened 2026-08-05
Require a version label only where a release can be cut

woksin:
# Summary

Scopes the semantic version label check to pull requests into `main`, the only base branch Publish releases from.

### DIFF
--- a/.github/workflows/verify-semver-label.yml
+++ b/.github/workflows/verify-semver-label.yml
@@ -14,8 +14,11 @@ concurrency:
 on:
   pull_request:
     types: [opened, reopened, synchronize, labeled, unlabeled]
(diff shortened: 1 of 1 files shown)


woksin:
Release-pipeline change only — no user-facing effect, so no release-note bullets.

The check was scoped to `branches: ["**"]` while Publish releases only from `main`. Stacking one pull request onto another's branch — the ordinary way to build up work — therefore demanded a version label from a merge that cannot publish anything, which is asking which version a non-release should carry.

Same scoping as `Cratis/Fundamentals`, which already carries it.

It still needs a `patch` label itself, because the check as it stands on `main` demands one from this pull request. Nothing is published by it: the Publish path filters do not include this file.

PR #3577, opened 2026-08-05
Document join precedence over local writes and warn on the collision (CHR0042)

woksin:
# Summary

A joined projection property always wins over a local write for the same property — this states that precedence in the docs and adds an analyzer that surfaces the collision.

## Added

- CHR0042 analyzer warning when a read model property is written by both a local mapping and a join, covering the model-bound attributes and the fluent builder

## Changed

- The projection joins documentation (model-bound and declarative) now states join precedence: a joined property is applied over any value a local mapping wrote for the same property, independent of arrival order

### DIFF
--- a/Documentation/client-snippets/code-analysis/chr0042/violation.md
+++ b/Documentation/client-snippets/code-analysis/chr0042/violation.md
@@ -1,0 +1,23 @@
+```csharp
+using Cratis.Chronicle.Events;
+using Cratis.Chronicle.Keys;
+using Cratis.Chronicle.Projections.ModelBound;
+
+[EventType]
+public record Chr0042SideReleasedForSigning(Guid RequestId);
+
+[EventType]
+public record Chr0042RequestLost(bool IsContractRoundLive);
+
+// Warning CHR0042: Property 'IsContractRoundLive' is written both locally by
+// [SetValue<Chr0042SideReleasedForSigning>] and by the join with 'Chr0042RequestLost'.
+// The re-release writes true, but the join re-applies false on top — the latch
+// silently sticks and the row is withheld forever.
+[FromEvent<Chr0042SideReleasedForSigning>]
+public record Chr0042ContractSide(
+    [Key] Guid Id,
+    Guid RequestId,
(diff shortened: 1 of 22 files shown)


woksin:

**Why the precedence is real and order-independent** — traced in the kernel: `ProjectionFactory.SetupJoinsForFromDefinition` (`Source/Kernel/Core/Projections/Engine/ProjectionFactory.cs`) subscribes a `ResolveJoin(...).Project(joinPropertyMappers)` chain on the same From observable *after* the From mappers, so every qualifying local event re-reads the joined stream's last stored event and applies the join mappings on top. From mappers are never re-applied on join events. The docs sentences describe exactly this mechanism.

**Why CHR0042 is a Warning, not an Error** — the same shape is legitimate when the intent is seed-locally-then-refresh-from-the-join; suppressing at the site records that intent. Only the latch reading (expecting stream-order merge) is a defect. AutoMap-driven collisions stay with CHR0025.

**Scope notes**
- The fluent analyzer correlates only explicit `Set` mappings on both sides and scopes siblings to the same `Define` method + read model type, so a join in another projection or a child scope never false-positives. Values a join carries via AutoMap are not modelled from syntax (CHR0025's domain).
- No change to the precedence itself — existing consumers rely on join-wins.
- Pre-existing drift noticed, not addressed here: CHR0040/CHR0041 have no documentation pages; the code-analysis index previously stopped at CHR0039.

**Verification**
- Full solution builds clean, Debug and Release (the only two warnings are pre-existing in `Integration/Client/for_ReadModels/.../and_parent_event_exists.cs`, untouched by this PR).
- 12 analyzer specs pass (`for_JoinOverridesLocalWriteAnalyzer`, `for_FluentJoinOverridesLocalWriteAnalyzer`): the reporting case for both spellings plus the quiet cases (join alone, local write alone, different members, join in another projection).
- The CHR0042 violation snippet compiles via `Documentation/validate-client-snippets.py`; the docs site builds with both joins pages and the CHR0042 page rendering, 0 lint errors, and no new broken links (the 16 reported are pre-existing in unrelated pages).

PR #3578, opened 2026-08-05
Backfill documentation for CHR0040 and CHR0041

woksin:
## Added

- Documentation pages for the CHR0040 (several [SetFromContext] for the same event type on one member) and CHR0041 (event filter attribute on a projection has no effect) analyzer rules, which shipped without pages


woksin:

Docs-only follow-up to #3577 — closes the pre-existing drift where the code-analysis index and toc stopped at CHR0039 while CHR0040/CHR0041 had shipped. Page content is grounded in the analyzers' own descriptors and remarks (`DuplicateSetFromContextAnalyzer`, `InertEventFilterOnProjectionAnalyzer`), including the deliberate severity contrast with CHR0025 (Info) that CHR0040's remarks call out.

Verification: both violation snippets compile via `Documentation/validate-client-snippets.py` (checked `[SetFromContext]` is `AllowMultiple` and `[FilterEventsByTag]` targets classes so the shapes are legal C#); the docs site builds with both pages rendering, 0 lint errors, and the broken-link count unchanged (the 16 reported are pre-existing in unrelated pages).

PR #3579, opened 2026-08-05
Correct Kotlin and Java client coverage notes, and drop the no-op link check

woksin:
# Summary

Client coverage notes said Kotlin and Java lacked reactor and reducer features they now have.

## Changed

- Coverage notes on the reactor, reducer, and tutorial pages now reflect that Kotlin and Java support `@OnceOnly`, `@Replay`, event sequence targeting, passive reducers, and an `EventContext` parameter on reducer handlers.

## Fixed

- The contributing guide now states that a claim a client does not support something belongs in that client's unsupported snippet, not in shared page prose.

### DIFF
--- a/.github/workflows/markdown-verification.yml
+++ b/.github/workflows/markdown-verification.yml
@@ -15,6 +15,10 @@ on:
 permissions:
   contents: read
 
+# Links are verified by 'npm run check' in the Documentation repo, which builds
+# the aggregated site and therefore knows its routing. A crawler pointed at this
+# folder cannot resolve a site-level route (/chronicle/...) or an extension-less
+# path to a .mdx page, and reports every one of them as broken.
 jobs:
   markdown-lint:
     runs-on: ubuntu-latest
@@ -26,19 +30,3 @@ jobs:
         uses: DavidAnson/markdownlint-cli2-action@v18
         with:
           globs: 'Documentation/**/*.md'
-
-  link-verification:
-    runs-on: ubuntu-latest
-    steps:
-      - name: Checkout code
-        uses: actions/checkout@v4
-
-      - name: Check links in Documentation
-        uses: JustinBeckwith/linkinator-action@v1
-        with:
-          paths: 'Documentation/**/*.md'
-          markdown: true
-          recurse: true
-          redirects: allow
-          statusCodes: '403:ok'
-          linksToSkip: '^(https?:\/\/)?(localhost|127\.0\.0\.1)(:\d+)?(\/|$)'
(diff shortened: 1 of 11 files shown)


woksin:

**Depends on Cratis/Chronicle.Kotlin#21.** That PR adds the capabilities these notes now describe, and the matching Kotlin/Java snippets for `reactors/once-only/basic`, `reactors/replay/basic`, and `reducers/passive-reducers/reducer-attribute` — the tab groups on those three pages currently have no Kotlin or Java tab at all. Merge that first, or these pages will describe support that is not in a released client yet.

**Why the contributing-guide change is in the same PR.** The eight corrected notes are one instance of a class. The guide already defined the unsupported snippet as the mechanism for "this client can't do that", but never said it was the only place, so the same claims were also written as prose on shared pages — where nothing can invalidate them. The remaining ~45 notes are tracked separately (see the issue linked below); this PR only corrects the ones that are now factually wrong and writes down the rule.

**Verification.** `Documentation/verify-markdown.sh` — markdownlint clean across 819 files. Note that its link step reports "scanned 0 links" and passes regardless: the `--skip` pattern excludes localhost, but linkinator serves the local files over `http://localhost:<port>/`, so the entire crawl is skipped. Chronicle.Kotlin hit and fixed this same bug in 57459a5. Filed separately with the fix. The two intra-page anchors added here (`#using-the-reactor-attribute`, `#using-the-reducer-attribute`) were therefore checked by hand against the headings on those pages.

woksin:
Consolidated the link-check work into this PR rather than opening a second one (third commit).

**Correction to what I reported earlier.** I initially said 4 links were genuinely broken (`code-analysis/CHR0040/41/42`, `concepts/observers/`). That was wrong — I had checked them against a stale worktree. All four resolve on `main`. The real count is **0 broken links out of 1020**; all 177 reports are the crawler failing to model the site's routing:

| Category | Count |
|---|---|
| Site-level routes (`chronicle/…`, `event-modeling/`) | 52 |
| Extension-less links to `.mdx` pages | 118 |
| Extension-less / trailing-slash links to `.md` pages | 7 |
| Genuinely broken | **0** |

That is why this removes the check rather than repairing it — there is nothing for it to catch, and repairing it means reimplementing the docs site's routing in a shell script and keeping it in sync. `npm run check` in the Documentation repo already gates on links with the real routing. #3581 has the full analysis and closes with this PR.

**One thing to confirm before merging:** whether `link-verification` is configured as a required status check on `main`. I could not read the branch protection settings. If it is required, it needs removing there too or PRs will block waiting for a job that no longer runs.

**Also worth a separate look:** `markdownlint-cli2 "Documentation/**/*.md"` lints 816 `.md` files and skips all 181 `.mdx` files — most of the narrative docs are unlinted. Left alone here since widening the glob will surface its own backlog.

**Verification.** `Documentation/verify-markdown.sh` passes, markdownlint clean across 819 files. The workflow parses with one job remaining (`markdown-lint`).

PR #3582, opened 2026-08-06
Write down the value contract every client has to meet

woksin:
## Added

- A page stating the value contract every Chronicle client has to meet - the wire form of guids, dates, times, durations and geospatial values, how concepts and derived types serialize, and what an event type schema has to get right. It also records why there is no Kotlin or Elixir Fundamentals package

### DIFF
--- a/Documentation/contributing/clients/toc.yml
+++ b/Documentation/contributing/clients/toc.yml
@@ -2,6 +2,8 @@
   href: ./index.mdx
 - name: Client Types
   href: ./client-types.md
+- name: The Value Contract
+  href: ./value-contract.md
 - name: .NET
   href: ./dotnet.md
 - name: REST API
--- a/Documentation/contributing/clients/value-contract.md
+++ b/Documentation/contributing/clients/value-contract.md
@@ -1,0 +1,92 @@
+# The Value Contract
+
+Every Chronicle client — C#, TypeScript, Kotlin, Java, Elixir — writes and reads the same values over the
+same wire. The kernel does not adapt to the client that happens to be calling it, so the client is what has
+to agree.
+
+This page is that agreement. It is deliberately short: the contract is small, and the reason it is written
+down is that until now it was not, which is how a client comes to diverge from it without anyone noticing.
+
+## What has to agree
+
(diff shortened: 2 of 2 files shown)


woksin:
Reviewer notes.

**Why now.** Two defects fixed in the last day are the same class, and neither is a duplication problem — so no shared library would have caught either:

- `Chronicle.Kotlin` registered every event type with `.setSchema("{}")`, so every Kotlin-registered event type landed with zero properties (Cratis/Chronicle.Kotlin#3, fixed in Cratis/Chronicle.Kotlin#22).
- `@Pii` in the Kotlin client emits no compliance metadata, so it is inert.

Both are a client disagreeing with the kernel about a shape nothing had written down.

**Everything on the page is read from source, not from memory.** The temporal forms come from `@cratis/fundamentals` (`DateOnly.ts`, `TimeOnly.ts`, `TimeSpan.ts` — the last documents its own form as "C# compatible"), and `_derivedTypeId` is confirmed identical on both stacks: `DerivedTypeJsonConverter.DerivedTypeIdProperty` in .NET and `JsonSerializer.DerivedTypeIdProperty` in TypeScript.

**One correction the page makes explicitly.** `_t` is easy to mistake for the derived-type discriminator and is not — it is MongoDB's, written by the kernel's storage layer, and no client sees it. I nearly wrote it into the contract myself before checking; the callout is there so the next person does not.

**The claim worth a second opinion** is the empty-schema one. The Kotlin fix traced the kernel and found the converter falls back to raw conversion when a schema declares no properties, so `{}` does not corrupt reads — it just registers an event type with nothing in it. The page states that distinction rather than the simpler "empty schema breaks reads", which was the earlier and wrong understanding on the issue.

**What this does not do.** It writes the contract down; it does not check it. The last line points at the conformance suite that would, which is filed separately as #3583 — that is the piece that turns this from a page someone has to remember to read into something CI enforces.

**Verification.** `markdownlint-cli2` clean. Registered in `contributing/clients/toc.yml` after Client Types. No shared-page client-language fences added, so `chronicle-client-docs:check` is unaffected.

PR #3584, opened 2026-08-06
Close the issues a shipped pull request resolves

woksin:
# Summary

Shipping a change now closes the issues it resolves, instead of only referencing them and leaving them open.


woksin:
Reviewer context: the rationale for closing after the merge rather than with a `Closes #N` keyword is that the published release notes are the pull request description verbatim, so the keyword would ship into the changelog. The new step verifies the issue state flipped, comments instead of closing when a change only partly addresses an issue, and reports a failed close rather than leaving it silent.

PR #3585, opened 2026-08-06
Release unique event type constraints and fix PII read model reads

woksin:
## Added

- `RemovedWith<TEventType>()` on the constraint builder, so a per-event-source unique constraint can express "at most one per cycle" rather than only "at most one ever"
- `[RemoveConstraint]` now releases a unique event type constraint

## Fixed

- A `[Passive]` read model carrying a `[PII]` property can be read; every keyed read previously threw
- A watched read model carrying a `[PII]` property streams its changesets; every one was previously dropped silently

### DIFF
--- a/Documentation/client-snippets/constraints/declarative/unique-event-type/releasing.md
+++ b/Documentation/client-snippets/constraints/declarative/unique-event-type/releasing.md
@@ -1,0 +1,21 @@
+```csharp
+using Cratis.Chronicle.Events;
+using Cratis.Chronicle.Events.Constraints;
+
+[EventType]
+public record ConstraintsUniqueEventTypeShiftStarted(string Location);
+
+[EventType]
+public record ConstraintsUniqueEventTypeShiftEnded;
+
+public class ConstraintsUniqueEventTypeOneOpenShift : IConstraint
+{
+    // At most one open shift per employee. Ending the shift releases the constraint,
+    // so the next shift is allowed - without it the constraint could only say
+    // "at most one, ever", and the employee's second shift would be refused forever.
+    public void Define(IConstraintBuilder builder) =>
+        builder
+            .Unique<ConstraintsUniqueEventTypeShiftStarted>()
+            .RemovedWith<ConstraintsUniqueEventTypeShiftEnded>();
+}
+```
--- a/Documentation/concepts/designing-read-models.mdx
+++ b/Documentation/concepts/designing-read-models.mdx
@@ -34,6 +34,30 @@ It's tempting to make one `Customer` read model serve every screen. Resist it. T
 
 You rarely write code that *updates* a read model — you declare how events map onto it and let Chronicle do the folding. Start with [model-bound projections](../projections/model-bound/), where attributes on the read model record describe the mapping. When the mapping needs logic the attributes can't express cleanly, step up to the fluent [declarative projection builder](../projections/declarative/) (`IProjectionFor<T>`), and only reach for a [reducer](../reducers/) when the model is genuinely easier to express as code folding over previous state.
 
+## The constructor may not run
+
+Sooner or later you'll want to put a small piece of logic in the read model record itself — a normalizer that turns a `null` collection into an empty one, a computed default, a guard that rejects a value that should never arrive:
+
+```csharp
+// Don't rely on this. It may never execute.
+public record OrderSummary(OrderId Id, IEnumerable<LineItem> Lines)
+{
+    public IEnumerable<LineItem> Lines { get; init; } = Lines ?? [];
+}
+```
+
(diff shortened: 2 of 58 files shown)


woksin:
Reviewer context. This groups four related pieces of work; each was developed on its own branch and merged here.

**Passive + PII (the keyed read).** `GetInstanceByKey` falls through to the immediate projection for a `[Passive]` read model, and that path stamps `__subject` into the projected JSON then hands it to the compliance manager, which rejects any property the schema does not declare. The sink-backed path never hits this because it round-trips through `ToJsonObject(instance, schema)` first, which carries only schema-declared properties — so it was self-consistent by accident of the round trip, not by design. Fixed by stripping bookkeeping on a clone before the compliance walk. Reproduced first: the spec fails without the fix with the exact `CompliancePropertyNotFoundInSchema` for `__subject`.

**Watched + PII (the silent one).** Found while tracing the above. `ProjectionObserverSubscriber.OnNext` does the same schema round trip and then stamps `__lastHandledEventSequenceNumber` back on top of it, unconditionally. `OnChangeset` is `[OneWay]`, so the resulting throw never surfaces — the observer offset had already advanced and the changeset is gone. No error anywhere; the client's observable simply stops updating. Only `[PII]`-carrying read models are affected, because `HasComplianceMetadata()` short-circuits otherwise, which is why it went unnoticed.

The strip is matched against `schema.GetFlattenedProperties()` rather than the `__` prefix. The prefix is not reserved — `Storage.Sql/.../ProjectedColumns.cs` explicitly anticipates read models that expose `__lastHandledEventSequenceNumber` as their own property — so a prefix rule would delete data a read model asked for.

Worth flagging for a follow-up, not fixed here: the stamping itself is the deeper problem. The subscriber does a schema round trip to produce a schema-shaped document, then writes two properties back on top, then hands it to a consumer that validates against that same schema. The changeset contract already carries the sequence number out of band (`ReadModelChangeset.EventSequenceNumber`), and the client reads it from there, never from the JSON — so the stamp is redundant for every read model that does not declare the property. `ReadModelsCompliance.Apply`'s carry-through loop is a third compensation for the same mismatch.

**Constraint release.** The client already resolved `[RemoveConstraint]` onto the definition; `ToContract` then dropped it, so it type-checked, serialized, versioned and did nothing. The contract already had a `RemovedWith` member (the value form uses it), so no proto change was needed. Now carried through client → contract → kernel concept → validator → all three storage implementations (MongoDB, in-memory, SQL — the SQL one was not in the original scope and was found during the work).

The storage interface takes the whole definition rather than an added parameter deliberately: `EventTypeId` has an implicit `string` conversion, so an `EventTypeId?` beside `string scopeKey` would silently bind positional scope keys instead of failing to compile.

Two known limitations, both reported rather than fixed:
- Intra-batch release is not honored. `AppendMany` validates the whole batch before writing, so a removal event earlier in the same batch is not visible to a later covered event. The value form solves this with `ConstraintBatchClaims`, which is value-keyed and does not transfer.
- `scopeKey` is ignored by all three unique-event-type storage implementations — passed in, never used to narrow. **This is pre-existing on `main`, not introduced here**, and it means `PerEventSourceType` / `PerEventStreamType` / `PerEventStreamId` scoping silently does nothing for this constraint kind. Kept at parity rather than fixed, since it is a separate defect that deserves its own change.

**Documentation.** Adds the reach of a MongoDB convention pack (it registers into the driver's global registry, so it does not cover Chronicle's own read-model deserialization), and that a read model record's constructor may not run. That second one is measured, not inferred: against MongoDB.Driver 3.10.0 on .NET 10, a record with a primary constructor gets no creator map at all — `HasCreatorMaps=False` — so the constructor runs zero times for an absent field, an explicit null and a fully present value alike. A record-body normalizer is dead code on that path, which makes it wider than the collection question. System.Text.Json, which is what every Chronicle read path uses, does run it. Also scopes the Kotlin/Elixir initial-values statement to definition-build time so it stops reading as a contradiction.

**Verification on the merged branch, not just the parts.** `dotnet build Chronicle.slnx --no-incremental`: 50 projects, 0 errors, 2 warnings — both `SA1515`/`RCS1181` in `Integration/Client/for_ReadModels/when_getting_instance_for_passive_projection/and_parent_event_exists.cs`, a file untouched here and already warning on `main`.

| Project | Result |
| --- | --- |
| Concepts.Specs | 135 passed |
| Core.Specs | 2236 passed, 1 skipped |
| Storage.InMemory.Specs | 109 passed |
| Storage.MongoDB.Specs | 617 passed |
| DotNET.Specs | 2127 passed |
| Testing.Specs | 335 passed |

Core.Specs reconciles exactly: 2217 on `main` + 3 + 10 + 6 for the three spec-adding branches. Every new spec was confirmed to fail without its fix. MongoDB-backed integration specs ran against a real container during development; the one pre-existing `for_JobsManager` failure set was confirmed identical on `main` in a throwaway worktree.

`Storage.Sql` has no coverage of `UniqueEventTypesConstraintsStorage` — before or after — so that implementation is compile-verified only.

Issue #3580, opened 2026-08-05
Client capability claims in shared docs cannot be maintained where they are written


woksin:
## The problem

46 shared Chronicle documentation pages assert what a specific client cannot do — "this is currently C#-only", "Kotlin and Java have no equivalent", "not shown for that client".

None of those claims can be maintained where they are written. A shared page lives in this repo; the capability lives in a client repo, on its own release cadence, with no CI here that fails when the claim stops being true. So the claim rots silently, and the first person to notice is a reader who believed it.

That is not hypothetical. Cratis/Chronicle.Kotlin#21 added `@OnceOnly`, `@Replay`, event sequence targeting, passive reducers, and an `EventContext` parameter for reducer handlers. Eight notes across the reactor, reducer, and tutorial pages instantly became wrong, and nothing in this repo would have caught it — the person who found them was a reader asking why the docs said the Kotlin client couldn't read events. #3579 corrects those eight and writes down the rule; this issue is the rest of the class.

## The mechanism already exists

`contributing/clients/index.mdx` defines it: when a client does not support a workflow, it ships an explicit unsupported snippet, in its own repo, at the same snippet id. The tab stays visible and carries the message.

That puts the statement next to the code, in the repo whose CI compiles the snippets, edited by the same person who adds the capability. It cannot go stale unnoticed the way prose can.

## What to do

1. For each of the 46 pages, decide per claim:
   - **Claim of absence** ("client X can't do this") → delete from the prose. If the tab group is missing that client, add an unsupported snippet in the client repo instead.
   - **Description of difference** ("TypeScript reaches the same result through the decorator's parameter") → keep. These are useful and don't rot the same way.
2. Where a claim of absence has no tab group to carry it, the page needs one — that is the real gap, not the sentence.
3. Enforce it. `npm run chronicle-client-docs:check` in the Documentation repo already ratchets direct-language fences in shared docs. Add a second ratchet over client-name-plus-negation phrasing in shared pages, baselined at the current count and allowed only to decrease.

Step 3 is what makes this stick. Without it the prose comes back the next time someone wants to explain a missing tab.

## Notes

- Verifying each claim means checking the current source of five clients, so this is best done per area (reactors, reducers, projections, read models, compliance) rather than in one pass.
[...]

PR #3586, opened 2026-08-06
Correct coverage notes for Kotlin and Java tagging and filtering

woksin:
# Summary

Coverage notes said Kotlin and Java lacked tagging, filtering and subject support they now have.

## Changed

- Coverage notes on the subject, tagging and filtering pages now reflect that Kotlin and Java support an append-time subject and tags, `@Tag` on reactors and reducers, and observer filtering by tag, event source type and event stream type.

### DIFF
--- a/Documentation/concepts/subject.mdx
+++ b/Documentation/concepts/subject.mdx
@@ -37,7 +37,7 @@ an implicit conversion from `EventSourceId`, so mixing the two is painless:
 
 <ChronicleClientTabs snippet="concepts/subject/explicit" />
 
-`Subject` also converts implicitly from `string` and `Guid`. Kotlin, Java, and TypeScript don't currently expose a `subject` append option at all — only C# and Elixir do.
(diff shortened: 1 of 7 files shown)


woksin:

**Follows Cratis/Chronicle.Kotlin#34 and #35**, both merged. Eight claims across seven pages became false when those landed.

**What I deliberately did not flatten.** These notes still carry real distinctions, so they were narrowed rather than deleted:

- static tags on **events** and **read models** are still C#-only — Kotlin/Java tag the *observer*, not the event type
- labelling a **projection** is still C#-only; Kotlin/Java have it for reactors and reducers only (Cratis/Chronicle.Kotlin#26 covers the projection side)
- the declarative `[Subject]` property is still C#-only; Kotlin, Java and Elixir pass the subject per append
- TypeScript still has neither a subject nor a tags append option

**This is the third such sweep**, which is the point of #3580. Each Kotlin release invalidates more of this prose, and nothing in this repo can detect it — the claims live here, the capability lives in another repo on another cadence. The ratchet proposed in #3580 (step 3) is what stops a fourth.

**Verification.** `verify-markdown.sh` passes. Every claim was checked against the merged Kotlin source, not against other docs pages.

PR #3594, opened 2026-08-06
Point the absent-collection driver advice at Arc's convention pack

woksin:
## Changed

- The empty child collections page points at Arc's convention pack for the `IMongoCollection<T>` path instead of telling you to configure the driver yourself, and names the record-constructor trap that path carries


woksin:
Reviewer context.

Follow-up to Cratis/Arc#2457, which shipped a convention pack materializing a non-nullable collection on a `[ReadModel]` type as empty for both an absent field and a stored `null`. That made this page's existing advice — "configure that on the driver" — stale: it was written when nothing configured it, and now sends a reader off to build something Arc already registers for them.

The second half is a trap the page never named. Measured against MongoDB.Driver 3.10.0 on .NET 10: a record with a primary constructor gets `HasCreatorMaps=False`, so the driver materializes it uninitialized and assigns members directly, on every document rather than only an incomplete one. A record-body normalizer (`Lines = Lines ?? []`) is therefore dead code on this path — and it is the first fix a competent reader reaches for, in-language and type-checking. A consumer deployed exactly that, measured the property still `null`, and concluded the absence was not real. Worth stating on the page where they land.

Both links verified against the built content: the `#the-constructor-may-not-run` anchor exists in `concepts/designing-read-models.mdx`, and `/arc/backend/mongodb/convention-packs/` documents the pack on Arc's `main` post-merge.

PR #3595, opened 2026-08-06
Ship a documentation-only pull request without a label or a wait

woksin:
# Summary

AI instruction corpus only — no user-facing change.

### DIFF
--- a/.ai/rules/general.md
+++ b/.ai/rules/general.md
@@ -144,7 +144,7 @@ Tagged **[contract]** (framework-enforced) or **[convention]** (house default). 
 | Frontend | test | zero failures when frontend specs/behavior changed |
 | Frontend | build | zero errors |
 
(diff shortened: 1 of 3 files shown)


woksin:
Reviewer context.

Records a decision made while shipping the Ada upstream backlog: a documentation-only PR carries no semantic version label and is merged without waiting for CI.

Two reasons. Nothing a docs change touches can break a build, a spec or a lint, so polling its checks buys nothing. And labelling it `patch` cuts a release for content that ships no artifact — the *absent* label is the correct state, because it is what makes the Publish run skip every publish step.

The corpus contradicted this in three places, all corrected here:
- `pull-requests.md` listed `docs` under `patch`, and its quality gates demanded green before any merge.
- `general.md` said all gates pass before merging, with no carve-out.
- `ship-changes` step 7 said to poll until every check passed.

The consequence worth naming: `verify` reds for the missing label on these PRs, and that now reads as the expected outcome rather than something to chase.

Also folds in two things learned the hard way during this batch. `gh pr edit --add-label` fails on the Projects-classic GraphQL deprecation and does so quietly enough that a body edit silently did not apply — `gh api -X POST .../labels` is the way round it. And a `verify` run that starts before the label lands sees `LABELS: []` and fails, so labelling wants to happen before the first run rather than after.

Edited only the canonical `.ai/` sources; `.github/` and `.claude/` are symlink adapters and were not touched, per `managing-ai-rules.md`.

Merged unlabelled and without waiting, per the rule it introduces.

PR #3599, opened 2026-08-07
Harden projection, compliance, seeding, and client safety

woksin:
# Summary

Chronicle 16.20 makes projection removal, PII handling, and seeding reliable under production failure modes while keeping existing reactor and seeding contracts compatible.

## Added

- Add result-aware seeding and a fresh buffer for retrying corrected definitions without changing existing `IEventSeeding` implementations. (#3612)
- Add `[ReleaseUnder]` support for composing a read model from PII owned by different subjects. (#3601)
- Add CHR0043 and CHR0044 diagnostics for key redirection and joins that can store PII under an unprovable subject. (#3640)
- Add event-store-aware reactor side-effect classification while retaining the previous handler contract. (#3629)

## Changed

- Validate TLS certificates by default; the Development connection string remains the explicit self-signed-certificate opt-out. (#3606)
- Apply `AutoMap` and `[NoAutoMap]` consistently in root, child, and nested projection builders and diagnostics. (#3624)
- Scope client event serialization and side-effect classification to the executing event store. (#3629)

## Fixed

- Keep removed immediate and passive read models absent across cached reads, previews, joins, no-op events, and testing; only a constructive later event can recreate them. (#3608)
- Preserve encrypted enum, nullable-enum, boolean, and numeric PII values through JSON and MongoDB storage and release. (#3603)
- Preserve failed seed batches for retry and retain repeated and tag-distinct events in author order. (#3612)
- Surface projection-processing failures instead of returning an empty model. (#3645)
- Honor `[JsonPropertyName]` when `[ReleaseUnder]` resolves a PII subject property. (#3601)
- Prevent root-level MongoDB joins from matching every document or throwing when a key is null or cannot be converted. (#3628)
- Make client testing artifact discovery deterministic and isolate fresh seeding buffers from previous failures. (#3646, #3612)

## Security

- Avoid logging compliance-subject identifiers during cross-store key propagation and unmatched projection joins. (#3647)
- Diagnose cross-subject PII joins and key redirection before they make erasure incomplete. (#3640)
- Ensure omitted client TLS settings cannot silently disable certificate validation. (#3606)


woksin:
Reviewer context, kept out of the body so it does not reach the release notes.

**What is in here.** Six upstream reports from a production Cratis application: two root-join defects in the MongoDB sink, the captive `IEventTypes` in the reactor side-effect handlers, two seeding defects, the cross-event-store encryption key seam, and the composed-read-model PII release seam. Each was reproduced before being fixed, and every spec was mutation-proven — reverted, confirmed red, restored.

**Why `major`.** Three public constructors lost a parameter, and the deprecated `CanHandle` overload throws where it previously answered. Restoring those constructors was attempted and rejected: Microsoft DI honours neither `[ActivatorUtilitiesConstructor]` nor `[Obsolete]` and picks the greediest resolvable constructor, so restoring them silently reinstates the captive dependency for every consumer. A compile-time break was judged better than a silent runtime one.

**Verification.** `Chronicle.slnx` builds clean in Debug and Release, zero warnings. Unit gate: 6,934 passed. Two failures in `Testing.Specs` `for_EventStoreForTesting.when_registering_all` are a pre-existing parallel-execution flake — the same two fail identically on a clean `origin/main` worktree, and a differing set fails run to run. Integration specs for the PII, subscription, seeding and revise paths pass against a real kernel and MongoDB.

**Two design questions this PR deliberately does not answer**, both recorded as design notes on the branch:
- Is erasing a subject in an (event store, namespace) pair a permanent refusal to hold that subject's key there again, or only removal of the key that exists now? Answer "permanent" and a re-registering person's new PII is silently blanked.
- Should a release-exempt PII marker exist, and where should `[ReleaseUnder]` on a stored read model be refused? Not at registration — `RegisterAll()`'s throw reaches ASP.NET Core through `ApplicationStarted.Register(...)`, where a throwing callback is logged rather than fatal, so the refusal would be invisible.

**Follow-up not in scope.** `dotnet build Chronicle.slnx` intermittently hits `MSB3231`/`MSB3073` on `Source/Kernel/Server/out`: `Integration/Api` and `Integration/Client` both wipe and republish that directory in parallel. It is a build-graph race that predates this work and deserves its own fix.

woksin:
The current `major` label is justified by the public breaks as written, but there is a concrete compatibility path that preserves all six fixes and lets this remain a `minor` release. The owner preference is to avoid a major train when compatibility can be retained without weakening correctness.

### Reactor handlers and `CanHandle`

- Mark only `EventResultHandler`, `EventsResultHandler`, and `MixedSideEffectsResultHandler` with `[IgnoreConvention]`.
- Register each singleton manually by factory in `AddCratisChronicleClient` (`TryAddSingleton<EventResultHandler>(_ => new())`, etc.). Fundamentals discovery skips their self-binding but still includes them in the contract-to-implementor map, so `IInstancesOf<IReactorSideEffectHandler>` continues to resolve them.
- Retain obsolete public `(IEventTypes)` constructors into a `_legacyEventTypes` field and add parameterless constructors for the factory. The new `CanHandle(context, eventStore, value)` always uses `eventStore.EventTypes`, preserving the captive-dependency fix.
- Retain explicit old `CanHandle(context, value)` adapters on built-ins. Legacy-constructed instances use `_legacyEventTypes`; framework-created instances use an additive `ReactorContext.EventStore` populated by `ReactorInvoker` and forward to the new overload.
- Keep the new-to-old interface default only for third-party legacy implementers. Do not add bidirectional defaults, which recurse for an implement-neither type. The dispatcher old overload forwards through `context.EventStore`.

Specs should pin ValidateScopes resolution/enumeration, singleton factory descriptors rather than implementation-type descriptors, old constructor plus old method behavior, old dispatcher equivalence across two event stores, and a third-party old-only handler.

### Seeding

Keep the public Orleans `Task Seed(IEnumerable<SeedingEntry>)` signature and add `Task<SeedingResult> SeedWithResult(...)`. `Seed` delegates to the same corrected core and discards the result; internal/global coordination uses `SeedWithResult` so incomplete batches remain untracked and are redispatched.

Specs should reflect-pin the old signature, prove old `Seed` still gets the rejection/retry/order corrections, prove `SeedWithResult` reports incomplete work, and prove the global grain withholds tracking and retries.

This avoids the three removed public constructors, old `CanHandle` behavior break, and `Task` → `Task<SeedingResult>` signature change without restoring the captive dependency or weakening the seeding result path. If this compatibility shape is adopted, the PR can move to `minor`; until then `major` remains honest.

PR #3653, opened 2026-08-10
Fix three defects shipped in v16.20.0

woksin:
# Summary

Three defects shipped in v16.20.0, found by a post-merge review of #3599.

## Fixed

- A rejected seed batch no longer fails the client's connection. Seeding runs inside `RegisterAll` from `OnConnected`, so the exception rolled the connection back to disconnected; because a seed rejection is deterministic, the watchdog reconnected and reproduced it indefinitely. The rejection is logged by the kernel and the batch is left unseeded for a corrected run.
- `ReviseEvent` in `Cratis.Chronicle.XUnit.Integration` no longer resolves `IEventSerializer` from the root provider. That service became scoped in 16.20.0, so the call either threw or returned a serializer bound to a different event store's event types.

## Security

- The seeding rejection log no longer writes event source ids or constraint violation messages, both of which carry the compliance subject. It names the violated constraints and reports counts instead.

### DIFF
--- a/Integration/Client/for_EventSeeding/when_a_seeded_batch_is_rejected.cs
+++ b/Integration/Client/for_EventSeeding/when_a_seeded_batch_is_rejected.cs
@@ -69,7 +69,14 @@ public class when_a_seeded_batch_is_rejected(context context) : Given<context>(c
             await EventStore.EventLog.GetFromSequenceNumber(EventSequenceNumber.First);
     }
 
-    [Fact] void should_fail_the_rejected_run() => Context.FirstFailure.ShouldNotBeNull();
+    /// <summary>
+    /// The rejected run must NOT surface as an exception. Seeding happens inside <c>RegisterAll</c>, which the
+    /// client runs from <c>OnConnected</c>; a failing handler there rolls the connection back to disconnected,
+    /// the watchdog reconnects, and the same deterministic seed rejection repeats - a permanent client outage
+    /// caused by a fixable mistake in seed data. The rejection is reported by the kernel grain's Error log, and
+    /// the batch is left unseeded so a corrected run retries it, which the facts below pin.
+    /// </summary>
+    [Fact] void should_not_fail_the_rejected_run() => Context.FirstFailure.ShouldBeNull();
     [Fact] void should_append_nothing_from_the_rejected_run() => Context.EventsAfterTheRejectedRun.ShouldEqual(0);
     [Fact] void should_append_the_corrected_seed_set() => Context.EventTypesAfterTheCorrectedRun.Count().ShouldEqual(2);
     [Fact] void should_append_the_innocent_entry_that_shared_the_rejected_batch() => Context.EventTypesAfterTheCorrectedRun.ShouldContain(nameof(OfficeOpened));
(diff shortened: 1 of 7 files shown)


woksin:

Found by a post-merge adversarial review of #3599. Three of that review's findings are fixed here; the rest are tracked separately.

**1 — the seeding throw.** The grain at `Source/Kernel/Core/Seeding/EventSeeding.cs:178-185` deliberately declines to throw and says why: a rejected batch is almost always a deterministic constraint violation in the seed set, so throwing would fail on every start. The gRPC service one layer up threw anyway. Chain, each hop read in the shipped tree: `Services/Seeding/EventSeeding.cs` → client `Seeding.Register()` → `EventStore.RegisterAll()` → `Connection.Lifecycle.OnConnected` → `ConnectionLifecycle` sets `IsConnected = false` and throws → `ChronicleConnection` catches only `TimeoutException` → propagates out of `Connect()`. Because the rejection is deterministic, the watchdog reconnects and reproduces it. The fix removes the service-layer throw; the grain's Error log remains the signal, and the batch stays unseeded so a corrected run retries it.

Note the first fix attempt was wrong and is worth recording: catching `EventSeedingIncomplete` client-side does not work, because that type lives in the kernel and the client cannot reference it — over gRPC it is not the typed exception. The service layer is the only place this can be fixed.

**Three specs pinned the old behaviour and are inverted here**, each with a comment saying why: the two `Core.Specs/Services/Seeding/for_EventSeeding/*/and_the_grain_reports_incomplete_seeding.cs`, and `Integration/Client/for_EventSeeding/when_a_seeded_batch_is_rejected.cs`. The integration spec's other three facts are unchanged and still pass — nothing appended from the rejected run, the corrected run lands, and the innocent entry that shared the rejected batch lands with it. Those are the facts that matter, and they now hold without the outage.

**2 — the shipped test package.** `EventSequenceMutationExtensions.cs:38` resolved `IEventSerializer` from `fixture.Services`, the root provider. CHR-48 moved that registration to `TryAddScoped` in the same release, so the call throws under `ValidateScopes` and otherwise hands back a serializer bound to a different event store's `IEventTypes` than the one being mutated. Now taken off the event store directly.

**3 — subjects in the log.** `SeededEventsRejected` wrote `ConcurrencyViolations.Select(_ => _.EventSourceId.Value)` in clear, and the constraint violation messages embed the event source id too (`UniqueEventTypeConstraintValidator.cs:45`). An event source id is the compliance subject by default, so both survive that subject's erasure. The log now names the violated constraints and reports counts. `UniqueConstraintValidator` in the same folder already kept the offending value out of its message; this path bypassed that discipline.

**Verification.** `Chronicle.slnx` Debug and Release both clean, zero warnings. Unit gate: 7,095 passed, 0 failed.

**Not in this PR, from the same review:** IMP-26 was silently reverted before #3599 merged and shipped nothing — restored separately. CHR0044 fires on the correct PII-join declaration and breaks Release builds under `TreatWarningsAsErrors`. `JoinHasNoKey` and `JoinKeyNotConvertible` are Debug but both describe a lost write. Four analyzer specs pass on a substring collision. The v16.20.0 release notes understate the upgrade impact.

woksin:
Folded into #3654 so the repo cuts a single release, per the one-PR-per-repo convention. Every commit here is included there, and the combined branch is verified on current main: 7,254 unit specs and 155 integration specs, all passing.

Issue #3639, opened 2026-08-10
[Resolved in Chronicle 16.22.0] Configure composite encryption-key storage for cutover


woksin:
> **Resolution:** Fixed by [PR #3654](https://github.com/Cratis/Chronicle/pull/3654) and released in Chronicle v16.22.0; this issue is closed.

## Historical filing context

This report identified a missing configuration seam for the existing `CompositeEncryptionKeyStorage` component in `Cratis.Chronicle`.

## Delivered behavior

`Compliance:Encryption:MigrateFromDefaultStorage` wires the dedicated store and default store through the composite during cutover. The Vault and Azure Key Vault extensions honor the option.

## Consumer impact

The configuration lets a consumer perform a controlled encryption-key-store cutover without supplying a custom implementation of the composite itself.

## What is explicitly not being asked for

This closed report does not prescribe a key-store migration policy, enable a migration by default, or replace a consumer's backup and operator safeguards.

woksin:
Fixed in #3654 and released in v16.22.0.

Issue #3650, opened 2026-08-10
Keep Orleans MongoDB singleton registrations scope-safe


woksin:
> **Resolution (2026-08-10):** This issue is closed as completed. The maintainer confirmed: “Fixed in [#3648](https://github.com/Cratis/Chronicle/pull/3648).” The report below describes the pre-fix failure; its filing-time evidence limit is superseded by that resolution.

## Summary

After Chronicle adopted Arc 21.1.0, every out-of-process + MongoDB client integration shard aborts its Chronicle server before scenarios run because Orleans' singleton Mongo client factory consumes Arc's scoped `IMongoClient`.

## Reproduction

Chronicle CI run `31375257151` fails all 17 out-of-process + MongoDB client shards. The server container exits with code 134 and the shared root exception is:

```text
Cannot consume scoped service 'MongoDB.Driver.IMongoClient' from singleton
'Orleans.Providers.MongoDB.Utils.IMongoClientFactory'.
```

The in-process counterparts, MongoDB unit integration, builds, specs, benchmarks, and kernel shards pass. The only functional current-main change since the previously green integration run was Arc 21.0.4 -> 21.1.0, which preserves host scope validation and exposes this pre-existing invalid lifetime chain.

## Impact

- Chronicle's entire out-of-process MongoDB client integration matrix cannot start.
- The current release train cannot prove the production server topology.
- Disabling scope validation would hide an invalid singleton-to-scoped dependency and weaken the production container contract.

## Expected behavior

Orleans should receive a singleton-safe Mongo client/factory built from the configured Mongo settings. Arc's scoped/proxied `IMongoClient` registration should remain scoped for Arc consumers; Orleans' singleton must not capture it.

## Acceptance

- A service-provider-validation spec proves the Orleans registration graph is valid with scope validation enabled.
- Orleans' singleton Mongo client factory no longer consumes Arc's scoped `IMongoClient`.
- One out-of-process + MongoDB client shard starts and passes as a focused proof.
- All 17 out-of-process + MongoDB client shards pass in the full integration workflow.
- No lifetime is weakened and scope validation remains enabled.

## Evidence limits

The exception and 17-shard classification are observed in CI. The proposed registration split is source-supported but not yet implemented or run.

woksin:
Fixed in #3648.

Issue #3608, opened 2026-08-09
Passive immediate projections retain read models after RemovedWith events


woksin:
## Describe the bug

A passive read model that is rebuilt directly from an event source remains present after an event configured through `RemovedWith<TEvent>`. The same projection definition removes a sink-backed document, but `IReadModels.GetInstanceById<TReadModel>()` over the passive/immediate path returns the pre-removal state.

This is security-relevant for passive command/reactor policies: a revoked or consumed grant can remain actionable even though its removal event is in the log.

## Steps to reproduce

1. Define a keyed `[Passive]` read model projected from a creation event and decorated with `[RemovedWith<Revoked>]`.
2. Append the creation event to one event source.
3. Read the passive model through `IReadModels.GetInstanceById` and observe it is present.
4. Append `Revoked` to the same event source.
5. Read the passive model again.

A second control using a different terminal event registered with `RemovedWith` produces the same result. Creation and field projection succeed; only removal fails.

## Expected behavior

The read after the terminal event returns no read model, matching the projection definition and sink-backed removal semantics.

## Actual behavior

The read returns the fully populated state from before the terminal event.

A kernel-backed reproduction against Chronicle 16.19.2 produced four facts: creation and pre-removal controls passed; both post-removal assertions failed.

## Source diagnosis

`Source/Kernel/Core/Projections/Projection.cs` rebuilds immediate/passive state with `ApplyActualChanges`. Its switch applies `PropertiesChanged`, `ChildAdded`, `Joined`, and `ResolvedJoin`, but has no `Removed` arm. The sink paths separately check `changeset.HasBeenRemoved()`. The same omission is present on the current inspected branch at commit `7cd011983`.

## Environment

- OS: macOS
- .NET: 10.0.9
- Chronicle: 16.19.2

## Acceptance criteria

- `ProcessForSingleReadModel` returns no model after a root `Removed` change.
- Other immediate projection entry points do not emit an empty/id-only phantom model for a removed key.
- Specs cover create -> remove for both the single-read and multi-result immediate projection paths.
- Existing sink-backed removal behavior remains unchanged.
- The fix is additive/patch-compatible.

woksin:
Fixed in #3599.

PR #3654, opened 2026-08-10
Add compliance-key migration and fix reliability and privacy defects

woksin:
## Added

- `Compliance:Encryption:MigrateFromDefaultStorage` supports read-through migration from default compliance key storage to Vault or Azure Key Vault. (#3639)

## Fixed

- Rejected seed batches no longer disconnect clients or trigger reconnect loops; they remain unseeded for correction. (#3662)
- `ReviseEvent` resolves `IEventSerializer` in the current event-store scope. (#3663)
- Compliance-key erasure attempts every configured store and reports incomplete erasure. (#3664)
- An unavailable compliance key store is reported as unavailable, not as a missing key. (#3665)
- Refused or unconvertible join writes are logged as warnings without exposing key values. (#3656)

## Security

- Rejected seed diagnostics omit event-source IDs and constraint messages that may contain compliance subjects. (#3662)


woksin:
## Reviewer context — this is a restore, not new work

### What was reverted

`9404b2b42` ("Harden projection compliance and release safety", merged in #3599 as v16.20.0) is a stale-worktree overwrite. Its own subject is about a different item; it reverted the whole of the `feat/imp-26-composite-encryption-key-storage` branch as a side effect. Nothing caught it: the specs were deleted along with the code, so CI stayed green, and `git log -- <file>` hides it because the release merge is TREESAME on those paths.

Ground truth, reproducible:

```
git rev-parse 63662f215:Source/Kernel/Storage/Compliance/CompositeEncryptionKeyStorage.cs
git rev-parse 9404b2b42:Source/Kernel/Storage/Compliance/CompositeEncryptionKeyStorage.cs
git rev-parse origin/main:Source/Kernel/Storage/Compliance/CompositeEncryptionKeyStorage.cs
# all three -> b1bb0f4eb135db3d9fff7ebf854bbe48414c8086
```

`63662f215` is `a3ce03467^1` — the commit *before* IMP-26 was merged. `9404b2b42` put that exact blob back.

### Full audit of the 43 deleted files

`9404b2b42` is 133 files: 45 added, 45 modified, 43 deleted. **All 43 deletions are IMP-26 collateral. None of them belongs to the commit's stated purpose.** They are the 6 `Core.Specs/Setup/for_ComplianceEncryptionKeyStorageExtensions/**` specs, `Core/Setup/{ComplianceEncryptionKeyStorageExtensions,UnresolvableDefaultEncryptionKeyStorage}.cs`, the 26 `Storage.Specs/Compliance/for_CompositeEncryptionKeyStorage/**` specs, the 4 `Storage.Vault.Specs/for_CompositeEncryptionKeyStorage/**` specs, and the 5 `Storage/Compliance/{CompositeEncryptionKeyStorageLogMessages,EncryptionKeyErasureIncomplete,EncryptionKeySaveIncomplete,EncryptionKeyStorageUnavailable,MissingInnerEncryptionKeyStorage}.cs`.

The deletions were not the whole loss. Auditing the 45 *modified* files by testing whether each new blob is an exact resurrection of an older version of that path found **5 more silent reverts, all IMP-26**:

| file | reverted to the version at |
| --- | --- |
| `Source/Kernel/Storage/Compliance/CompositeEncryptionKeyStorage.cs` | `dd683383a` (2026-07-10) — 208 lines dropped |
| `Source/Kernel/Core/Configuration/Encryption.cs` | `213d14fef` (2026-05-16) |
| `Source/Kernel/Storage.Vault/VaultChronicleBuilderExtensions.cs` | `a39f5c4b4` (2026-06-07) |
| `Source/Kernel/Storage.AzureKeyVault/AzureKeyVaultChronicleBuilderExtensions.cs` | `7de9c73c2` (2026-05-20) |
| `Documentation/hosting/configuration/compliance-storage.md` | `2c4df66ca` (2026-07-23) |

So the collateral is **48 files, and IMP-26 is the entirety of it** — 43 deleted plus 5 reverted.

The other 85 files were checked too, and are clean. The 45 added files are all genuinely new (no path among them has any prior history, so none is a re-add masking a delete). For the remaining 40 modified files I blamed every line the commit removed against `9404b2b42^`: the removed lines trace to the IMP-27 analyzer WIP (`6e6d0928a`, `1667be80b`), the CHR0038/CHR0043 analyzers, and the event-serializer scope work (`cb09708c2`, `63662f215`) — i.e. code the same commit is deliberately rewriting, consistent with its stated purpose. The two that touch IMP-24 (`ReadModelReleaser.cs`, `Integration/Client/.../and_the_source_subject_is_erased.cs`) I read in full; both are forward changes (honouring `[JsonPropertyName]`, and making two assertions conditional on `DocumentCanBeInspected` like their siblings), not losses. **Nothing outside IMP-26 needs restoring.**

### What I restored, and what I adapted

Restored from `fb2e5098e`, `7ae8127d0`, `9e3393fbf` and `b1ec4b3f8`, kept as four commits mirroring the originals. The branch delta against `main` is 48 files / +1605 / −44, which is byte-for-byte the exact inverse of `9404b2b42` on those paths and identical to the IMP-26 merge stat.

**Nothing needed adapting.** That is a finding, not an omission, so here is why it holds rather than an assertion that it does. `git diff b1ec4b3f8 origin/main` restricted to `Storage/Compliance`, `Core/Setup`, `Core/Configuration`, `Storage.Vault`, `Storage.AzureKeyVault` and the three spec trees returns *only* the revert itself — IMP-24's compliance work landed in `Clients/DotNET` (`ReadModelReleaser`, `[ReleaseUnder]`), and #3608/#3645 landed in seeding, projections and the MongoDB sinks. None of them touches the key-storage area. I then re-read the seams that could have moved underneath the restore and confirmed each on current `main`:

- `IEncryptionKeyStorage` — unchanged, including the `GetOrAddFor` default implementation the composite relies on for idempotent mirroring.
- `Source/Kernel/Server/Program.cs:208-219` — the backend (`WithInMemory`/`WithSql`/`WithMongoDB`) is still registered before `WithVaultComplianceStorage`, which is what lets the seam's `LastOrDefault` find the descriptor to take over.
- The in-memory backend still registers by `ImplementationType` (convention self-binding) and MongoDB/SQL still register through a factory — the two shapes `Materialize` handles, and the premise of two of the restored setup specs.
- `EncryptionKeyCacheGrainService.Evict` still guards on `if (keyStore is IEvictEncryptionKeyCache)`, and `CacheEncryptionKeyStorage`'s positive `_keys` dictionary still has no TTL (only `_absentKeys` does). The eviction fan-out is therefore still load-bearing exactly as originally argued.

### Verification

Restoring specs whose code was deleted with them is worthless if they pass for a new reason, so every restored spec was re-falsified rather than just re-run.

**Whole-behavior falsification.** I put `origin/main`'s composite algorithm back verbatim behind the restored signature (so the specs still compile) and ran them: **21 of the 66 unit assertions fail**, including every unreachable-store, partial-erasure, revision-per-read and eviction assertion. The 14 registration specs falsify more strongly still — `WithComplianceEncryptionKeyStorage` and `Encryption.MigrateFromDefaultStorage` do not exist on `main`, so that spec tree cannot compile there at all.

**Mutation testing.** 23 single-behavior mutations, each applied in isolation, rebuilt and run. **Every one is killed**, 59 of the 80 unit assertions are killed by at least one, and the four load-bearing behaviors are each killed by a mutation that targets only them:

| behavior | mutation | killed by |
| --- | --- | --- |
| eviction fan-out | `EvictFromCache` body emptied | `when_evicting_from_cache.should_carry_the_eviction_into_every_inner_cache` |
| eviction fan-out | `IEvictEncryptionKeyCache` dropped from the class | `and_migration_is_on.should_stay_reachable_by_the_cluster_wide_crypto_shred_eviction` |
| unreachable ≠ absent | `TryGetFor` returns `null` instead of raising | `when_getting_a_key.and_a_store_is_unreachable_and_no_other_has_it.should_fail_rather_than_report_the_key_as_absent` (+1) |
| unreachable ≠ absent | `HasFor` returns `false` instead of raising | `when_checking_if_a_key_exists.and_a_store_is_unreachable_and_no_other_has_it.should_fail_rather_than_report_that_it_does_not_exist` |
| delete reaches every store | `DeleteFor` stops at the first failure | `when_deleting_a_key.and_the_first_store_is_unreachable.should_still_erase_it_from_every_reachable_store` (+2) |
| delete reaches every store | `DeleteFor` only reaches the primary | `and_every_store_holds_it.should_erase_it_from_the_secondary_store`, `.should_not_heal_it_back_on_a_later_read` (+2) |
| no backwards heal | heal targets every other store (the original bug) | `and_only_the_primary_store_has_it.should_not_write_the_key_backwards_into_the_secondary_store` (+3) |

The other 16 mutations cover store ordering, the empty-composite guard, `GetOrAddFor` primary-failure semantics, mirror idempotency, revision-scoped deletes, and all four registration behaviors — shadowing instead of taking over the descriptor is killed by `should_take_over_the_default_registration_rather_than_shadow_it`, ignoring the setting by both `and_migration_is_off` assertions, and flipping the primary/secondary order by `should_move_that_key_into_the_dedicated_storage`.

The 21 surviving assertions are the negative controls and happy-path guards that sit beside a killed sibling (`should_not_return_a_key` next to `should_fail_rather_than_report_the_key_as_absent`, `should_return_the_key` next to the heal assertions). They pin the converse of an assertion that is already proven and would only be killed by breaking the inner store contract itself. One is worth naming: `when_evicting_from_cache.should_serve_the_stale_key_until_the_eviction_arrives` guards that the eviction spec is not vacuous, and is killed only by giving `CacheEncryptionKeyStorage`'s positive cache a TTL — a change to a file this PR does not touch.

**Real-Vault tier.** All 23 `Storage.Vault.Specs` pass against the Vault container. Because that tier passing quickly proves little on its own, I mutation-proved it too: reversing store order kills `should_heal_it_into_vault` and `should_heal_it_as_the_initial_revision`; healing backwards kills `should_not_write_backwards_into_the_default_storage_once_vault_serves_the_key`; deleting only from the primary kills `should_erase_it_from_the_default_storage` and `should_not_heal_it_back_on_a_later_read`; provisioning with `SaveFor` kills `should_not_mint_a_second_revision_in_vault`. That last one only fails if Vault's real path versioning is in play, so the container is genuinely being exercised.

**Gates.** `dotnet build Chronicle.slnx` in Debug and Release: 0 warnings, 0 errors both. `dotnet test Chronicle.slnx -c Release --settings specs.runsettings --filter "FullyQualifiedName!~Integration"`: 17 suites, 7189 tests, 0 failures. The first attempt hit the known `MSB3231` race on `Source/Kernel/Server/out`; `rm -rf` and retry was green, and no parallel-execution flake reproduced.

### Nothing outstanding

Everything IMP-26 delivered is back, including the four real-Vault specs. There is no follow-up work implied by this PR.

woksin:
Reviewer context, kept out of the body so it stays out of the release notes. This PR now carries what was previously split across #3654 and #3653; #3653 is closed in favour of this one so the repo cuts a single release.

**Half of this is a restore, not new work.** Commit `9404b2b42`, whose message is about a different item, silently reverted the whole IMP-26 branch 40 minutes before #3599 merged. It was a stale-worktree overwrite: 43 files deleted, plus 5 more *modified* files whose new blob was an exact resurrection of an older version — 48 files in total, all IMP-26, nothing else. The 5 resurrections would not show up in a deletion audit; they were found by testing every modified file for blob identity against its own history. CI caught none of it, because the deleted specs went with the deleted code, and `git log -- <file>` hides it because the merge is TREESAME on those paths. The restored delta is byte-for-byte the inverse of that revert on those paths, and the audit confirmed nothing outside IMP-26 was lost.

**The seeding fix.** The grain declines to throw on a rejected batch and says why in a comment; the gRPC service one layer up threw anyway. That reaches the client through `RegisterAll` on `OnConnected`, where a failing handler sets `IsConnected = false`; because a seed rejection is deterministic, the watchdog reconnects and reproduces it — a permanent client outage from a fixable mistake in seed data. Three specs pinned the old behaviour and are inverted here, each with a comment explaining why. The integration spec's other facts are unchanged and still pass: nothing appended from the rejected run, the corrected run lands, and the innocent entry that shared the rejected batch lands with it.

A first attempt at this fix was wrong and is worth recording: catching `EventSeedingIncomplete` client-side does not work, because that type lives in the kernel and the client cannot reference it — across gRPC it is not the typed exception. The service layer is the only place it can be fixed.

**Verification.** Debug and Release both clean, zero warnings. Unit gate 7,254 passed / 0 failed. Integration tier against a real kernel and MongoDB: 155 passed / 0 failed across `for_EventSeeding`, `for_PIIManager`, `for_ReadModels` and `for_EventSequence.when_revising`.

The integration tier was blocked until an hour ago by a kernel startup abort on `main` (#3660) — `IMongoClientFactory` capturing Arc's tenant-scoped `IMongoClient`, surfaced when Arc 21.1.0 restored Development scope validation. Fixed on main by #3650, so this PR is the first to be verified end to end since.

For IMP-26 specifically, the load-bearing behaviours were mutation-proved rather than assumed: eviction fan-out (without it, cluster-wide crypto-shred eviction is a silent no-op and a peer silo serves a shredded key indefinitely, since the positive cache has no TTL), unreachable-is-not-absent, delete-reaches-every-store, and no-backwards-heal. 23 isolated mutations, all killed; the real-Vault tier passes 23/23 against a container.

**Related, tracked separately:** #3655 (CHR0044 warns on a correct PII join and breaks Release builds under `TreatWarningsAsErrors`), #3656 (two diagnostics describing a lost write are logged at Debug), #3657 (four analyzer specs pass on a substring collision), #3658 (v16.20.0 release notes understate the upgrade impact).

Issue #3664, opened 2026-08-10
Composite compliance-key erasure stops after the first failing store


woksin:
## Defect

`CompositeEncryptionKeyStorage.DeleteFor` stops at the first configured store that fails. Stores later in the sequence are never attempted, so a subject key can survive there and a later read can heal the erased copy back into another store. The caller also lacks a bounded result identifying incomplete erasure.

## Expected behavior

Deletion is attempted against every configured store. If any store fails, Chronicle reports incomplete erasure only after all stores have been attempted, preserving the failure details without treating a partial delete as success.

## Shipped fix

Filed retrospectively to give the shipped behavior a durable owning issue. Fixed by [PR #3654](https://github.com/Cratis/Chronicle/pull/3654) and released in [v16.22.0](https://github.com/Cratis/Chronicle/releases/tag/v16.22.0). The release adds composite-store specifications for full deletion, a failing first store, and revision-specific deletion.

woksin:
Fixed in #3654 and released in v16.22.0.

Issue #3600, opened 2026-08-07
[Resolved in Chronicle v16.34.0] Cross-store key erasure could be resurrected by subscription forwarding


woksin:
## Resolution

**Resolved in Chronicle v16.34.0 (#3713; closed 2026-08-15).** Before v16.34.0, subscription forwarding could copy a subject's encryption key into another event store while `DeleteEncryptionKeyFor` removed it only from the selected store and namespace. A surviving copy could subsequently restore the deleted key through forwarding.

v16.34.0 records a namespace-wide, two-phase erasure fence before destroying key material, and requires an explicit `AllowNewEncryptionKeyFor` action for a later lifecycle. Operators must re-run erasure for every subject erased before upgrading, because earlier deletion left no durable erased-state marker.

## Historical context

This report identified the cross-store restoration risk and requested a durable erasure boundary. The resolution above is the authoritative outcome.

## What is explicitly not being asked for

This closed report does not require copied keys for target-store decryption, prescribe a permanent ban on a subject identifier, or reopen the product decision resolved by the shipped fence.

woksin:
Ada has now made the product/security ruling that was blocking this design (D98, 2026-08-10): **erasure removes the key incarnation that exists now; it is not a permanent ban on the stable subject identifier.**

The required safety contract is stricter than a plain delete:

- the erased key must never be returned again by a cache, backing store, replica, migration, retry, failover, forwarding path, or restored process;
- ciphertext written under that key remains permanently unreadable and is never re-keyed as part of erasure;
- a later legitimate lifecycle for the same subject identifier may create a fresh independent key, but it protects only data written after that lifecycle begins;
- a key for subject A must never be returned, copied, healed, or used for subject B, and A's data must never be migrated under B's key;
- divergent live keys for one subject fail closed rather than selecting, overwriting, merging, or falling back to one;
- restoring old key material is an attempted resurrection and must block return-to-service until re-erasure is proven.

So the tombstone/generation mechanism must distinguish **old key incarnation** from **subject identifier forever**. The acceptance bar is adversarial coverage for in-flight cache fills, cross-store forwarding, replica fallback, migration, divergence, delete-then-fresh-create, and proof that the fresh key cannot decrypt the old ciphertext.

This is launch-blocking for Ada because cross-subject access or restoration of erased personal data would be catastrophic. We intend to finish this within the consolidated #3599 minor release train if the implementation remains product-appropriate and compatibility-safe.

woksin:
[...]

Issue #2505, opened 2026-01-18
Ensure Monaco Language service has full support for the Projection Definition Language


einari:
Look through the documentation and make sure the Monaco Projection Editor supports all of the Projection Definition Language.
Also update the

PR #3667, opened 2026-08-10
Stop a read in flight from restoring an erased compliance key

woksin:
# Summary

A read already in flight could put an erased compliance key back into the cache.

## Fixed

- The compliance key cache no longer restores a key that was deleted or evicted while a read of it was in flight. The cache releases its lock to call the backing store and takes it again to record the answer; an erasure landing in that window was overwritten by the older answer, so a shredded key came back and was served again
- An erasure no longer leaves the key cacheable for the length of the durable delete. The cache was cleared before that call, so a read starting afterwards could still find the key in the store and cache it, and cached keys have no time-to-live. Erasing one revision no longer evicts the others
- A partially failed erasure still clears the cache rather than keeping a key that may already be destroyed, and still evicts from every silo. A composite key store attempts every store and then reports the failure, so the key can be gone and the call still throw
- Saving a key no longer loses to a read that began before it. After a rotation the cache could serve the superseded key, so values encrypted under the new revision failed to decrypt until something evicted the entry


woksin:
Reviewer context, kept out of the description because it is not release notes.

**Provenance.** This recovers abandoned local work — uncommitted, never pushed, last touched 2026-08-06, found five days stale in a checkout that was 138 commits behind. It is not superseded: `origin/main` had zero generation-counter markers and no upstream commit had touched `CacheEncryptionKeyStorage.cs` since. The original patch no longer applied, so the design was re-derived against current `main`.

**Verification.** `dotnet build Chronicle.slnx` Debug and Release both 0 warnings / 0 errors. `Storage.Specs` 136 passed / 0 failed; `Core.Specs` 2,398 passed / 0 failed (1 pre-existing skip). Integration and the MongoDB / SQL / Vault / Azure Key Vault suites were not run — no infrastructure was started. That is not a coverage gap here: `Storage.Specs` is the only place in the repository that specs this type, and its other references are DI registrations the full-solution build compiles.

**The abandoned specs covered one of three paths.** They exercise `TryGetFor` only. `GetOrAddFor` and `HasFor` release the lock identically and had no coverage at all — reverting the `GetOrAddFor` guard left both abandoned specs green. A spec per uncovered path was added, and each of the three mutations now kills exactly the spec covering it and nothing else.

**The race specs were checked for vacuity empirically, not just structurally.** With the production guard still reverted, removing the spec's interleaving gate so the fake store completes synchronously made both specs pass 2/2 against broken code; restoring the gate made them fail 2/2. The interleaving is doing the work.

**Set aside deliberately, not overlooked.** The same abandoned tree carried a `DivergentEncryptionKeys` exception plus one spec, with nothing throwing it. Finishing it is not a wiring job: the spec's second fact requires the composite to *remember* a conflict (the primary is unreachable in that scenario, so divergence cannot be re-derived at read time), which means a stateful per-silo divergence registry with no eviction or reconciliation path. It would also invert `and_the_secondary_store_already_has_a_different_one`, which landed in `b1b0e1553` "Restore the CompositeEncryptionKeyStorage hardening" on 2026-08-10 — a current, deliberate decision whose rationale is stated in the class docs (failing provisioning would stop every write of a protected value while a secondary store is unavailable). Turning fail-open into fail-closed on the PII write path belongs with #3600, not with a recovery. Both files are preserved outside the repo; nothing half-wired was left behind.

**Also recovered:** four `for_NamespacesReactor` specs, kept as-is. They compile and pass against current `main` and assert the *upstream* behaviour — two identical seed entries stay two facts, and incomplete seeding fails the observation. The abandoned tree's own reactor deduped seeds and used the non-result-aware seeding interface, so these specs lock in exactly what that approach would have regressed.

**Dropped:** an `EventSequenceMutationExtensions` hunk. Upstream already made that change; the abandoned version's only remaining effect was deleting the explanatory comment.

**No issue reference in the description** because none covers this specific race — #3664 is closed and #3600 is the broader erasure-semantics question. Happy to file one if you would rather it be tracked.

woksin:
Reviewer context for `e547197ca` and `5f89369df`, kept out of the description.

**Where these came from.** Independent code and security review of the original commit. The security review said merge; the code review reproduced a remaining interleaving and it turned out to be the more important finding.

**The gap.** The original guard closed the case where a read is *already* in flight when the erasure begins, and left its mirror image open. `DeleteFor` cleared the cache and advanced the generation *before* the durable erase, so a read that missed the cache after that point captured a generation nothing would move again, reached a store that still held the key, and cached it. Positive entries have no TTL, so that entry would never leave. End to end the erasure still held — but only because `PIIManager` broadcasts an eviction afterwards, which is not a property this decorator can rely on, and which the class's own doc claimed it did not need.

**Two things added beyond the reviewers' asks**, both easy to drop if you disagree:
- The post-erase invalidation runs in a `finally`. `CompositeEncryptionKeyStorage` sits *outside* the caches (each inner store is wrapped — `ComplianceEncryptionKeyStorageExtensions.cs:49`, `MongoDBChronicleBuilderExtensions.cs:79`), and a store that erases revision by revision can throw part way. A failed erase is not "nothing happened".
- Invalidation was extracted to a revision-scoped helper, so erasing one revision no longer evicts unrelated revisions. Three copies of the removal collapsed to one.

**Interleavings.** Enumerated from the state machine — every (mutator, reader) pair × which side reaches the backing store first × which side commits to the cache first — rather than from the shape of the guard. Fifteen cases; four were already covered, four are new, four are degenerate (no store round-trip, nothing to interleave), and the rest fall out of existing specs. Every new spec also asserts what the *racing caller itself* received, which none of the four original specs did.

**The evidence that matters.** With the erasure fix reverted, the entire pre-existing suite is blind:

```
--- WHOLE FOLDER UNDER THE FIX-1 MUTATION: which specs notice? ---
  Failed ...when_erasing_fails_in_the_backing_store.should_not_keep_serving_what_may_already_be_gone
  Failed ...when_erasing_fails_in_the_backing_store.should_read_the_backing_store_again
  Failed ...when_reading_during_an_in_flight_deletion.should_not_leave_the_erased_key_in_the_cache
  Failed ...when_reading_during_an_in_flight_deletion.should_read_the_backing_store_again
Failed!  - Failed: 4, Passed: 30, Skipped: 0, Total: 34
```

All twenty-five original specs pass. They were each a schematic inversion of the guard that had been written, so the suite was complete against the code that existed and blind to the code that did not — which is invisible to mutation testing by construction, since mutation coverage only asks whether written code is exercised.

Four mutations, each applied → built → run → captured → restored in one command under `trap … EXIT`, restoring by writing original bytes back then `touch`ing. Each kills only its own spec and leaves the siblings green: the erasure fix, the `finally`, the `SaveFor` generation advance, and the `PIIManager` `try/finally`. `PIIManager` had **zero** coverage before this.

**Gate.** Debug and Release both 0 warnings, 0 errors across `Storage.Specs` and `Core.Specs` and everything they transitively build. `Storage.Specs` 145 passed (136 + 9). `Core.Specs` 2,394 passed / 1 skipped — the skip is pre-existing, and the net −4 is 8 namespace-seeding facts moved out against 4 `PIIManager` facts added. `Storage.InMemory.Specs` 109, `Storage.Vault.Specs` 23, `Services.Specs` 46, all green.

`Storage.AzureKeyVault.Specs` fails 10 — pre-existing and environmental. They run out-of-process against a live Key Vault and fail with `Azure.RequestFailedException : The SSL connection could not be established` inside `Azure.Core.Pipeline.RetryPolicy`, i.e. the TLS handshake fails before any Chronicle code runs. No Azure Key Vault code path is touched here.

**Deferred deliberately.**
- `_generations` has no reclamation path — insert-only, one entry per erased subject per silo for the process lifetime, and `KeyScope` holds the subject identifier. Recommendation is a single process-wide counter rather than per-scope refcounting: the value is never read, ordered, persisted, or compared across scopes, only compared for equality against one the same call captured seconds earlier. A global counter can only fail safe (an extra backing-store read), removes the map entirely, and retains no subject identifiers — which matters, because Chronicle deliberately keeps the subject out of the kernel log and this map quietly kept it in memory. Refcounting would add increment/decrement to every early return and exception path on the one code path where a mistake un-erases a key, and a dropped decrement would silently disarm the guard with the suite still green. Left out of this PR because changing the fencing granularity deserves its own review.
- `CompositeEncryptionKeyStorage.Heal` (`:73-109`, `:227-242`) can restore an erased key into the **durable** store during a `MigrateFromDefaultStorage` window. Not a cache window — permanent. Separate PR.
- Azure Key Vault never purges: no `PurgeDeletedSecretAsync` anywhere, so a soft-deleted secret stays recoverable for 7–90 days while the Vault backend implementing the same interface destroys permanently. Worse, the AKV specs assert erasure through `HasFor`, which enumerates live secrets only — so the assertion cannot tell "purged" from "soft-deleted", which is how this stayed green. Needs an owner decision, and its fix needs that assertion rewritten or it will land as green as the bug did.
- New, from the interleaving enumeration: a provisioning in flight across an erasure re-mints the key **in the durable store**. `GetOrAddFor` that reached the backing store before `DeleteFor` did will have created a key there; the second invalidation correctly keeps it out of the cache, but the store now holds a fresh key for a subject just erased. The cache decorator cannot fix this — it needs ordering at the store, or an erasure tombstone. Same class of problem as `Heal`.

**Note on adoption.** This ships above the consumer version ceiling. Consumers are held at 16.19.3 by cli #67 (16.20.0 inverted the TLS default) and Arc #2505 (16.21.0 drags Arc 21.1.0), both still open — so upstream shipping is not the same as any consumer being fixed.

Issue #3622, opened 2026-08-10
a projection replay rewrites every stored child whole, defaulting every unset member


woksin:
## Historical defect — fixed

A projection replay could synthesize a CLR default for an unset read-model property during the compliance round-trip. For an enum whose schema excluded `0`, this could persist `0` even though the registered schema did not permit it. The resulting invalid stored value could be silently omitted by schema-aware rendering or, on a bare serialization path, rejected by the .NET enum converter.

## Resolution

Fixed in Chronicle 16.16.0 by `f42241082` (“Stop synthesizing a schema-forbidden default for an unset read model property”).

`GetDefaultValue` now:

1. recognizes nullable properties from the schema `type` array; and
2. does not synthesize a type default that is outside the property's declared `enum` members.

A declared zero member remains valid: if `0` is an allowed enum member, it may still be written as that property's default.

The follow-up validation in #3668 (`fa94756747bafc1a78546c24e58d2c03a8197155`) adds mutation-proven coverage for both guards, their interaction, generated-schema nullability, and the non-nullable declared-zero case.

## Operator action

The fix is forward-only. Documents written before 16.16.0 are not repaired automatically. Replay the affected projection after upgrading to rebuild any documents that already contain an illegal synthesized value.

woksin:
Fixed, and now guarded.

**The defect** was fixed by `f42241082` "Stop synthesizing a schema-forbidden default for an unset read model property", first released in **v16.16.0**. `GetDefaultValue` (`Source/Infrastructure/Schemas/JsonSchemaExtensions.cs:191-216`) gained two guards: `IsNullable()` now also reads the `type` array (`:232`), and `IsDeclaredMember` (`:283-292`) refuses to synthesize a value the property's own `enum` forbids. All four acceptance arms hold, and the stated non-goal is preserved — a type default that **is** a declared member is still written.

**What was missing until now is that none of it was falsifiable.** Verified by mutation before the fix landed in #3668:

- Inverting `IsDeclaredMember` left **the entire suite green**. No fixture anywhere declared an enum whose zero is a real member, so the non-goal was asserted nowhere.
- The nullable-enum spec picked a subject **both** guards suppress, so deleting the type-array nullability leg also left it green — the only thing catching it was a nullable *boolean* assertion in the file named for the non-nullable enum.
[...]

Issue #3619, opened 2026-08-10
the testing harness materializes an empty child collection as [] while the live sink omits it, so no spec at any tier can catch the divergence


woksin:
## Chronicle resolution released in v16.16.0

Chronicle’s reader/harness divergence was fixed by `5f5f058b0` (reader half) and `42cf14304` (harness half). The harness now obtains children-property paths from the kernel and uses the same declared-collection serialization policy as the live client. Regression coverage was added in #3668.

## Scope boundary: Arc follow-up remains separate

This closed Chronicle issue does **not** establish that Arc `CommandScenarioReadModels.SeedEvents` reaches the corrected harness path. As recorded in the closing comment, it still supplies a non-null seed and reaches its result by a different mechanism. Any required Arc behavior change is separate; Arc is not resolved by this Chronicle issue.

## Historical defect

The testing harness materialized an empty child collection differently from the live read-model path. As a result, tests could pass with `[]` while production read an absent collection as `null`, making a non-nullable collection dereference fail only outside the harness.

## Reproduction at filing

1. Define a read model with a non-nullable `[ChildrenFrom]` collection.
2. Create the read model with no child events.
3. Compare the live persisted/read-back result with the harness result.

**Expected:** the harness and live client use the same declared-collection semantics.

**Historical result:** the harness and live path produced divergent representations.

## Resolution

Chronicle now derives collection paths from the kernel schema and uses the same declared-collection serialization policy for the reader and harness paths. The change preserves the distinction between nullable and non-nullable collection declarations.

## Scope

This issue concerns Chronicle reader/harness fidelity. It does not resolve Arc command-scenario seeding behavior or replace a live-kernel measurement.

woksin:
Fixed, and now guarded.

**The divergence** was closed by `42cf14304` "Apply the kernel's children-path exclusion in the in-process read model harness", first released in **v16.16.0**, alongside `5f5f058b0` for the reader half (#3618).

The harness no longer pre-seeds every array-typed property. It sources the projection's children paths from the kernel's own `GetChildrenPropertyPaths` — made public precisely so that anything reproducing the initial-state write has to reproduce the exclusion with it — and materializes through the **same** reader options as the live client (`WithDeclaredCollectionsNeverNull`). Two tiers now run one function rather than two constants, which is the shape this report asked for.

**But the exclusion itself was unfalsifiable**, which is the more interesting half:

[...]

Issue #10, opened 2021-09-27
Figure out how to get MSBuild to do the build of Workbench and embedding in the AspNetCore Client


einari:
Locally on both Mac and Linux, it will build correctly leveraging the `CreateItem` task for dynamically injecting `EmbeddedResource`. Running this on GitHub Workflows it does not do this and the files aren't there and not included in the assembly.

The inspiration for this is from [here](http://www.databaseforum.info/8/593285.aspx).

einari:
Not applicable anymore

PR #3668, opened 2026-08-11
Pin the read model shape guards that no spec could falsify

woksin:
# Summary

Makes the guards behind the read model shape fixes falsifiable. No behavior changes.

### DIFF
--- a/Integration/Client/for_ReadModels/when_getting_instance_with_an_unset_optional/and_the_optional_is_an_enum.cs
+++ b/Integration/Client/for_ReadModels/when_getting_instance_with_an_unset_optional/and_the_optional_is_an_enum.cs
@@ -73,6 +73,16 @@ public class and_the_optional_is_an_enum(context context) : Given<context>(conte
     [Fact] void should_have_read_the_stored_document_when_the_backend_allows_it() => (!Context.DocumentCanBeInspected || Context.StoredDocument is not null).ShouldBeTrue();
     [Fact] void should_not_store_an_answer_the_read_model_never_gave() => (!Context.DocumentCanBeInspected || !Context.StoredDocument!.Contains(nameof(SignedContract.Outcome))).ShouldBeTrue();
     [Fact] void should_store_the_property_that_was_set() => (!Context.DocumentCanBeInspected || Context.StoredDocument!.Contains(nameof(SignedContract.Signer))).ShouldBeTrue();
+
+    /// <remarks>
+    /// This read model reaches the code under test only because it carries a <see cref="PIIAttribute"/>, which is what
+    /// sends its whole state through the registered schema on the way into the sink. That is a precondition, not a
+    /// subject, and every other assertion here is satisfied identically when it stops holding - so without this one a
+    /// regression in compliance detection would leave the spec green while it quietly stopped exercising anything.
+    /// Reading it back cannot tell the difference either, since the released value is the plaintext either way; only
+    /// the stored bytes can.
+    /// </remarks>
+    [Fact] void should_store_the_pii_property_encrypted() => (!Context.DocumentCanBeInspected || Context.StoredDocument![nameof(SignedContract.Signer)].AsString != "Ada Lovelace").ShouldBeTrue();
 }
 
 [PII]
--- a/Integration/Client/for_ReadModels/when_getting_instance_with_an_unset_optional/and_the_optional_is_an_enum.cs
+++ b/Integration/Client/for_ReadModels/when_getting_instance_with_an_unset_optional/and_the_optional_is_an_enum.cs
@@ -1,0 +1,106 @@
+// Copyright (c) Cratis. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#pragma warning disable SA1402
+
+using Cratis.Chronicle.Compliance.GDPR;
+using Cratis.Chronicle.Events;
+using Cratis.Chronicle.Projections.ModelBound;
+using MongoDB.Bson;
+using context = Cratis.Chronicle.Integration.for_ReadModels.when_getting_instance_with_an_unset_optional.and_the_optional_is_an_enum.context;
+
+namespace Cratis.Chronicle.Integration.for_ReadModels.when_getting_instance_with_an_unset_optional;
+
+/// <summary>
+/// The enum case of the unset optional, driven through the compliance write path and asserted on the document at
+/// rest. A read model carrying a <c>[PII]</c> value has its whole state converted through its registered schema
+/// on the way into the sink, and every schema property with no value is offered a type default there. An optional
+/// one must be offered none, so its field is simply never written - and after a projection replay, never stamped
+/// onto every stored record.
+/// </summary>
+/// <remarks>
+/// The enum declares a zero member, and that is the whole reason this spec can fail. The obvious subject - a
+/// 1-based enum, the one the defect was reported against - cannot be observed here at all: the converter encodes
+/// an enum by member name, so a value outside the declared list has no name to write and the field stays absent
+/// whether the default was withheld deliberately or refused incidentally. Both answers look identical at rest, so
+/// a spec built on that subject would be green against a kernel with the fix removed. An enum whose zero <em>is</em>
+/// declared has a name for it, and the difference between withheld and written becomes a difference in the stored
+/// document.
+/// <para>
+/// The read-back assertion is deliberately paired with the at-rest one rather than trusted alone - a round trip
+/// resolves through the same member list and cannot, on its own, separate an absent field from a stored value the
+/// schema forbids.
+/// </para>
+/// </remarks>
+/// <param name="context">The scenario context.</param>
+[Collection(ChronicleCollection.Name)]
+public class and_the_optional_is_an_enum(context context) : Given<context>(context)
+{
+    public class context(ChronicleFixture fixture) : Specification(fixture)
+    {
+        const string CollectionName = "SignedContracts";
+
+        public EventSourceId ContractId { get; } = "unset-optional-enum-contract-1";
+        public SignedContract? Instance { get; private set; }
+        public BsonDocument? StoredDocument { get; private set; }
+
+        public bool DocumentCanBeInspected => StoredReadModelDocument.CanBeInspected(ChronicleFixture);
+
+        public override IEnumerable<Type> EventTypes => [typeof(ContractSigned), typeof(ContractRejected)];
+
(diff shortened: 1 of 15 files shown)


woksin:
Reviewer context. **Spec-only — zero production files changed** (`git diff --stat HEAD` touches only `*.Specs` and `Integration/`).

### A policy question for the maintainer — the label

This PR carries `patch`, but arguably it should carry **no label at all**, exactly like a documentation-only PR. Nothing here ships: `.Specs` projects are not packaged, so merging cuts a release whose contents are byte-identical to the previous one. `pull-requests.md` scopes the no-label carve-out narrowly to documentation, and `verify-semver-label` treats a missing label as a loud error, so extending that carve-out to spec-only PRs is a policy decision rather than mine to take. `patch` is the conservative choice; say the word and it comes off.

### Why this exists

Two Critical defects — #3622 and #3619 — were fixed in 16.16.0 and both remain fixed. The problem is that their guards could not be falsified by any spec in the repo, so a future change could silently restore either one with the whole suite green.

| Guard | Before |
|---|---|
| `JsonSchemaExtensions.cs:291` (member-list check) | inverting it left **the entire suite green** |
| `JsonSchemaExtensions.cs:232` (type-array nullability leg) | rescued only by a nullable **boolean** assertion in the file named for the non-nullable enum — species 15 |
| `ProjectionExtensions.GetChildrenPropertyPaths` | **zero** spec coverage, three production call sites |
| `ChangesetExtensions.cs:78-88` (children exclusion) | the only spec calling `AddPropertiesFrom` omitted the argument that enters the branch |
| `HandleEvent.cs:92` | the pipeline spec asserted only `_addCallCount > 0` |

### Phase 0 — the check that could have invalidated the whole premise

`:232` reads `schemaProperty.Type.HasFlag(JsonObjectType.Null)`, but every shipped spec asserted against a **hand-written** schema. If the real generator emitted `oneOf` or `$ref` instead, the fix would never fire in production and those specs would be green against a shape production never produces. Probed with the real `JsonSchemaGenerator`:

```
reason: {"type":["integer","null"],"enum":[1,2,3],"x-enumNames":[...]}
reason Type flags = Integer, Null   IsNullable() = True   GetDefaultValue = <null>
```

Confirmed — the fix fires. The generator specs now pin that wire so a future STJ change cannot un-fix it silently.

### Verification

Debug **and** Release, both `--no-incremental`: **0 warnings, 0 errors**. Infrastructure.Specs 607, DotNET.Specs 2319, Core.Specs 2425 (1 pre-existing skip), Testing.Specs 351, Integration `for_ReadModels` 115 — all passing.

Orchestrator re-ran the Release build independently (0 warnings) and ran an independent mutation: `JsonSchemaExtensions.cs:291` → `return false` reddened **exactly the two new assertions** (605 passed, 2 failed), where previously it reddened nothing.

### An eighteenth vacuity species, found while writing the integration spec

**The downstream encoder that launders the defect.** The first integration attempt used the actual CHR-34 subject — a 1-based nullable enum, `[PII]`, replay through the compliance write path, assertion on at-rest BSON. It was green, the document was genuinely inspectable, the PII genuinely encrypted, the field genuinely absent — **and it stayed green with the fix's nullability leg deleted.** `ExpandoObjectConverter.cs:528` encodes an enum by member *name*, so a synthesized `0` outside the member list has no name and is dropped anyway. The stored bytes are byte-identical with and without the fix.

The catalogue already says "assert on stored bytes, not read-back". This is the case where **stored bytes are also insufficient**: a fence at the outermost observable boundary can be neutralised by a *later* stage that independently refuses the bad value. The test is not "how far downstream did I assert" but **"is there a subject for which the two branches produce different bytes"**. Switching to an enum declaring `NotSet = 0` gave the defect a name to write, and the spec now dies to `:232` at rest and on read-back.

### Not pinned — stated plainly

- `and_it_is_nullable.should_withhold_a_default_for_it` (1-based nullable enum) survives mutation of *either* guard, because both suppress it. Left as a behavior statement and annotated as such in the XML; the new declared-zero sibling is what fences `:232` alone.
- `should_mark_it_as_nullable` — the `"null"` type entry is emitted by `System.Text.Json`'s exporter, an unmodified module. It has **no Chronicle mutation handle** (species 9). Documented in the file's `<remarks>`; do not count it as coverage.

### Two follow-ups, not addressed here

- A **pre-existing build race**: `Integration/Api` and `Integration/Client` run the same `<RemoveDir>` on `Source/Kernel/Server/out` concurrently, producing `MSB3231` on a first Release build. Passes on retry and deterministically under `-m:1`.
- `integration.runsettings`'s `TestCaseFilter` does **not** match `Cratis.Chronicle.Integration.for_ReadModels`, so these integration specs run under a different CI filter — worth confirming that job exists, or the new integration spec never runs in CI.

woksin:
**Correction to my earlier comment, and the security review result.**

### The CI concern I raised was wrong — the integration spec does run

I flagged that `integration.runsettings`'s `TestCaseFilter` does not match `Cratis.Chronicle.Integration.for_ReadModels`. That filter is irrelevant: **`integration.runsettings` is referenced nowhere** — not by any workflow, `.csproj`, `.props`, `.targets`, `.sh`, `.slnx` or `.json`. It is dead config, and reasoning from it produces wrong conclusions. It should be deleted or annotated as unused, separately.

What actually selects the Client integration job is `.github/workflows/integration.yml:163` with a matrix from `.github/scripts/generate-integration-matrix.py`. Ran it:

```
{'namespace': 'Cratis.Chronicle.Integration.for_ReadModels', 'filter': 'FullyQualifiedName~Cratis.Chronicle.Integration.for_ReadModels', 'mode': 'inprocess',    'database': 'mongodb'}
{'namespace': 'Cratis.Chronicle.Integration.for_ReadModels', 'filter': 'FullyQualifiedName~Cratis.Chronicle.Integration.for_ReadModels', 'mode': 'outofprocess', 'database': 'mongodb'}
```

Two rows, both MongoDB. And because `DEFAULT_DATABASES = {"mongodb"}`, a pull request runs **only** MongoDB configs — where `CanBeInspected` is `true`. So the at-rest assertions are live on every PR and are never the silently-skipped leg; the SQL no-op legs appear only on the nightly `--all-providers` run, where they are fenced. That is a stronger position than my original comment implied.

### Security review: APPROVE

Verified spec-only by `--numstat`: 15 files, **603 insertions, 0 deletions**, every path under `*.Specs/` or `Integration/`, `JsonSchemaExtensions` untouched. The shared fixture's added `feedback`/`decision` properties are disjoint from every pre-existing assertion in that folder, and the compliance-adjacent neighbours inherit from different fixtures entirely, so none can pass for a new reason.

### One follow-up finding worth acting on

`and_the_optional_is_an_enum.cs:71` asserts the released `[PII]` value through a blind read-back — it passes identically whether the value was encrypted-then-released or never encrypted at all. `[PII]` is not decoration in this spec: it is *the route to the code under test*, because it converts the whole state through the registered schema on the way into the sink. If PII classification silently stops engaging, this spec keeps passing while no longer exercising the path it was built for, and its own remarks become false.

It has the stored `BsonDocument` in hand at `:75` already, so the fix is one line asserting `Signer` is ciphertext at rest — exactly what the sibling `when_a_read_model_has_a_pii_value_object/and_reading_the_instance.cs:57` does on the same helper.

### A nineteenth vacuity species, from that finding

**The unpinned enabling condition.** The spec depends on a *precondition* to reach the code under test, and nothing asserts the precondition. The environment is capable, the input arrives, the assertion sits at the right layer on the right artifact — but the **route** is unpinned. When the precondition stops holding, the spec does not fail; it silently degrades from "proves X" to "proves nothing" and stays green forever.

Distinct from #5 (the input never arrives), #10 (the environment is incapable) and #12 (the spec never executes): here the spec runs, in a capable environment, on real data, and measures nothing.

**The tell:** delete the attribute, config or registration that makes the spec meaningful and ask whether any fact goes red. If none does, the premise is unpinned.

woksin:
**Both reviews in: APPROVE / APPROVE. One finding fixed in `f95c0ceda`, and a correction to my own accounting.**

### The finding both reviewers found independently — fixed

The integration spec reached the code it fences **only** because its read model carries a `[PII]` value, which is what sends the whole state through the registered schema on the way into the sink. Nothing asserted that precondition. The code reviewer proved the consequence by running it: mutating `HasComplianceMetadata()` to `false` left **all nine assertions green** — the released value reads back as plaintext either way, and the unset property was absent because the round trip never happened at all, not because the guard under test held.

`f95c0ceda` adds one assertion on the stored document, which was already being read. Verified by me:

```
arming condition disabled →
  Failed …and_the_optional_is_an_enum.should_store_the_pii_property_encrypted
  Failed! - Failed: 1, Passed: 6, Total: 7
restored + kernel rebuilt →
  Passed! - Failed: 0, Passed: 7, Total: 7
```

Exactly one assertion reddens, and it is the new one.

### Correction to my earlier comment — I under-sold a working fence

I wrote that `should_mark_it_as_nullable` "has **no** Chronicle mutation handle (species 9) … do not count it as coverage", and the spec file says the same. **That is wrong.** The assertion reads through Chronicle's own `JsonSchemaExtensions.IsNullable()`, and mutating `:232` reddens it. The System.Text.Json dependency is the `"null"` **emission**, not the **reading**. The other admission — `should_withhold_a_default_for_it` surviving mutation of either guard — is correct and stands.

### Two operational notes for anyone repeating this

1. **A green mutation result must be re-verified with a forced build; a red one cannot be stale.** The code reviewer's first run of one mutation reported all-green purely as a stale-build artifact.
2. **For integration specs, restoring a kernel source file is not enough.** The fixture bakes the kernel into its container image, so a mutated kernel survives the restore until the Server project is rebuilt. I hit exactly this: after restoring, all 7 failed until `dotnet build Source/Kernel/Server` republished the image.

### Review outcome

Ten independent single-line mutations against the eight claimed production targets, applied one at a time — **no claim broken**. Zero warnings in Debug *and* Release with `--no-incremental`, independently verified. The `integration.runsettings` concern I raised was a false alarm and is corrected above; the spec runs.

### Follow-ups, not addressed here

- `ChangesetExtensions.cs:82-85` (the `IsRoot` continue) is now unreachable from production since `ProjectionExtensions.cs:33` filters roots — defence in depth, but a reader should not assume it is fenced.
- The pre-existing species-4 blindness in `for_EncryptChangeset/given/all_dependencies.cs` (`IExpandoObjectConverter` substituted outright) still means no `JsonSchemaExtensions` mutation can redden anything in that folder. The new integration spec now covers that path end to end, which mitigates it.
- `integration.runsettings` is dead config and should be deleted or annotated.

PR #3669, opened 2026-08-11
Give the PII precondition assertion a summary

woksin:
# Summary

Restores the zero-warning build. No behavior changes.


woksin:
Self-inflicted, from #3668. I added the `<remarks>` explaining why that assertion exists and re-ran only the integration spec afterwards, never a full solution build — so RCS1139 (`<remarks>` without `<summary>`) reached `main`.

Verified fixed: `dotnet build Chronicle.slnx -c Debug --no-incremental -m:1` → **Build succeeded. 0 Warning(s), 0 Error(s)**.

`-m:1` is deliberate and worth recording separately: with default parallelism the full build fails intermittently because `Integration/Client/Client.csproj:45` and `Integration/Api/Api.csproj:32` both `RemoveDir` and then publish into the same shared `Source/Kernel/Server/out` in their `BuildDocker` targets, and MSBuild schedules them concurrently — `MSB3231`, or a `docker … "/Source/Kernel/Server/out": not found`. Serialized it is clean. That is a pre-existing repo build race, unrelated to this change, and deserves its own issue.

Issue #3626, opened 2026-08-10
two [SetFromContext<T>] for the same event type on one property silently discard all but the last


woksin:
## Resolution

Fixed by [#3670](https://github.com/Cratis/Chronicle/pull/3670) (`8a53d6fc429bb7d9080dcd7fb5c26ea4a7f30b2b`) and released in Chronicle 16.16.0. This issue was closed as completed on 2026-08-11.

The report below is retained as historical technical evidence.

## Historical defect

When a projection configured more than one `[SetFromContext<TEvent>]` mapping for the same target property and event type, only the last mapping was retained. Earlier mappings were silently discarded during model-bound projection construction.

## Reproduction at filing

1. Declare two `[SetFromContext<TEvent>]` mappings for one target property.
2. Append `TEvent` with the context values required by both mappings.
3. Inspect the projected property.

**Expected:** the declared mappings compose according to the projection contract.

**Historical result:** only the last mapping survived; earlier mappings were silently dropped.

## Resolution

The model-bound projection builder now retains the declared context mappings instead of overwriting them. The fix is covered by the corresponding analyzer/builder specifications.

## Scope

This issue concerns repeated context mappings for one event type and target property. It does not redefine context-mapping semantics or projection policy.

woksin:
Fixed in #3670 (`8a53d6fc429bb7d9080dcd7fb5c26ea4a7f30b2b`).

The analyzer shipped in 16.16.0 caught two `[SetFromContext<T>]` for one event type only when both attributes sat on the **same symbol**. On a positional record they need not: one can target the primary-constructor parameter and the other the generated property, and those are two distinct symbols. Grouping them apart left the analyzer silent while **both mappings still reached the builder under the same key**, where the property pass runs second and quietly wins — the exact discard the diagnostic exists to report, reachable through a spelling it could not see.

A positional record's generated property is now paired against its parameter and the pair is judged as the single member a reader sees. Declared properties with no matching parameter keep their own path.

Confirmed RED before the fix, and the snippet compiled — so it was genuine analyzer silence, not a broken fixture:
```
and_the_attributes_are_split_between_parameter_and_property.should_report_the_discarded_mapping [FAIL]
   Assert.Equal() Failure: Expected: 1  Actual: 0
```

The state machine is now covered: {parameter-only, property-only, split, both-on-one-symbol} × {same event type, distinct event types}, plus the declared-property cell the pairing skip could have dropped.

Two follow-ups worth recording, neither a regression:
[...]

Issue #3642, opened 2026-08-10
stream metadata attributes claim an EventType placement that does nothing


woksin:
## Describe the bug

This issue promotes a previously recorded upstream item from a production Cratis consumer. The complete evidence and requested fix follow.

**Status:** Resolved upstream 2026-08-11 · fixed by [#3670](https://github.com/Cratis/Chronicle/pull/3670). The documentation now describes the supported placements and `CHR0045` diagnoses inert placement on a plain event type.
**Product:** Chronicle .NET client — `[EventSourceType]`, `[EventStreamType]`
**Type:** Documentation / silent inert placement

## Symptom

Both attributes' XML summaries say they apply to an event type, command or observer. On an `[EventType]` record,
the placement is inert: append stream metadata comes from `Append`/`AppendMany` parameters, while the event CLR
type contributes static `[Tag]` values only. Attribute consumers use them for reactor/reducer registration, and
Arc uses them on commands; nothing reads them from an appended event type.

## Requested fix

Either implement the documented event-type behavior with explicit precedence against append parameters, or narrow
the XML contract and emit a diagnostic for the inert placement. This belongs with CHR-39 and IMP-12 as the third
instance of an attribute whose documented placement compiles and does nothing.

## Honest limits

Source-verified at Chronicle checkout `623db1b95`. Ada has zero event records using the inert placement and pays no
present code cost; it is filed because the public contract is still substantively false.

woksin:
Fixed in #3670 (`8a53d6fc429bb7d9080dcd7fb5c26ea4a7f30b2b`).

Both attributes documented themselves as applying to "an event type, a command, or an observer". Only the latter two are read: the sole consumers take the metadata off a reactor, a reducer, a command type or an aggregate root, and an append resolves stream metadata from its own parameters. Declaring either on an **event type** compiles, reads as deliberate, and has no effect anywhere.

The documentation is corrected and `CHR0045` now reports the placement, so the mistake is caught where it is written rather than discovered by an observer that never sees the events it was meant to narrow.

**Honoring the placement was deliberately refused.** Both values are persisted in every appended event's context and are load-bearing keys for concurrency scopes and observer filtering, so letting an event type contribute them would retroactively change stream identity for every event type already declared, against data already written.

### One correction the review caught, worth recording

[...]

PR #3670, opened 2026-08-11
Diagnose inert stream metadata and a split duplicate context mapping

woksin:
## Added

- A diagnostic reports `[EventSourceType]` and `[EventStreamType]` declared on an event type, where they have no effect (#3642)

## Fixed

- The duplicate context-mapping diagnostic now reports two `[SetFromContext<T>]` for one event type split across a positional record's parameter and its generated property (#3626)
- `[EventSourceType]` and `[EventStreamType]` no longer document a placement on an event type that nothing reads (#3642)


woksin:
Reviewer context. Resolves #3626 (the residual) and #3642. Commits `b6e883fd8`, `dc1f542a6`.

Both are analyzer work and share a folder, so they ride together; the higher of the two labels (`minor`, for the new diagnostic) applies.

### #3626 — the residual, and it was RED before the fix

The diagnostic shipped in 16.16.0 but only caught two `[SetFromContext<T>]` on the **same symbol**. On a positional record they need not be: one can target the primary-constructor parameter and the other the generated property. Grouping them apart left the analyzer silent while both mappings still reached the builder under the same key, where the property pass runs second and quietly wins — the exact discard the diagnostic exists to report, reachable through a spelling it could not see.

Confirmed as a genuine silence, not a broken given, before writing the fix:

```
and_the_attributes_are_split_between_parameter_and_property.should_report_the_discarded_mapping [FAIL]
   Assert.Equal() Failure: Values differ   Expected: 1   Actual: 0
```

The snippet compiled — no `SpecSourceDoesNotCompile` — so the analyzer really did stay quiet.

### #3642 — refused, not honoured

Inertness was re-verified independently, and the check found **one reader the earlier investigation missed**: besides `Reactors.cs:319-320` and `Reducers.cs:437-438`, `EventSourceTypeValuesProvider`/`EventStreamTypeValuesProvider` under `Reactors/SideEffects/` also read it — off `reactor.GetType()`. All observer types. `EventSequence.Append` resolves stream metadata from its own parameters. No call site passes an event type; the extra readers do not change the conclusion.

**Honouring was deliberately not implemented.** Both values are persisted in every appended event's context and key concurrency scopes and observer filtering, so letting an event type contribute them would retroactively change stream identity for every event type already declared, against data already written.

Scoped to the two attributes that **made** the claim. `[FilterEventsByTag]` is equally inert on an event type but never documented that placement, so reporting it would be a new build break with nothing behind it.

### Registration proof — vacuity species 12

The spec harness instantiates analyzers directly, which proves nothing about a real build. Both new behaviours were driven through `Samples/SimpleConsole` — the one in-repo project referencing the analyzer as `OutputItemType="Analyzer"` — against the real attributes, then the probes removed:

```
warning CHR0045: 'EventStreamTypeAttribute' on event type 'Chr0045Probe' has no effect …
warning CHR0045: 'EventSourceTypeAttribute' on event type 'Chr0045Probe' has no effect …
warning CHR0040: 'Stamp' carries more than one [SetFromContext<Chr0040ProbeOpened>] …   (exactly one — no double report)
```

### Verification

Debug **and** Release, both `--no-incremental`: `Build succeeded`, **0 errors**. `DotNET.CodeAnalysis.Specs` **346**, `DotNET.Specs` **2319**, passing in both configurations.

Six mutations, each guard broken independently and restored byte-identically, with GREEN re-verified under a forced rebuild. All seven new specs are individually reddenable — none is decoration. The state machine is fully enumerated: {parameter-only, property-only, split, both-on-one-symbol} × {same event type, distinct event types}, plus the declared-property cell the pairing skip could have dropped, plus command and observer negatives for the placements the narrowed docstring keeps.

### A pre-existing build race worth its own issue

`-m:1` is load-bearing. With default parallelism the full solution build fails intermittently: `Integration/Client/Client.csproj:45` and `Integration/Api/Api.csproj:32` **both `RemoveDir` and then publish into the same shared `Source/Kernel/Server/out`** in their `BuildDocker` targets, and MSBuild schedules them concurrently — `MSB3231`, or `docker … "/Source/Kernel/Server/out": not found`. Serialized, both configurations are clean. Nothing here touches an MSBuild target or an Integration project.

### Not pinned — stated plainly

- The **kernel** `[EventType]` spelling (`Cratis.Chronicle.Concepts.Events.EventTypeAttribute`) is accepted by `HasEventTypeAttribute`, but the given stubs only the client one, so that branch of CHR0045's enabling condition is unpinned.
- Parameter↔property pairing uses `StringComparison.Ordinal` (the C# record contract), while the runtime `ModelBoundProjectionBuilder.ProcessProperties` uses `OrdinalIgnoreCase`. A hand-declared property differing from a parameter only in case would be one member to the builder and two to the analyzer. No spec covers it; Ordinal is judged correct and the divergence was not chased.
- The `InstanceConstructors` tie-break for a single-parameter positional record relies on stable ordering. Pre-existing, unchanged, unpinned.
- Docs are not rendered — `Documentation/web` is not in this repo. `CHR0045.mdx` and its snippet follow the shipped `CHR0041` pattern exactly but are unverified against a real render.

Issue #3607, opened 2026-08-09
Make production hosting guidance executable and encryption-certificate lifecycle recoverable


woksin:
## Problem

Chronicle's production-hosting documentation and helpers currently describe several incompatible or incomplete production shapes.

At exact release `v16.19.2`:

- The production Docker Compose examples start a standalone `mongo:7`, while Chronicle's transactions, change streams, projections and observable queries require a MongoDB replica set.
- The examples do not form a complete secure certificate configuration. One omits TLS material; the multi-node example supplies a TLS certificate but omits the encryption certificate required by the default internal OAuth path.
- The Aspire configure callback is described as suitable for production/staging and its XML documentation says it uses the standard production image, but the implementation selects `latest-development-slim`.
- The encryption-certificate guide permits reusing the TLS certificate and correctly says that certificate must be backed up and rotated, but the runtime accepts only one encryption certificate. Data Protection/OpenIddict/webhook decryption uses the current certificate, so automatic TLS-leaf replacement has no documented or supported previous-key overlap.
- The hosting corpus does not define a backup/restore ordering for Mongo data, the database-backed Data Protection key ring, the encryption certificate, and optional compliance-key storage.

These are generic hosting-contract gaps. They are separate from the already filed client TLS-default issue #3606 and health-only listener issue #3604.

History was checked before filing:

- the Mongo replica-set requirement was deliberately documented and the Aspire provisioning helper shipped in #3400;
- multi-node clustering is intentional;
- the dedicated health listener intentionally shares the application pipeline;
- the Data Protection production startup check was deliberately removed while the default internal OAuth path retained encryption-certificate enforcement.

## Requested compatible contract

Provide one executable, version-pinned production reference topology and a recoverable certificate lifecycle:

1. Correct the production examples to use a replica set and complete TLS/encryption-certificate configuration.
2. Make the Aspire API/XML/docs agree on which image class is selected. A production/staging callback must not silently select a floating development image.
3. Add an additive active/previous encryption-certificate key ring, or another documented rollover mechanism, so a new certificate can decrypt data protected by the preceding certificate during a bounded migration window.
4. Document certificate backup, restore and rotation ordering together with the database-backed Data Protection key ring and optional compliance-key store.
5. State precisely when the encryption certificate is mandatory: the default internal OAuth path requires it; configurations using an external authority or disabling that path have different startup behavior.
6. Keep development helpers convenient and explicit; do not weaken production defaults to preserve a development shortcut.

[...]

PR #3677, opened 2026-08-11
Correct the hosting recipes that cannot start Chronicle

woksin:
## Added

- Environment variable reference entries for the encryption certificate, clustering and health settings

## Fixed

- The production hosting recipes now run MongoDB as a replica set, without which Chronicle cannot open the transactions it uses for event sequence storage (#3607)
- The production hosting recipes now configure the encryption certificate, without which a release build refuses to start (#3607)
- Documented that what selects the production path is the `DEVELOPMENT` compile-time symbol rather than `ASPNETCORE_ENVIRONMENT`, so the published image always takes the production branch (#3607)
- The production `HEALTHCHECK` no longer invokes `curl` inside a chiseled image that has no shell; the health endpoint and an external probe are documented instead
- `Cratis__Chronicle__Authentication__DefaultAdminPassword` is no longer documented as a production setting — it exists only in development builds; the `AdminUser` settings are documented in its place
- The job throttling example is no longer nested one level too deep to bind to anything
- The encryption certificate page no longer attributes the lazy value-encryption failure and the development certificate to Data Protection
- The Docker Compose example no longer points the dashboard's own OTLP endpoint at another host

### DIFF
--- a/Documentation/hosting/configuration/configuration-file.md
+++ b/Documentation/hosting/configuration/configuration-file.md
@@ -34,7 +34,20 @@ Chronicle Server loads configuration from a `chronicle.json` file in the applica
     "authentication": {
         "authority": null,
         "defaultAdminUsername": "admin",
-        "defaultAdminPassword": "admin"
+        "adminUser": {
+            "username": "admin",
+            "password": "your-secure-password",
+            "email": "admin@example.com",
+            "requirePasswordChangeOnFirstLogin": true
+        }
+    },
+    "tls": {
+        "certificatePath": "/certs/chronicle.pfx",
+        "certificatePassword": "your-password"
+    },
+    "encryptionCertificate": {
+        "certificatePath": "/certs/encryption-cert.pfx",
+        "certificatePassword": "your-password"
     },
     "identityProvider": {
         "certificate": {
@@ -46,7 +59,13 @@ Chronicle Server loads configuration from a `chronicle.json` file in the applica
 }
 ```
 
+Note that the sections are **not** wrapped in `Cratis` / `Chronicle` here — Chronicle republishes everything it reads from `chronicle.json` under the `Cratis:Chronicle:` configuration path itself. That prefix belongs on environment variables only.
+
 Environment variables can override any of these values. See [Configuration Precedence](configuration-precedence.md) for details.
+
+> [!IMPORTANT]
+> Keep passwords out of the file itself. Because every value above can be overridden by an environment variable,
+> supply secrets that way from your secret store — see [Environment Variables](environment-variables.md).
 
 | Section | Description |
 | --- | --- |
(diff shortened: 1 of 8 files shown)


woksin:

**Scope.** Documentation-only — all 8 changed files are under `Documentation/hosting/`. No label, and the missing-label `verify` failure is the expected outcome rather than a problem.

**Where the replica-set recipe came from.** Copied from `Documentation/get-started/choose-hosting-model.mdx:144-171`, which already worked, rather than authored fresh — so the two pages now agree instead of offering a third variant. The code-side equivalent is `Source/Clients/Aspire/ChronicleMongoDBDistributedApplicationBuilderExtensions.cs:76-92`.

**Every environment variable written was checked against the options class, not against the old docs:**

| Key | Proof |
|---|---|
| `EncryptionCertificate__CertificatePath` / `__CertificatePassword` | `Source/Kernel/Core/Configuration/EncryptionCertificate.cs:14,19` |
| `Tls__CertificatePath` / `__CertificatePassword` | `Tls.cs:20,25` |
| `Authentication__AdminUser__{Username,Password,Email,RequirePasswordChangeOnFirstLogin}` | `AdminUserBootstrapConfig.cs:17,25,30,38` |
| `Clustering__{Type,ClusterId,ServiceId,SiloPort,GatewayPort,AdvertisedIP}` | `Clustering.cs:21,38,43,27,33,50` |
| `Health__Port` / `Health__Tls` | `Health.cs:28,39` |

This matters because the requirement record this work came from had the encryption certificate keys as `Path`/`Password`. They are `CertificatePath`/`CertificatePassword`.

**Why the replica set is required, from code rather than assumption:** `Source/Kernel/Storage.MongoDB/EventSequences/EventSequenceStorage.cs:217-221` calls `StartSessionAsync()` then `session.StartTransaction()`. A standalone `mongod` cannot serve that.

**Why a release build throws:** `OpenIddictServiceCollectionExtensions.cs:113-119`, reached because the `DEVELOPMENT` symbol is defined only for Debug (`Directory.Build.targets:10-12`) — so the gate is a compile-time symbol, not `ASPNETCORE_ENVIRONMENT`.

**Verification.** `markdownlint-cli2 "Documentation/**/*.md"` — 838 files, 0 issues (baseline was also 0). Each edited fence was re-read for internal consistency, e.g. every service referenced by a Compose file exists in it.

**Deliberately not changed.**
- `Documentation/hosting/aspire.mdx:26` and `:113` describe a real code defect — `ChronicleDistributedApplicationBuilderExtensions.cs:53` selects `DevelopmentSlimTag` on the production branch, contradicting its own XML doc. Those doc lines must change together with the code, so both are held for a separate `patch` PR. Worth knowing: `ChronicleContainerImageTags.Tag` is referenced nowhere in the repository — a dead public constant, which corroborates the defect.
- `encryption-certificate.md`'s H1 is now a partial misnomer, since the certificate serves three subsystems. Renaming it changes the derived page title and inbound link text from three other pages, so it is flagged rather than done.
- `"These keys are stored in MongoDB"` (`encryption-certificate.md:7`) is probably storage-agnostic via `GrainBasedXmlRepository`, but that was not cheap to verify, so the claim was left alone.

**Follow-up outside this repository.** `docker-compose.md` embeds a snippet range from `cratis/samples`; that file almost certainly still carries the bad `DOTNET_DASHBOARD_OTLP_ENDPOINT_URL` line. Not fixable from here.

**Not verified.** No recipe was executed — no container was started from these files. The corrections are proved against source, not against a running deployment.

Issue #3673, opened 2026-08-11
Compliance walk throws CompliancePropertyNotFoundInSchema on geospatial Point leaf schemas (read model with [PII] + Point)


Petteroe:
## Summary

`JsonComplianceManager.HandleActionFor` throws `CompliancePropertyNotFoundInSchema` when a read model has **any** `[PII]`-annotated property **and** a geospatial `Point` (or `LineString`/`Polygon`) property. The projection stalls on the first event that populates the geospatial field, and the read model never catches up.

## Environment

- Chronicle server: `16.13.3` (reproduced) — also confirmed the bug is present on `main` (fetched `JsonComplianceManager.cs` from `main` today, 2026-08-11).
- Client: Cratis Arc `20.69.1` (.NET 10).

## Reproduction

Read model:

```csharp
[ReadModel]
[FromEvent<DraftStarted>]
public record Vibe(
    [Key] string Id,
    // [PII] resolved via the DisplayName concept TYPE (not the property).
    // HasComplianceMetadata() returns true for the whole read model because of this.
    DisplayName? OrganizerDisplayName,
    Location? Location);   // Location { Point LocationPoint, ... }

public sealed record Location(Point LocationPoint, string DisplayName, string City);

// VibeLocationSet event carries the Location; Point serializes as GeoJSON:
// { "locationPoint": { "type": "Point", "coordinates": [lon, lat] } }
```

Append a `VibeLocationSet` event. The `Vibe` projection fails:

```
Could not apply compliance metadata for property 'location.locationPoint.type'
of 'v-100-000-000' because the schema does not declare it. The schema declares: no properties.
```

Stack trace (top frames):

```
Cratis.Chronicle.Compliance.JsonComplianceManager.HandleActionFor(...)
Cratis.Chronicle.Compliance.JsonComplianceManager.HandleActionFor(...)   // recursion into locationPoint
Cratis.Chronicle.ReadModels.ReadModelsCompliance.Apply(...)
Cratis.Chronicle.Projections.Engine.Pipelines.Steps.EncryptChangeset.Perform(...)
Cratis.Chronicle.Projections.ProjectionObserverSubscriber.OnNext(...)
```

The partition enters a failed-partition loop (11 attempts, exponential backoff) and the projection never advances past the `VibeLocationSet` event.

## Root cause

[...]

PR #3678, opened 2026-08-11
Fix compliance handling failing on geospatial and other opaque values

woksin:
# Summary

A `[PII]` value sitting alongside a geospatial value made compliance handling fail — the append was rejected and the projection partition stalled, so the read model never caught up.

## Fixed

- Applying or releasing compliance no longer fails with `CompliancePropertyNotFoundInSchema` when a read model or event carries a `[PII]` value alongside a geospatial `Point`, `LineString` or `Polygon`; the geospatial value is left in the clear (#3673)

## Changed

- Geospatial documentation now describes the JSON schema Chronicle emits — a leaf carrying only its format, with GeoJSON on the wire — and records that marking a geospatial value `[PII]` is not supported, since an erased subject cannot be materialized back into one (#3673)


woksin:
Reviewer context.

**Root cause.** `JsonComplianceManager.HandleActionFor` walks every property in a document and looks it up in the schema, which presumes the schema declares the members of what it describes. Three shapes deliberately do not: a geospatial leaf (`JsonSchemaGenerator.TransformNode` strips `properties`/`required` and leaves only `format`), a polymorphic base type (kept open so the discriminator round-trips), and a dictionary. Descending into any of them made every member look like schema drift.

**Design.** The distinction is a schema question, so it lives with the schema: `JsonSchemaExtensions.DescribesOpaqueValue()`. It deliberately excludes a schema that declares nothing because its declaration could not be *reached* — an unresolved `$ref`, a composition that flattened to nothing. That shape is broken rather than opaque, and calling it opaque would let compliance silently skip a value the schema does mean to describe, so it keeps failing. The guard sits at the top of `HandleActionFor`, which covers the property descent, the array-element descent and the root in one place. Nothing under an opaque value can carry compliance metadata (there is no schema property for a marker to sit on), so skipping is lossless — the same reason `HasComplianceMetadata` never finds anything there either.

**Blast radius.** Every compliance path funnels through `IJsonComplianceManager` — event append (`EventSequence`), event release (`EventCompliance`), constraint reindex, read-model apply/release, and the client-side release over gRPC — so the one guard covers all of them.

**Verification.**
- `for_JsonComplianceManager` unit specs reproduce the reported failure verbatim: with the guard disabled they fail with `Could not apply compliance metadata for property 'location.locationPoint.type' … The schema declares: no properties.`
- `when_a_read_model_has_pii_beside_a_geospatial_value` — all 9 facts fail without the fix (the partition never advances) and pass with it, against a real MongoDB via Testcontainers.
- `when_appending_event_with_pii` — 11/11 green (2 new).
- `Core.Specs` `for_JsonComplianceManager` 58/58, `Infrastructure.Specs` `for_JsonSchemaExtensions` 16/16.
- Release builds of `Kernel/Core.Specs` and `Infrastructure.Specs`: 0 errors, 0 warnings.

**Not changed.** PII *inside* a dictionary is still not classified — `HasComplianceMetadata` does not descend into `additionalProperties`, so that gap predates this change and is untouched by it. It is documented under "Values Chronicle does not look inside" rather than left implicit.

woksin:
Update after review — the approach changed and the scope narrowed.

**What the review caught.** The first shape of this fix asked "does the schema declare no properties?" and asked it on `ActualTypeSchema`, while the walk enumerates `schema`. Those disagree: a value object composed through `allOf: [{$ref: Base}]` beside its own `properties` resolves to the base alone, so a member the type declares — and marks `[PII]` — was classified as "nothing to classify" and written **in the clear** beside a correctly encrypted sibling. Dictionaries and polymorphic payloads lost their append rejection the same way: a loud `CompliancePropertyNotFoundInSchema` became silently storing PII as plaintext, which is precisely what `HandleActionFor`'s own comment forbids.

**What it is now.** A positive question — `DescribesGeospatialValue`, answered by the complex `format`, the same signal `ExpandoObjectConverter` and `ProjectedColumns` already use to decide a value is typed rather than decomposable (so it is no longer a competing definition of "declares nothing"). The guard moved from the top of `HandleActionFor`, where it swallowed every shape, to the point of descent. A geospatial value marked `[PII]` is still handled as a whole; dictionaries, polymorphic types, unresolvable `$ref`s and genuine drift all fail loudly again.

**Scope.** Only geospatial. The earlier claim that this also fixed polymorphic types and dictionaries is withdrawn — those still fail, deliberately, and the docs now say so instead of implying nothing causes that diagnostic.

**Docs corrections the review forced.** The page claimed the value stays `2dsphere`-indexable; the same page documents the sink writing `{longitude, latitude}`, which no geospatial index accepts. And it offered marking the property `[PII]` as a trade — verified in `ExpandoObjectConverter` that an erased subject releases an empty object the GeoJSON converter cannot deserialize, and that branch has none of the compliance-aware catch the scalar branch has, so the read fails for the whole read model. Both corrected.

**Verification.** `for_JsonComplianceManager` 64/64 (was 58 — new specs pin the `allOf` leak, the dictionary failure as contract, and a marked geospatial value still being encrypted), `for_JsonSchemaExtensions` 22/22, integration 20/20 against real MongoDB, Release builds 0 errors 0 warnings.

**Known, not addressed here** — each needs its own issue rather than riding this one:
- `[PII]` on a geospatial value cannot be released after erasure (documented as unsupported; the code path is untouched).
- An `object`/`dynamic`/`JsonElement` member is exported as the boolean schema `true` and dropped by `JsonSchema.Properties`, so it never becomes a schema property and the same crash still reproduces for it. That is a schema-generation defect, not a walk defect.
- Chronicle does not classify `[PII]` inside a dictionary or a polymorphic payload at all.

Issue #3676, opened 2026-08-11
Integration/Client and Integration/Api race on Source/Kernel/Server/out, causing intermittent MSB3231


woksin:
## Summary

Two projects in `Chronicle.slnx` — `Integration/Client/Client.csproj` and `Integration/Api/Api.csproj` — each define a `BuildDocker` target that wipes and repopulates the **same** directory, `Source/Kernel/Server/out`. Neither references the other, so MSBuild is free to schedule them in parallel and they collide. The visible symptom is an intermittent `MSB3231`. The invisible symptom is worse.

## Evidence

`Integration/Client/Client.csproj:38,45-46`

```xml
<Target Name="BuildDocker" AfterTargets="Build" Condition="'$(DisableDockerBuild)' != 'true'">
    <RemoveDir Directories="$(MSBuildThisFileDirectory)../../Source/Kernel/Server/out" />
    <Exec Command="dotnet publish $(MSBuildThisFileDirectory)../../Source/Kernel/Server/Server.csproj -c Debug -p:DefineConstants=DEVELOPMENT --no-restore $(_DockerRidArg) -o $(MSBuildThisFileDirectory)../../Source/Kernel/Server/out" />
```

`Integration/Api/Api.csproj:31-33,38-40`

```xml
<Target Name="BuildDocker" AfterTargets="Build" Condition="'$(DisableDockerBuild)' != 'true'">
    <RemoveDir Directories="$(MSBuildThisFileDirectory)../../Source/Kernel/Server/out" />
    <MakeDir Directories="$(MSBuildThisFileDirectory)../../Source/Kernel/Server/out" />
    <Copy SourceFiles="@(_ServerFiles)" DestinationFolder="$(MSBuildThisFileDirectory)../../Source/Kernel/Server/out/%(RecursiveDir)" />
```

Both are listed in `Chronicle.slnx:54-55`. `Benchmarks/Chronicle.Benchmarks/Chronicle.Benchmarks.csproj:30-32` writes the same path as well, but is not in the solution, so it only participates in an explicit build.

## Reproduction

```
dotnet build
```

from the repository root, with default parallelism and `DisableDockerBuild` unset. Intermittent — it requires `Api`'s `Copy`/`docker build` to overlap `Client`'s `RemoveDir`.

## Impact

[...]

PR #3679, opened 2026-08-11
Run the production image in Aspire production mode, and let it be given certificates

woksin:
## Added

- `WithTlsCertificate` and `WithEncryptionCertificate` on the Aspire builder, which mount a certificate into the Chronicle container read-only and point the configuration at it (#3607)

## Changed

- An Aspire AppHost configured for production must now supply a TLS certificate, and an encryption certificate when the internal OAuth authority is enabled — without them the production image does not start (#3607)

## Fixed

- An Aspire AppHost that configures Chronicle for production now runs the production image, instead of the development image that ships development credentials and an anonymous state-reset endpoint (#3607)
- A TLS certificate without a password now loads, instead of failing the server's startup with a cryptographic error
- A certificate path that does not exist now fails the AppHost immediately, instead of producing a container that reports itself running while its process has died (#3607)
- Configuring either certificate twice now replaces the earlier call, instead of producing a container Docker refuses to start (#3607)

## Security

- An Aspire AppHost configured for production no longer runs a development-compiled Chronicle, which seeds a documented default administrator password and a fixed OAuth client secret, and exposes an unauthenticated endpoint that erases the event store (#3607)

### DIFF
--- a/Documentation/client-snippets/hosting/aspire/certificates.md
+++ b/Documentation/client-snippets/hosting/aspire/certificates.md
@@ -1,0 +1,18 @@
+```csharp
+using Aspire.Hosting;
+using Aspire.Hosting.ApplicationModel;
+using Cratis.Chronicle.Aspire;
+
+public static class HostingAspireCertificates
+{
+    public static IResourceBuilder<ChronicleResource> ConfigureAppHost(IDistributedApplicationBuilder builder)
+    {
+        var mongo = builder.AddConnectionString("chronicle-mongo");
+
+        return builder.AddCratisChronicle("chronicle", chronicle => chronicle
+            .WithMongoDB(mongo)
+            .WithTlsCertificate("certs/chronicle.pfx", "YourPassword")
+            .WithEncryptionCertificate("certs/encryption.pfx", "YourPassword"));
+    }
+}
+```
--- a/Documentation/client-snippets/hosting/aspire/complete-example.md
+++ b/Documentation/client-snippets/hosting/aspire/complete-example.md
@@ -9,8 +9,10 @@ public static class HostingAspireCompleteExample
 
         var mongo = builder.AddConnectionString("chronicle-mongo");
 
-        var chronicle = builder.AddCratisChronicle("chronicle", c =>
-            c.WithMongoDB(mongo));
+        var chronicle = builder.AddCratisChronicle("chronicle", c => c
+            .WithMongoDB(mongo)
+            .WithTlsCertificate("certs/chronicle.pfx", "YourPassword")
+            .WithEncryptionCertificate("certs/encryption.pfx", "YourPassword"));
 
         builder.AddContainer("api", "my-org/my-api")
             .WithReference(chronicle);
--- a/Documentation/get-started/choose-hosting-model.mdx
+++ b/Documentation/get-started/choose-hosting-model.mdx
@@ -294,6 +294,15 @@ MongoDB) and the `With*` method you pick sets the kernel's storage type and conn
 resource you hand it — the same two environment variables you saw in the Compose files, now derived from
 your app model instead of hand-written:
 
+<Aside type="caution" title="The production image also wants certificates">
+Passing a configure callback moves you off the development image, and the production image does not
+generate the TLS certificate it serves port 35000 with, nor the certificate that protects its OAuth keys —
+it throws at startup and stops. The snippets below show the storage wiring only, so pair them with
+`WithTlsCertificate(...)` and `WithEncryptionCertificate(...)`; the
+[Aspire integration guide](/chronicle/hosting/aspire/) shows both. If you only wanted a kernel to develop
+against, stay on `AddCratisChronicle()` above — it brings its own database *and* its own certificate.
+</Aside>
+
 <Tabs syncKey="database">
 <TabItem label="MongoDB">
 
--- a/Documentation/get-started/choose-hosting-model.mdx
+++ b/Documentation/get-started/choose-hosting-model.mdx
@@ -289,7 +289,7 @@ For development, one line gives you the same batteries-included kernel as the `d
 
 <ChronicleClientTabs snippet="get-started/choose-hosting-model/basic-apphost" />
 
-To choose a database, pass a configure callback. Aspire then switches to the slim image (no embedded
+To choose a database, pass a configure callback. Aspire then switches to the production image (no embedded
 MongoDB) and the `With*` method you pick sets the kernel's storage type and connection string from the
 resource you hand it — the same two environment variables you saw in the Compose files, now derived from
 your app model instead of hand-written:
--- a/Documentation/hosting/aspire.mdx
+++ b/Documentation/hosting/aspire.mdx
@@ -29,7 +29,7 @@ For production or staging environments, use the configure callback. This switche
 |------|--------|------------|
 | A database | `WithMongoDB` / `WithPostgreSql` / `WithMsSql` / `WithSqlite` | There is no storage to connect to |
 | A TLS certificate | `WithTlsCertificate` | The server throws `No TLS certificate is configured` and exits |
-| An encryption certificate | `WithEncryptionCertificate` | The server throws `An encryption certificate is required in production` and exits |
+| An encryption certificate | `WithEncryptionCertificate` | The server throws `An encryption certificate is required in production` and exits — unless you turn the `OAuthAuthority` feature off or point `Authentication:Authority` at an external authority, which skips the internal authority that needs it |
 
 :::caution[The production image will not start without certificates]
 The development image generates a self-signed TLS certificate on the fly and falls back to ephemeral OAuth keys. The production image does neither — the production code paths are selected by a compile-time symbol, so no environment variable switches them back. A configure callback that wires up only a database therefore produces a container that starts, throws, and stops. Each storage section below shows **only** the storage wiring; add the two certificate calls from [Certificates](#certificates) to every one of them.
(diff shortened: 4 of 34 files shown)


woksin:
Reviewer context, kept out of the body. This is the code arm of #3607; the documentation arm shipped in #3677, and **#3607 stays open until both have landed.**

## The defect

`ChronicleDistributedApplicationBuilderExtensions.cs:53` selected `DevelopmentSlimTag` on the **production** branch, contradicting the method's own XML doc at `:23-27`. It matters beyond tidiness: only `latest-development` and `latest-development-slim` are compiled with the `DEVELOPMENT` symbol, so an AppHost that configured Chronicle for production was running a development-compiled kernel.

Corroborating evidence that this was never intended: `ChronicleContainerImageTags.Tag` had **zero** references anywhere in the repository — a dead public constant, exactly what you would expect if the production branch never used it. Verified by repo-wide sweep before and after.

## The real finding is the coverage, not the line

An earlier analysis of this issue claimed existing Aspire specs *pinned the wrong behavior* and would have to be changed. **That is false, and worth correcting in the record.** `Source/Clients/Aspire.Specs/` had exactly three files, all targeting `AddCratisChronicleMongoDB`. **Nothing covered `AddCratisChronicle` at all**, so correcting line 53 turns nothing red — there was no wrong-behavior pin to update, there was no pin.

Public API with zero coverage is the defect the repo's own framework rules name explicitly, so the specs are the bulk of this change: 13 assertions enumerated from the state machine (development mode / production mode / explicit name). The API exposes no image-tag parameter — only `name` and `configure` — so that dimension genuinely does not exist rather than being skipped.

## Mutation evidence

Every one of the 13 assertions was killed by an actually-applied mutation, not a reasoned one. **Re-verified independently before merge:** reinstating the bug (`Tag` → `DevelopmentSlimTag`) fails exactly **2 of 29** — `should_use_the_production_image_tag` and its paired negative `should_not_use_the_development_slim_image_tag` — and nothing else. The mutation discriminates rather than reddening the family, so the fix is provably covered and the other 11 assertions are pinned elsewhere.

The connection-string assertion was deliberately rewritten to derive the resource name from `_result.Resource.Name`, so it dies to one line rather than two.

## Documentation

The false claim was in **four** places in `aspire.mdx`, not the two originally identified — `:26`, `:37`, `:45`, `:53` all described the production path as "the slim image". The tag table at `:109-113` was not itself false but was where the ambiguity lived, so it gained a "Selected by" column mapping each tag to its call, marking `latest-development-slim` as not selected by the Aspire integration.

**One file outside the planned scope:** `Documentation/get-started/choose-hosting-model.mdx:292` carried the identical claim. Left alone it would have contradicted the corrected code, so it is fixed here. Say the word if you would rather this PR were narrower.

## Gates

Debug and Release, both `--no-incremental` with `-m:1`: **0 warnings, 0 errors** (Release sets `TreatWarningsAsErrors`). `Aspire.Specs`: **29 passed, 0 failed** in both configurations — 16 pre-existing plus 13 new.

`-m:1` is required for a full Chronicle build because `Integration/Client` and `Integration/Api` both `RemoveDir` and publish into `Source/Kernel/Server/out`; that race is tracked separately as #3676.

## Not verified

The docs site gate (`Documentation/web && npm run check`) was not run — it lives in the sibling Documentation repository and regenerates from every sibling clone. The edits are prose-only; the single structural change is a GFM table gaining a third column with a matching separator row.

woksin:
Scope widened after review, and relabelled `patch` → `minor`. `35a96d14a`, `fc9826956`.

## Why it grew

The one-line tag fix was correct and **unshippable on its own**. A reviewer ran the published `cratis/chronicle:latest` with exactly the environment `WithMongoDB` sets and it terminates immediately:

```
System.InvalidOperationException: No TLS certificate is configured.
```

`CertificateLoader.cs:24-35` returns null when TLS is enabled with no path, and `Program.cs:87-100` self-generates **only** under the `DEVELOPMENT` symbol. The OAuth encryption certificate behaves the same way. `IChronicleAspireBuilder` had no way to supply either — so correcting the image would have turned a booting AppHost into a crash loop. That is worse than the defect. Issue #3607's own acceptance criterion 5 asks for exactly this API, so it belongs here rather than in a follow-up.

## API shape, and why

Two methods rather than one: they configure unrelated subsystems (Kestrel's TLS listener vs. the authority's Data Protection keys), fail independently, and either can be present without the other. One combined method would take four positional strings and make "TLS only" inexpressible.

Password is `string? = default`, because both `CertificatePassword` options are nullable — a required parameter would be stricter than the server's own contract.

**Each method mounts the file read-only into the container** and points the configuration at the path *inside* it. Setting a host path would name a file the container cannot resolve, and making the consumer remember a matching `WithBindMount` is boilerplate the framework can own. `IChronicleAspireBuilder.ResourceBuilder` remains public as the escape hatch.

## Config keys verified against source, not against the record

| Env var | Proof |
|---|---|
| `Cratis__Chronicle__Tls__CertificatePath` / `__CertificatePassword` | `Core/Configuration/Tls.cs:20,25` |
| `Cratis__Chronicle__EncryptionCertificate__CertificatePath` / `__CertificatePassword` | `Core/Configuration/EncryptionCertificate.cs:14,19` |

Prefix mechanism at `ChronicleOptions.cs:17,142,148`. It is `CertificatePath`/`CertificatePassword` — **the requirement record for this issue had them as `Path`/`Password`**, which is why every key here was checked against the options class rather than inherited.

## The survivor is closed

A reviewer found that every assertion compared a tag to the constant it came from, so **nothing asserted what the constants say**: repointing `Tag` at `"latest-development"` or even `"not-a-real-tag"` left all 29 specs green while every AppHost pulled the wrong image. Same self-referential shape as a wire contract.

**Re-verified independently:** `Tag = "not-a-real-tag"` now fails `should_publish_the_production_image_as_latest`, and only that. Restored byte-identically, confirmed with `diff`.

## Evidence

58 new assertions, **each killed by an actually-applied single-line mutation and re-run** — 44 mutations in total. Scenarios enumerated from the state machine: no certificate / TLS only / encryption only / both / one file serving both / neither password, plus the named-and-configured cell that was missing. The four `should_not_configure_*` negatives are paired with a positive baseline that keeps them non-vacuous, and a stray unconditional mount is caught by `should_not_mount_anything_into_the_container`.

## A kernel defect found and deliberately not fixed here

`Source/Kernel/Server/CertificateLoader.cs:38-53`: with no password it calls `X509CertificateLoader.LoadCertificateFromFile`, which cannot read PKCS#12 and yields no private key — so a **password-less `.pfx` cannot serve TLS**, even though `Tls.CertificatePassword` is nullable. The new documentation tells you to supply a password. The kernel fix (always `LoadPkcs12FromFile`) is a separate change and does not belong in an Aspire PR.

## Gates

`Aspire` and `Aspire.Specs`, Debug and Release, `--no-incremental`: **0 warnings, 0 errors**. `Aspire.Specs`: **87 passed, 0 failed** in both configurations (was 29). `Documentation/validate-client-snippets.py` compiles the new `certificates.md` and the updated `complete-example.md` against the real Aspire assembly — passes.

## Not verified

`npm run check` in the sibling Documentation repository (no `node_modules` installed there; a full install and Astro build would be needed). Link shapes follow the repo's conventions. No AppHost was launched end to end against a real certificate — the mount and configuration are pinned by spec, not observed in a running container.

woksin:
All four re-review findings fixed. `739a99b73`, `f0d677eba`.

## The kernel regression is worse than first reported, and is fixed here

A password-less `.pfx` did not merely lose its private key — the server died with `CryptographicException: ASN1 corrupted data` and never started. It is **asymmetric**: a password-less *encryption* certificate boots fine, because OpenIddict uses `LoadPkcs12FromFile` whose password is nullable. Only TLS was affected. And it is a **regression** from the `X509CertificateLoader` migration — the `#if NET8_0` branch read PKCS#12 correctly, the net9/net10 branch did not.

Fixed here rather than deferred, because deferring guarantees a shipped release where `WithTlsCertificate(path)` compiles and hard-crashes, and the API and the kernel ship from the same version. There is nothing for the password to select between: TLS needs the private key only PKCS#12 carries, so the other path could never have produced a usable certificate.

**Re-verified independently:** reinstating the old branching fails exactly the three password-less facts (`should_load_the_certificate`, `should_load_the_configured_certificate`, `should_load_the_private_key_the_tls_listener_needs`) with the password-protected baseline, TLS-disabled and missing-file specs all staying green. Restored byte-identically, confirmed with `diff`.

## The other three

**A missing path reported the opposite of the truth.** Docker creates a *directory* when a bind mount source is absent; the kernel then logged "No TLS certificate is configured — provide one" when one *was* provided, and the container stayed `Status=running` with a dead process, so Aspire showed it green. The path is now resolved against `AppHostDirectory` and checked before mounting, failing with a named domain exception that reports both the configured path and what it resolved to.

**Configuring either certificate twice** appended a second mount on the same target, which Docker refuses (`Duplicate mount point`). Every other `With*` on this builder is last-call-wins, so a base helper plus an override silently produced an unstartable container. The later call now replaces the earlier one — including clearing a password the overriding call does not supply.

**The "cannot start without this either" absolute** is softened in both the XML doc and the table, naming the `OAuthAuthority` and external-authority escapes that `OpenIddictServiceCollectionExtensions.cs:33` actually takes.

## Mutation calibration — stated honestly

The implementation reported per-mutation whether a kill is **specific** or **class-redden collateral**, which corrects an overstatement in my earlier comments on this PR. Summary:

- Kernel branch restored → 3 failures, **each specific** (`Because` uses `Catch.Exception`, so no collateral).
- Existence check deleted → 6 failures, **specific per fact**.
- Mount-removal loop deleted → 4, **specific**. Password-clearing branch deleted → 1, **specific**.
- Dropping `AppHostDirectory` → 42 failures, but only **3 are specific kills**; the other 39 are collateral because `Because` throws. One spec deliberately plants a decoy file at the same relative path in the working directory so that `Because` still succeeds and the assertion discriminates.
- One attempted mutation (`Path.Combine` instead of the explicit join) produced **0 failures** and was reported as not a real mutation on POSIX, since `Path.Combine` returns a rooted second argument unchanged — rather than being quietly dropped from the tally.

## Gates

`Source/Clients/Aspire` and `Source/Kernel/Server`, Debug and Release, `--no-incremental`: **0 warnings, 0 errors**. `Aspire.Specs` **103 passed** (was 87). `Server.Specs` **33 passed**, 25 of those facts new.

## Found and deliberately not fixed

`Source/Clients/Connections/CertificateLoader.cs:66-81` carries the **identical** password-less bug on the *client* side, reached from `ChronicleConnection.cs:338`. That project multi-targets, so its `#if NET8_0` branch is live and the net9/net10 branch is not. It is not caused by anything in this PR and belongs in its own change — flagging it here so it is not lost.

Issue #3689, opened 2026-08-12
ReadModelScenario drops a same-event child projection when the event also maps parent fields


woksin:
`ReadModelScenario<T>` silently drops a child collection when a single event type both maps parent fields and feeds a child projection. The parent scalars materialize; the child stays empty. No exception, no warning — the spec just goes green on a false negative.

Reproduces in both fluent forms, which share one definition (`AddChild` desugars into `Children(...).From<TEvent>(...)` in `ReadModelPropertiesBuilder.cs:137-148`):

- `.From<E>(...set parent...)` combined with `.Children(x => x.Items, c => c.From<E>(...))`
- `.From<E>(from => from.Set(...).AddChild(m => m.Items, child => child...))`

**Mechanism.** The harness resolves one key from the *root* projection's resolver map and passes that same context to every child unchanged — `Source/Clients/Testing/ReadModels/ProjectionReadModelProcessor.cs:565-576` (`HandleEventFor`) does no child key resolution at all. The root map only inherits a child's resolver for event types the root does not already own (`ProjectionFactory.cs:827-836`, see the TODO at :834), so when the event also has a root `From`, the root resolver wins and the resulting key carries no array indexer for the child property. The child then returns without applying at `ProjectionEventContextExtensions.cs:151-156`.

Child-only events work because the root inherits the child's indexed resolver — which is why all ~20 existing specs under `when_projecting_with_children_from/` pass. They are key-type variations, and every one of them uses a *different* event for parent and child.

**The production kernel is not affected.** `Source/Kernel/Core/Projections/Engine/Pipelines/Steps/HandleEvent.cs:42-69` asks each child for its own resolver and recurses with a re-keyed context. The integration projections under `Integration/Client/Projections/Scenarios/when_projecting_from_event_that_adds_child_from_object/` are exactly this shape and are green against a real kernel. So this is a `Cratis.Chronicle.Testing` fidelity defect: the harness disagrees with the engine it is supposed to stand in for, in the silent direction.

**Fix.** Make `HandleEventFor` mirror `HandleEvent.Perform` — per-child `HasKeyResolverFor` / `GetKeyResolverFor`, a re-keyed context, including the deferred and unresolvable arms.

**Regression cover.** An event doing both root setters and a keyed child append, with at least two occurrences so child-key collapse is visible, in `Source/Clients/Testing.Specs/ReadModels/for_ReadModelScenario/when_projecting_with_children_from/`.

[...]

PR #3693, opened 2026-08-12
Resolve a child projection's own key in the testing harness

woksin:
## Fixed

- `ReadModelScenario` no longer drops a child collection when one event type both maps parent fields and feeds a child projection, in both the `Children(...).From<T>` and `AddChild(...)` forms and for model-bound `[ChildrenFrom<T>]` (#3689)


woksin:

**Mechanism.** `HandleEventFor` passed the root's `ProjectionEventContext` to every child unchanged, doing no child key resolution. `ProjectionFactory.cs:827-836` folds a child's key resolver into the root's map only for event types the root does *not* already own (see the TODO at :834), so a child-only event inherits the child's indexed resolver and works, while an event that also has a root `From` lets the root's resolver win — the key then carries no array indexer and the child returns early at `ProjectionEventContextExtensions.cs:151-156`. That is why all ~20 existing specs under `when_projecting_with_children_from/` pass: they are key-type variations and every one uses a *different* event for parent and child.

The fix mirrors `Source/Kernel/Core/Projections/Engine/Pipelines/Steps/HandleEvent.cs:42-69`. One deliberate deviation: the kernel calls `context.AddDeferredFuture(...)`, but the harness has no future store and never reads `DeferredFutures`, so a deferred child key skips the child instead — the same outcome the kernel reaches for that event. Commented at the site.

**Model-bound path also reproduces** — `[FromEvent<ShiftLogged>]` plus `[ChildrenFrom<ShiftLogged>]` hits the identical branch, covered by a third spec.

**Mutation evidence.** Verified independently of the implementer, with an assert that the pattern matched exactly once before running.
- Reverting the re-key to `HandleEventFor(child, context, …)` — exactly the pre-fix behavior — turns all three new specs RED. **SPECIFIC:** `should_have_an_instance` and `should_map_the_parent_scalar` still *pass*, `Because()` does not throw, and only the child assertions die. That passing-parent / failing-child split is the defect's own signature, which is what makes the kill discriminating rather than collateral.
- Dropping the `HasKeyResolverFor` guard: RED, 157 of 400.
- Restore verified with `shasum -a 256` (never `diff`): `73be3481…b8d6e9` before and after.

**Three surviving mutations, reported rather than papered over:** dropping the `DeferredKey` arm, dropping the `UnresolvableKey` arm, and removing the resolved-key memoization all stay GREEN. No spec in `Testing.Specs` drives a child key resolver to a deferred or unresolvable result for an event type the root already owns. Both arms are still correct to keep — without them the `ResolvedKey` cast NREs — but they are guards nothing exercises. The memoization survival is expected (it is an optimization), though its removal takes the suite from 4s to 11s, so it is load-bearing for runtime.

**Gate.** Debug `--no-incremental -m:1`: 0 errors. Release: 0 errors, 0 warnings. `Testing.Specs` 400/400 passed in Release.

**Pre-existing Release break, not from this PR.** `Source/Kernel/Contracts` fails Release with `PBN2013` ("protobuf-net contracts and no `[ProtoModel]`") — a warning in Debug, an error in Release, and not in that project's `NoWarn` (which lists `PBN2008`/`PBN0022`). I reproduced it on a clean worktree that this branch never touched, building `Contracts.csproj` alone, with `protobuf-net.BuildTools` pinned at 3.3.8; `-p:TreatWarningsAsErrors=false` does not suppress it. The Release figures above therefore used `-p:WarningsNotAsErrors=PBN2013`, which neutralizes only that one unrelated diagnostic. Filed separately.

**Not verified:** integration specs (container-baked kernel) were not run; the kernel itself is unchanged. No full-solution build. Whether making `HandleEventFor` async changes ordering for a projection that mutates shared state re-entrantly was not constructed — the resolvers here complete synchronously against in-memory storage.

woksin:
Added a second commit that is **not** part of the fix, to unblock CI.

A high-severity advisory for SSH.NET (GHSA-q939-rpr3-3284) was published 2026-08-12 15:19 UTC. SSH.NET arrives transitively through Testcontainers, and 4.13.0 is the latest release and still resolves the vulnerable 2025.1.0 — so there is no upgrade path. Every `dotnet-build` in this repo (and in Arc) started failing `NU1903` within hours, including PRs that had nothing to do with it; #3452's build passed only because it ran before the advisory landed.

Pinning a transitive package under central package management needs `CentralPackageTransitivePinningEnabled`, which then enforces the centrally declared version for *all* transitively pulled packages. That surfaced three `NU1109` downgrades where `Microsoft.Extensions.DependencyModel`, `Microsoft.Extensions.Logging.Configuration` and `Microsoft.EntityFrameworkCore` were still on 10.0.10 while eight sibling `Microsoft.Extensions.*` entries had already moved to 10.0.11. Aligning those three is what the rest of the manifest already assumed; it is not a version bump chosen here.

Verified: `dotnet restore Chronicle.slnx` goes from failing to **50 projects, 0 errors, 0 warnings**, `SSH.NET/2026.0.0` is what actually resolves in the lock graph, and `Testing.Specs` still passes 400/400 afterwards.

It is deliberately kept out of the PR body — Testcontainers is test-only, so nothing about this reaches a published package and it is not release-note material. It rides here rather than in a separate PR at the requester's direction. The same pin is on Arc #2539; the other two Chronicle PRs (#3694, #3695) deliberately do **not** carry it, to avoid three-way conflicts on the same manifest — they will pick it up from `main` once this merges.

Issue #3613, opened 2026-08-09
JsonSchema.Validate accepts wrong property value types


woksin:
# CHR-55 — JSON schema validation accepts wrong property value types

**Status:** Resolved upstream · filed 2026-08-10 · fixed in [Cratis/Chronicle#3694](https://github.com/Cratis/Chronicle/pull/3694) and released in **v16.27.0**

> ✅ **Resolved 2026-08-13.** Chronicle#3694 restored declared property-type validation for `JsonSchema.Validate`, including nested objects, array items, and `$ref` resolution. This deliberately tightens append-time validation. It was not verified against a corpus of stored events; nested `required`, `enum`, `minimum`, `pattern`, and `format` validation remain outside this fix.

## Symptom

Chronicle's public `JsonSchema.Validate(...)` methods report a JSON object as valid when a declared property has
the wrong JSON value type. A schema declaring a required integer property accepts both a string and an object at
that property. A missing required property is rejected, so callers receive a partial validation result that can
look like complete schema validation.

The XML contract for `Validate(string)` explicitly promises "basic type and required property checks". The
`Validate(JsonObject)` overload says it is equivalent for already-parsed object content.

## Reproduction and mechanism

At the current Chronicle branch, `Source/Infrastructure/Schemas/JsonSchema.cs` parses object content and then
calls only `ValidateRequiredProperties(...)` from both overloads. It never compares each supplied value with the
declared property schema. Consequently both of these payloads pass a schema whose required `status` property is
an integer enum:

```json
{ "status": "not-an-integer" }
{ "status": { "unexpected": true } }
```

while this payload is correctly rejected:

```json
{}
```

## Ada cost and impact

Ada exposed the defect while adding Chronicle's regression coverage for [Chronicle#3603](https://github.com/Cratis/Chronicle/issues/3603).
That fix must validate plaintext before encrypting PII leaves. A deliberately type-invalid enum payload did not
exercise the rejection path because `JsonSchema.Validate(JsonObject)` accepted it; the regression had to narrow
its malformed input to an absent required property.

This means Chronicle's event append boundary does not currently enforce the declared JSON representation beyond
presence. Malformed property shapes can enter the event log even though the append path reports schema
validation success. A later serializer, compliance pass, projection, or client can then fail farther from the
writer that supplied the invalid content.

## Requested fix

- Make both `Validate(string)` and `Validate(JsonObject)` validate supplied property value types recursively
  against the schema they expose.
[...]

Issue #3632, opened 2026-08-10
Resolved: default unique-constraint violation messages exposed client-visible identifiers


woksin:
# Resolved: default unique-constraint violation messages exposed client-visible identifiers

## Resolution

This historical report is resolved.

- Chronicle v16.16.0 removed the offending unique value from `UniqueConstraintValidator`'s default message.
- The remaining `EventSourceId` interpolation in `UniqueEventTypeConstraintValidator` was removed in #3713 and released in Chronicle v16.34.0.
- The event-source ID was not moved to violation details, because those details reach the same caller. A caller that needs its own event-source ID already has it.

## Original problem

When a consumer declared a unique constraint without supplying its own message, Chronicle's server-composed default message reached `CommandResult.ValidationResults`. The original unique-value message interpolated the offending value; the sibling unique-event-type message interpolated the event-source ID.

The relevant defaults are therefore now safe from those two unintended client-visible interpolations. This issue is retained as the historical report and resolution record.

woksin:
**The main ask shipped in v16.16.0** as `88a6e46e4`, "Carry the offending value, not its index key, in a unique constraint violation detail". `UniqueConstraintValidator` now puts `PropertyValue = pv.Value` — the actual offending value — into the violation details, rather than the SHA-256 index key. Verified against current `origin/main`.

**Leaving this open for one residual**, because closing it would quietly drop something worth a decision.

Both default violation messages still interpolate `EventTypeId` and the sequence number, and `UniqueEventTypeConstraintValidator` still interpolates `context.EventSourceId` into a `ValidationResult` — which is client-visible. An event-source id reaching an end user's validation error is a small information leak, and the internal type id and sequence number are noise to whoever reads the message.

Two things make this cheap to act on:

- **No spec asserts either string.** The literals appear only in the two source files, so trimming them requires no spec change — this is a `patch` needing a product nod, not a redesign.
- It is the same client-visible rejection surface that #3634's sibling work touches, so if anything else is in flight there it should land in one pass.

If the current messages are considered fine as they are, say so and this can close on that basis.

woksin:
## Residual re-verified at `origin/main` (v16.26.0) — and the issue's own framing understates it

**The primary ask shipped.** `UniqueConstraintValidator.cs:67` now composes:

```
Event '{context.EventTypeId}' on member '{pv.Property}' violated a unique constraint on sequence number {sequenceNumber}
```

[...]

PR #3694, opened 2026-08-12
Validate property value types against the schema

woksin:
## Fixed

- Schema validation now checks property value types, so a payload whose value does not match its declared type is rejected at append instead of being reported as valid (#3613)
- An object schema handed a non-object value now reports a wrong-property-type error; the check was inverted and fired only when the schema did not want an object (#3613)


woksin:

**Not NJsonSchema.** `JsonSchema` is Chronicle's own type, introduced by `2d3ded266` ("Add custom JSON schema types to replace NJsonSchema"). The replacement kept `ValidateRequiredProperties` and dropped value-type validation entirely, so this is a regression from that removal rather than a never-implemented feature. Corroboration that it was a stub with a finished-looking vocabulary: `JsonSchemaValidationErrorKind` declares `WrongPropertyType`, `AdditionalPropertiesNotAllowed` and `FormatMismatch`, and the latter two were emitted nowhere at all.

**Severity.** The one production caller is `EventSequence.cs:834` inside `ValidateAgainstSchema`, which converts errors into `ConstraintType.Schema` / `SchemaValidation` violations on the append path — so wrong-typed data was appended and reported as schema-valid.

**Which node is walked, and why.** `FlattenedProperties` (the list), not `FlattenedPropertiesByName` and not the raw `properties` node. `FlattenedProperties` also traverses `allOf`, which is how Chronicle expresses inherited event shapes, and `allOf` merges into the same object level so dotted paths stay correct. `FlattenedPropertiesByName` was avoided for two reasons: it is `OrdinalIgnoreCase`, which would apply a schema to a differently-cased property, and its `ToDictionary` throws if `allOf` ever yields two names differing only by case — a pre-existing latent throw in `ValidateRequiredProperties` that this change deliberately does not extend exposure to.

**Deliberate conservatism, because this tightens the append boundary.** Absent `type` constrains nothing; an unresolvable `$ref` falls back to the referencing node; an ambiguous union with two or more kind-compatible branches stays silent rather than blaming an arbitrary branch (this is what keeps polymorphic `oneOf` payloads from being falsely rejected); an unreadable numeric representation is accepted. Nullability is `Type.HasFlag(Null)` or a trailing `?` on `format`.

**Mutation evidence.** Verified independently of the implementer, with an assert that the pattern matched exactly once first.
- `JsonValueKind.String => type.HasFlag(JsonObjectType.String)` changed to `=> true`: `and_a_string_is_supplied_for_an_integer_property` goes **3 of 4 RED**, each dying at its own assertion with an empty result list, `Because()` never throwing. **SPECIFIC.** The fourth, `should_agree_across_both_overloads`, correctly survives — it asserts overload parity, and both overloads returned empty together.
- Restore verified with `shasum -a 256` (never `diff`): `cd047a2d…39185` before and after.
- The implementer additionally killed the path-combining, whole-number, `$ref`-following and wire-up lines, each SPECIFIC; the whole-number one kills in both directions, proving the must-PASS assertion is not vacuous.

**No pre-existing spec regressed.** `Infrastructure.Specs` 656/656 in Debug and Release. Kernel `Core.Specs`, which owns `ValidateAgainstSchema`, 2467 passed / 1 skipped / 0 failed. The three pre-existing `when_validating` specs pass unchanged, including the one using `.Single()`.

**Gate.** Debug and Release `--no-incremental`: 0 errors, 0 warnings for both `Infrastructure` and `Infrastructure.Specs`.

**Semver: minor, deliberately not patch.** The public signatures and the error-kind enum are untouched, and this restores intended pre-`2d3ded266` behavior rather than changing a contract — so `patch` is defensible. It is `minor` because a patch bump would let an unattended upgrade start rejecting writes that a deployment accepts today, and that deserves a release note an operator actually reads.

**Not verified, and this is the main residual risk:** the new validation was never run against a real corpus of stored event content. The over-rejection argument rests on the conservative fallbacks above plus 3123 green specs, not on production data. Also deliberately out of scope: nested `required` is still only checked at the top level, and `additionalProperties`, `format`, `enum`, `minimum`/`maximum`, `pattern` and `minLength` remain unvalidated — adding any of them would tighten the boundary considerably more.

woksin:
Independent security review. One real issue, now fixed in a follow-up commit; the rest came back clean.

**Exponential work on the append path.** When a union had two or more branches accepting the value's kind, the code probed each with a full recursive validation. Since a probe is itself a full validation, a nested union branched again at every level — `k^depth` work reachable from a single append payload. `JsonNode` parsing caps depth at 64 so a stack overflow is not reachable, but the CPU cost is.

Worse than wasteful: **the probing decided nothing.** The probe collected into a discarded local, so whether one branch validated or none did, the ambiguous case always accepted and never added an error. The loop was observationally dead. Removing it is behavior-preserving — 656 specs pass unchanged, Debug and Release, 0 warnings — and removes the blowup.

**Clean:** `$ref` cycles cannot loop — recursion is *content*-driven, descending only into an object or array actually present in the payload, so a self-referential `$ref` terminates when the payload stops nesting, and `Effective()` follows one hop. **Error messages do not repeat #3632** — `Mismatch` emits only the schema-declared property path and a JSON *kind* name (`"string"`, `"number"`, `"null"`), never the offending value. An attacker cannot *force* a fail-open: content only selects among branches the registered schema already declares.

**Carried, not fixed — worth knowing before release.** `AllowsNull` consults only `Type.HasFlag(Null)` or a `Format` ending in `?`, and `Effective()` discards the referencing node in favour of the `$ref` target. A nullable *complex* property declared as `oneOf:[{$ref:X}]` plus a nullable marker would now be rejected for a legitimate `null`. The new specs cover nullable **primitives** only, so that is precisely the uncovered corner. Also note `{"count": null}` for a non-nullable property is now a hard failure where it previously appended, and JavaScript producers routinely serialize `undefined` as `null` — the outcome is loud (`AppendResult.Failed`, `ConstraintType.Schema`; the event is rejected at the call, not lost or quarantined), but it is a behaviour change for existing writers and is the reason this is labelled `minor` rather than `patch`.

Issue #3604, opened 2026-08-09
Add an opt-in health-only dedicated listener


woksin:
## Problem

Chronicle's `health.port` creates a second HTTP/1.1 Kestrel listener on the same ASP.NET Core application. Every HTTP/1.1 route remains mapped there, including Workbench, REST, identity/OAuth, static files, and connection management.

Current documentation accurately discloses this, so the existing behavior is intentional. The missing capability is a genuinely health-only listener for deployments that cannot rely only on network policy.

## Proposed capability

Add an opt-in setting such as `health.exclusive: true` or a separately named health-only listener mode.

## Acceptance criteria

- Absent/disabled preserves existing behavior.
- Enabled serves the configured health endpoint on the dedicated port and rejects every other route.
- The main port keeps all existing endpoints, including health.
- Health TLS, equal-port handling, IPv4/IPv6 binding, and forwarded-header behavior remain unchanged.
- Routing uses the actual local listener/port, never spoofable host or forwarding headers.
- Specs probe health, identity, REST, static/Workbench, connection-count, and unknown paths on both ports.
- Documentation distinguishes an additional HTTP/1.1 listener from a health-only listener.

This is additive, opt-in, and suitable for a minor release.

woksin:
Release-train note: please group this opt-in addition as a logical commit with the existing Chronicle train after #3599 is made minor-compatible. The current health-port behavior must remain the default, so this does not require a major release.

woksin:
Triage: **confirmed not achievable with current configuration** — this needs implementing as specified.

`Source/Kernel/Server/Program.cs:120-131` binds the health port as a second Kestrel endpoint on the *same* `WebApplication`, and every HTTP/1.1 route is mapped app-wide with no port predicate — workbench static files `:313`, `/identity` `:344-346`, `MapControllers` `:352`, health `:358`, connection-count `:363`/`:371`, fallback `:383`. There is no `RequireHost`, `UseWhen`, `MapWhen` or `LocalPort` check anywhere under `Source/Kernel`. The documentation states the exposure outright at `Documentation/hosting/configuration/health-endpoint.md:45`: "an additional listener on the same application rather than a separate, health-only server. The HTTP/1.1 endpoints (Workbench, REST API, OAuth) are therefore reachable on it too."

The Kernel Server is top-level statements in a shipped executable, so a consumer cannot inject the filter themselves. Network policy is the only current mitigation, which is the case this issue explicitly rules out.

[...]

PR #3695, opened 2026-08-12
Add an opt-in health-only dedicated listener

woksin:
## Added

- `health.exclusive`, which restricts the dedicated health port to the health endpoint so the Workbench, REST API and OAuth endpoints are no longer reachable on it. Off by default (#3604)


woksin:

**Premise verified before changing anything:** `grep -rn -e RequireHost -e UseWhen -e MapWhen -e LocalPort Source/Kernel` returns zero hits, so no port predicate existed anywhere in the Kernel. `Program.cs:120-131` binds the health port as a second Kestrel endpoint on the same `WebApplication`, and every route is mapped app-wide — Workbench static files, `/identity`, `MapControllers`, the connection-count endpoints and the fallback all answered on it.

**Non-spoofable by construction.** The decision keys on `HttpContext.Connection.LocalPort` — the port the socket actually accepted on — rather than `Host` or `X-Forwarded-*`, which are client-supplied. That was an explicit acceptance criterion on the issue.

**Why a new type.** `Program.cs` is top-level statements, so the logic would otherwise be unreachable from a spec. `HealthOnlyPortPolicy` is pure and references no ASP.NET types, so it lives in `Core` and is directly specifiable; `Program.cs` keeps only the `app.Use` wiring, registered before `UseForwardedHeaders`/`UseRouting`.

**Path comparison, decided deliberately.** Both sides normalized — trim whitespace, strip a trailing `/`, add a leading `/` if the configured value lacks one, collapse empty to `/` — then compared `OrdinalIgnoreCase` for whole-path equality. Case-insensitive because ASP.NET Core routing already matches endpoint paths case-insensitively; a stricter comparison here would 404 a `/Health` probe that the same endpoint serves happily on the main port. Equality not prefix, so `/healthy-secrets` is rejected — specced.

**Mutation evidence, including the fail-open you asked for.** Baseline `7ef7c119…0784`; every run restored original bytes via `trap … EXIT` + `cp` + `touch`, shasum-verified back to baseline on all five.
- **Fail-open** — `return !IsHealthCheckEndpoint(...)` replaced with `return false`, so the port admits everything: **CAUGHT**, 5 failed / 8 passed. Worth noting exactly how: the four "health path is admitted" specs *survive* this mutation, precisely as predicted. Only the rejection specs kill it — `should_reject_the_workbench_root`, `should_reject_the_rest_api`, `should_reject_the_identity_endpoints`, `should_reject_a_path_merely_prefixed_by_the_health_path`. A suite that only asserted admission would have been vacuous here.
- Port comparison inverted → the main-port specs die. **SPECIFIC.**
- `!options.Health.Exclusive` inverted → the not-exclusive spec dies. **SPECIFIC.**
- `OrdinalIgnoreCase` → `Ordinal` → the differently-cased path spec dies. **SPECIFIC.**
- Trailing-slash trim dropped → the trailing-slash spec dies. **SPECIFIC.**

No mutation survived. 21 new facts; `Core.Specs` 2488 passed / 0 failed / 1 pre-existing skip.

**Gate.** Debug `--no-incremental -m:1`: 0 errors, 0 warnings from this change. Release: 0 errors, 0 warnings for `Core`, `Core.Specs` and `Server`. One warning this change did introduce (`RCS1238`, a nested ternary in the normalizer) was found and fixed before the final gate.

**Pre-existing Release break, not from this PR.** `Source/Kernel/Contracts` fails Release with `PBN2013`. It is a warning in Debug and an error in Release, is not in that project's `NoWarn`, is not suppressed by `-p:TreatWarningsAsErrors=false`, and reproduces on a clean worktree building `Contracts.csproj` alone — a project that is *upstream* of everything touched here. The Release figures above therefore used `-p:WarningsNotAsErrors=PBN2013`, neutralizing only that one unrelated diagnostic. Filed separately.

**Not verified:** no end-to-end run. The middleware is covered only through the extracted predicate — that `app.Use` really is first in the pipeline, that `Connection.LocalPort` holds what is expected under Kestrel, and that a 404 short-circuit does not disturb the health endpoint's own routing are all unspecced, because `Program.cs` top-level statements are not reachable from a spec. Integration specs were out of scope. `Exclusive` was bound from an in-memory configuration source but not from a real `appsettings.json`, and it was not surfaced in Helm charts or compose samples.

woksin:
Independent security review found a **real bypass in this PR**, now fixed in a follow-up commit.

**The hole.** The path check trimmed whitespace from the *request path*; ASP.NET Core routing does not. `GET /health%20` on the exclusive port decodes to `Path = "/health "`, which the check trimmed to `/health` and admitted — while routing saw a different string, missed `MapHealthChecks("/health")`, and fell through to `MapFallbackToFile(...).AllowAnonymous()`. Result: **200 and the Workbench entry document on a port advertised as health-only.** Same for `%09`, a leading space, and `/health/%20`. This is the check-versus-routing mismatch class, and it resolved *open* — the worst direction.

**The fix.** The request path now gets only the normalization routing itself performs — trailing slash ignored, case-insensitive, nothing else. The configured endpoint is operator-supplied rather than attacker-supplied and is still trimmed. Separately, a `HealthCheckEndpoint` that is empty or whitespace previously resolved to `/` and served the fallback; it now matches nothing, so the port rejects everything — the safe direction for a misconfiguration on a port whose whole purpose is to expose one thing.

**Proven, not assumed.** Reinstating the old `Trim()` on the request path turns all four new assertions in `and_the_health_path_is_padded_with_whitespace` RED, each at its own assertion; restore verified by `shasum -a 256`. `Core.Specs` is 2500 passed / 0 failed, Debug and Release, 0 warnings.

**Checked and clean by the same review:** `LocalPort` is genuinely non-spoofable (it comes from the accepted socket on a separate `Http1`-only listener, no header feeds it); registering before `UseForwardedHeaders` is correct, and that configuration enables only `XForwardedFor | XForwardedProto`, not `XForwardedPrefix`, so `PathBase` is never rewritten; registering before `UseRouting` short-circuits ahead of static files, auth, controllers and gRPC; `/health/../workbench`, `%2e%2e` and `//health` all fail closed; `OrdinalIgnoreCase` is culture-invariant so there is no Turkish-I or homoglyph bypass; query and fragment are excluded from `Request.Path`; the 404 is uniform and is not an existence oracle.

**Known and accepted:** `Exclusive: true` with no dedicated health port configured is a silent no-op — safe, since no second listener exists, but nothing logs that the hardening is inert. Worth a follow-up.

Issue #3641, opened 2026-08-10
scalar clear declarations compile but silently leave stale projection values


woksin:
## Describe the bug

> **Resolved:** This issue’s diagnostic scope shipped as CHR0047 in Chronicle v16.29.0. The separate scalar-clear feature shipped through [#3711](https://github.com/Cratis/Chronicle/issues/3711) in v16.34.0.

This issue promotes a previously recorded upstream item from a production Cratis consumer. The complete evidence and requested fix follow.

**Status:** Diagnostic scope resolved in Chronicle v16.29.0; scalar-clear feature resolved separately in Chronicle v16.34.0.
**Product:** Chronicle model-bound projections — `[ClearWith]` and `[SetValue<T>]`
**Type:** Correctness / silent no-op

## Symptom and Ada cost

`[ClearWith<T>]` clears a whole `[Nested]` object but cannot clear a scalar property. The apparently equivalent
`[SetValue<T>(null)]` both conflicts with nullable warnings and is silently ignored by projection construction,
leaving the stale value standing. The declaration compiles far enough to look like a supported clear and produces
no Chronicle diagnostic. Ada instead writes type-specific `NotSet` sentinels such as an empty string, which is not
equivalent to null and leaks storage mechanics into the domain shape.

## Requested fix

Provide a scalar clear operation that emits an explicit unset/null change, or diagnose every unsupported scalar
`[ClearWith]` / null `SetValue` declaration at build time. Cover root, child and nested scalar members and replay.

## Honest limits

The current limitations are source-read and recorded in Ada's projection rules. A live-kernel reproduction has not
been run for this filing.

woksin:
Diagnosed in #3702, released in **v16.29.0** as **CHR0047**.

Read the scope honestly: this closes the *silently* half, not the feature. Two declarations that compile and then do nothing are now build warnings:

- **A null `SetValue`.** Verified against every consumer — `ModelBoundProjectionBuilder.cs:506-516` and `:358-367`, `ChildrenDefinitionExtensions.cs:233-242` and `:320-334` all guard `if (value is not null)` with no `else`, and the member is registered as mapped *before* that guard. `ChildrenDefinitionExtensions.cs:343` also counts the attribute toward `hasExplicitMapping` regardless of null, which suppresses CHR0024 as well. So the value stays stale forever, replay included.
[...]

Issue #3609, opened 2026-08-09
Document and diagnose passive projections that re-key events from another stream


woksin:
## Problem

A fluent projection can use top-level `UsingKey(...)` to key an active read model by an event property rather than the event source id. The same projection shape can be combined with `[Passive]` and registers successfully, but a keyed passive read does not locate events by that projected key.

The passive path runs an immediate projection with its event cursor constrained to `eventSourceId = ReadModelKey`. It therefore replays only the physical stream named by the requested key. If the source event was stored on another stream, the read returns a default-initialized model instead of the re-keyed result.

The in-memory `ReadModelScenario<T>` can still pass for the same projection, so this is especially easy to miss until a real-kernel or deployed read.

## Minimal scenario

1. Append `OwnershipEstablished(Hash, OwnerId)` on event source `user-1`.
2. Define a fluent projection for `Ownership`, with `.From<OwnershipEstablished>(from => from.UsingKey(e => e.Hash)...)`.
3. Mark `Ownership` `[Passive]`.
4. Read `Ownership` by key `Hash`.

The active/sink projection can materialize the re-keyed document. The passive immediate read replays the `Hash` stream, never sees the event on `user-1`, and returns default property values.

## Suggested correction

Please make the boundary explicit and hard to misuse:

- Document that a passive keyed read replays only the physical event-source stream named by the read-model key; top-level re-key/composite/constant-key metadata does not provide a cross-stream lookup index.
- Add a Chronicle analyzer diagnostic when a `[Passive]` read model uses a top-level re-keying projection (`UsingKey`, composite key, or constant key), unless the consumer explicitly suppresses it after proving the contributing events are physically co-located under that key.
- Add a real immediate-projection spec that distinguishes an event on the requested stream from one merely carrying the requested key as payload.

If supporting cross-stream passive re-keying is not intended—and it appears incompatible with a sink-free lookup without a global scan or index—the diagnostic/documentation path is sufficient. This issue does not ask Chronicle to add a global event scan.

## Acceptance criteria

- The documentation states the physical-stream requirement for passive keyed reads.
- A consumer cannot accidentally combine `[Passive]` with cross-stream top-level re-keying without a diagnostic.
- Upstream specs pin both the supported co-located case and the unsupported cross-stream case.
- Existing active/sink re-key behavior remains unchanged.

woksin:
Bookkeeping: **the documentation half of this is already shipped**, so what remains open is the diagnostic alone.

[...]

PR #3702, opened 2026-08-13
Diagnose passive re-keying and inert clear declarations

woksin:
## Added

- `CHR0046` warns when a passive read model's projection redirects the root document key, which makes the read replay a stream the events were never appended to (#3609)
- `CHR0047` warns on a `SetValue` declaring null and on `ClearWith` applied to a property or parameter, neither of which the projection builder reads (#3641)


woksin:

**IDs.** CHR0046 and CHR0047. CHR0045 was the highest in use; **CHR0033 is a hole and was deliberately left alone** — `index.md` says gaps may be reserved for a rule in progress, so it was not reused.

**CHR0046** ships as a pair mirroring the existing `KeyRedirectionPiiAnalyzer` / `FluentKeyRedirectionPiiAnalyzer` split, so both the fluent and the model-bound form are covered — the model-bound half turned out to be the same amount of work. The nested `KeyRedirection` class and `IsDocumentRedirectionFor` were promoted out of the PII analyzer into `Analyzers/KeyRedirection.cs` and shared; all 54 pre-existing `for_FluentKeyRedirectionPiiAnalyzer` specs still pass after that move. Passivity is gated on **both** `[Passive]` and a `.Passive()` in the same builder chain, the latter tracked the way `FluentProjectionMappings` already tracks `AutoMapIsOn`.

*Deliberate under-report, documented in the code:* reusing the existing `ContextMemberRedirects` also excludes `c => c.Subject`, which can legitimately differ from the event source id. Conservative in the safe direction — it will miss that shape rather than false-positive on it.

**CHR0047** needs no semantic gating: both shapes are provably inert against Chronicle's own attribute types, so this half is fully statically decidable. Verified against every consumer — `ModelBoundProjectionBuilder.cs:506-516` and `:358-367`, `ChildrenDefinitionExtensions.cs:233-242` and `:320-334`, with `:343` counting a `SetValue` toward `hasExplicitMapping` regardless of null and thereby suppressing CHR0024; and `NestedDefinitionExtensions.cs:135 → :251` reading class-level attributes only.

**Mutation evidence — nine mutations, six of them deliberately false-positive probes**, which is the mutation that matters for an analyzer: proving it does *not* fire on a shape it must leave alone. Each apply→run→restore used `trap … EXIT` with **absolute** paths, and every restore was `shasum -a 256` verified independently of what the trap printed. All five touched files re-verified clean at the end.

**One mutation SURVIVED on the first run, and it is the most valuable result here.** Dropping the child-scope guard (`if (isChildScope) return`) was not caught, because every child-scope spec used `UsingParentKey` — which the `Root` method-name filter already excludes, so the guard itself was never exercised. A new spec, `and_a_child_block_keys_the_child` (a `UsingKey` inside `Children` on a passive parent), now kills it specifically. Without that the guard could have been deleted silently. The other false-positive probes — forcing the passivity gate true, dropping `ContextMemberRedirects`, treating `parentKey` as `key`, forcing `DeclaresNullValue` true, forcing `IsOnMember` true — each die to their matching must-not-flag spec.

False-positive shapes specced: active/sink read model in both forms; `c => c.EventSourceId`; `UsingParentKey` in `Children`; `UsingKey` in `Children`; `UsingKey` in `AddChild`; `parentKey:` only; no key at all; a lookalike `UsingKey` on a user's own type; a passive reducer-backed read model.

**Gate.** `DotNET.CodeAnalysis` and `DotNET.CodeAnalysis.Specs`, Debug **and** Release, `--no-incremental`: 0 errors, 0 warnings in all four. Full suite Release **385/385**. Re-run by me independently of the implementer.

**Semver: `minor`, and the reason matters.** Both are new warnings that fire on existing consumer code, and the analyzers ship inside the `Cratis.Chronicle` NuGet package, so under `TreatWarningsAsErrors` — the Cratis default in its own repos — either one breaks a consumer build on upgrade. That is a real break and deserves a release-note callout; it is the same call CHR0040–CHR0045 each took.

**A decision I did not make alone.** Narrowing `ClearWithAttribute`'s `AttributeUsage` to `Class` would make CHR0047's second half redundant, and in-repo it is safe — zero property/parameter applications exist, only class-level ones. But it is source-breaking for consumers (member-level applications become CS0592). Recommendation: ship the warning now, narrow the attribute in the next major. Not done here.

**Not verified:** neither defect was reproduced against a live kernel — both rules rest on reading the client and kernel source plus #3609's account of the replay path. The documentation pages were not render-checked, so `ChronicleClientTabs` snippet resolution and the cross-links are unconfirmed, and the snippet C# is not compiled anywhere. No full-solution build. `ClearWith` on a class never used as a `[Nested]` property is not diagnosed — undecidable across assemblies, and a deliberate gap. No downstream Cratis repo was scanned for shapes either rule would newly flag.

Issue #3601, opened 2026-08-07
Resolved: follow-up design questions on per-property PII subject ownership


woksin:
> **Resolution (2026-08-13).** The original per-property ownership seam shipped through #3599. The remaining design questions recorded here are no longer being pursued: Chronicle 16.32.0 removed `[ReleaseUnder]` and `[SubjectFrom]`, and determines ownership from projection provenance instead. Consequently, this issue does not add a release-exempt marker or a separate stored-read-model restriction.

**Concrete example.** A query-edge composed row can lift a `[PII]` value from a stored source record even though the row’s identity is not that value’s subject. If the row has no resolvable subject, release leaves genuinely encrypted bytes unchanged and ciphertext reaches the client. If it resolves a different subject, release uses the wrong subject. The original proposal attempted to declare ownership per property; the released design instead derives it from projection provenance.

Historical note: this report previously raised two follow-up questions after #3599; the resolution above supersedes them.

**Background.** A read model composed in memory at the query edge that lifts a `[PII]` value had no way to say which subject the value belongs to — the release pass inferred it from the record's identity, so the value was either released under the wrong subject or not released at all, both silently. `[ReleaseUnder(nameof(SomeProperty))]` closes that. Two questions were deliberately not answered.

**1. Should a release-exempt marker exist — and would it suppress encryption at rest?**

Release-only is a footgun on a stored read model: the value is encrypted at rest and the marker tells the release pass to leave it alone, so it reaches the client as ciphertext. Suppressing both sides overrides the kernel's own rule that storing unprotected PII is never acceptable. Scoping the marker to unstored models only is unenforceable, because nothing distinguishes them at the type level.

**2. Where should `[ReleaseUnder]` on a *stored* read model be refused?**

Not at registration. `RegisterAll()`'s throw reaches ASP.NET Core through `ApplicationStarted.Register(...)`, where a throwing callback is logged rather than fatal — so the refusal would be invisible, which is the same class of silence this whole change exists to remove. An analyzer is the obvious alternative; that is a decision, not an implementation detail.

Related: the upstream report suggested refusing *any* `[PII]` value on a record with no resolvable subject. That was **refuted** — with no subject only one outcome exists, and it is correct for a value that was never encrypted, which is the prescribed shape. Whether it is a failure depends on whether the bytes are ciphertext, which is runtime data.

woksin:
[...]

Issue #3644, opened 2026-08-10
stored read models cannot track PII owned by more than one subject


woksin:
## Problem / motivation

This issue promotes a previously recorded upstream item from a production Cratis consumer. The complete evidence, proposed seam and limits follow.

**Resolution:** Completed and released in Chronicle 16.32.0. Chronicle now derives per-property PII ownership from projection provenance and persists a `__subjects` per-property ownership map when multiple subjects are present, while retaining scalar `__subject` as the default and backward-compatible form.
**Product:** Chronicle read-model compliance and storage sinks
**Class:** Multi-subject compliance model

## Gap

`ReadModelsCompliance.Apply` encrypts a stored document under one resolved identifier and stamps one `_subject`.
The release side can now group properties by `[ReleaseUnder]`, but the apply/storage side cannot persist which
subject owns each protected property. A stored composed model therefore cannot keep multi-subject PII inside each
owner's erasure reach; the schema would need a per-property subject map across every sink.

## Requested design conversation

Define whether stored multi-subject documents are supported. If yes, introduce a versioned subject map, encrypt
each PII leaf under its declared owner, make read/release/erase operate per owner and specify migration from the
scalar `_subject`. If no, diagnose the unsupported declarations before storage.

## Corrected reasoning and limits

Partial erasure is not itself an objection: crypto-shredding already blanks protected properties while leaving the
document and non-PII fields. The material distinction is tracked versus untracked ownership. Ada has no current
surface that requires multi-subject storage and pays no code cost; the issue records a substantive compliance
boundary raised by the Cratis owner, with deliberately weak consumer evidence.

woksin:
Triage: **blocked on #3601**, and the coupling is visible in the code, not just in the prose.

[...]

PR #3706, opened 2026-08-13
Say where a value's subject comes from rather than what to do with it

woksin:
## Added

- `[SubjectFrom(nameof(property))]` declares which property holds the subject a read model value's personal data belongs to (#3601)

## Deprecated

- `[ReleaseUnder]` in favor of `[SubjectFrom]`; it still compiles and behaves identically, and now warns at the point it is applied (#3601)

### DIFF
--- a/Documentation/client-snippets/compliance/read-models/subject-from-two-subjects.md
+++ b/Documentation/client-snippets/compliance/read-models/subject-from-two-subjects.md
@@ -13,6 +13,6 @@ public record ComplianceReadModelsDuplicatePair(
     string ContactPointHash,
     string FirstPersonId,
     string SecondPersonId,
-    [ReleaseUnder(nameof(FirstPersonId))] ComplianceReadModelsPersonFullName FirstName,
-    [ReleaseUnder(nameof(SecondPersonId))] ComplianceReadModelsPersonFullName SecondName);
+    [SubjectFrom(nameof(FirstPersonId))] ComplianceReadModelsPersonFullName FirstName,
+    [SubjectFrom(nameof(SecondPersonId))] ComplianceReadModelsPersonFullName SecondName);
 ```
--- a/Documentation/client-snippets/compliance/read-models/subject-from.md
+++ b/Documentation/client-snippets/compliance/read-models/subject-from.md
@@ -15,12 +15,12 @@ public record ComplianceReadModelsRetentionSubject(
     ComplianceReadModelsPostponementComment Comment);
 
 // The row a query composes in memory. Its identity is not named Id and it carries no [Subject], so
-// it has no compliance subject of its own — [ReleaseUnder] says which subject the lifted comment
+// it has no compliance subject of its own — [SubjectFrom] says which subject the lifted comment
 // belongs to.
 public record ComplianceReadModelsRetentionDueSubject(
     string SubjectId,
     DateTimeOffset DueAt,
-    [ReleaseUnder(nameof(SubjectId))] ComplianceReadModelsPostponementComment Comment);
+    [SubjectFrom(nameof(SubjectId))] ComplianceReadModelsPostponementComment Comment);
 
 public class ComplianceReadModelsRetentionDueService(IEventStore eventStore)
 {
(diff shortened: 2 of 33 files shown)


woksin:

**This implements @einari's ruling on #3601**, which was closed with a design direction but no code:

> Not a fan of the name though, as we are consistently using Subject — this explicitly says the action Release … rather than declaratively saying where to find information. E.g. `[SubjectFrom(nameof(property))]` could be one way to go. Default behavior should then be for the properties not adorned with this to use the default for the read model, either it be the property adorned with `[Subject]` or the id of the read model.

**Why an alias and not a rename.** `ReleaseUnderAttribute` was introduced by `6d28b14f2` and first shipped in **v16.20.0**, so it is public released API — a rename would be `major`. `ReleaseUnderAttribute` now derives from `SubjectFromAttribute`, and every reader looks up `SubjectFromAttribute` once. That is deliberate over matching both names by hand: the compatibility path *is* the new path, so the two cannot drift. Adding a base class and moving `PropertyName` up is binary-compatible — memberrefs resolve through the inheritance chain.

**A refinement worth knowing.** `[Obsolete]` at the *type* level broke the build: `Cratis.Fundamentals.TypeDiscovery` enumerates every type in the assembly and emitted CS0618 inside generated code that cannot be `#pragma`'d. Moving the deprecation to the constructor fixes it and targets better — applying the attribute is the only way to consume it, so the warning lands at consumer usage sites while reflection and type-discovery code can still name the type. Verified empirically: removing the pragma from a compat spec produces `warning CS0618: 'ReleaseUnderAttribute.ReleaseUnderAttribute(string)' is obsolete`.

**The ruling's second half already held** — `ReadModelSubjectResolver.FindSubjectProperty` resolves explicit `[Subject]` on a property, then on a constructor parameter, then `Id`; and `ReadModelReleaser` releases every unclaimed key under that own-subject. The `Id` branch was already pinned, but **the explicit `[Subject]` branch was not pinned in the declaration context** — now covered by `and_an_undeclared_property_falls_back_to_an_explicit_subject`, which deliberately also has an `Id` property present so a wrong fallback to it would show.

**The two refusal types keep their names** (`ReleaseUnderPropertyNotFound`, `ReleaseUnderNotSupportedBelowReadModel`). Renaming them breaks `catch` clauses for no benefit — the ruling was about the declaration. Their messages now name `[SubjectFrom]`, pinned by two new assertions. They should be renamed in the next major alongside dropping the alias.

**Mutation evidence — 5 applied, 5 killed, 0 survivors.** I re-ran the one that matters independently: making the property lookup reject derived types (breaking `[ReleaseUnder]` while leaving `[SubjectFrom]` working) turns exactly 3 assertions RED in `and_the_declaration_uses_the_superseded_attribute` — the three property-form ones — while the 9 constructor-parameter assertions correctly pass. SPECIFIC. Restore verified by `shasum -a 256` (`db036264…d1c7ea`) independently of the trap.

**The most useful thing mutation testing did here:** that mutation initially *survived*. Not because of a weak assertion, but because the compatibility specs only exercised the record-parameter form and never reached the property-lookup line. Property-form coverage was added, after which it kills. Without that, half the compatibility guarantee would have been unverified while reading as covered.

**Checked before merge, since the implementer flagged it as unknown:** no analyzer hard-codes `ReleaseUnder`. Neither name appears anywhere in `DotNET.CodeAnalysis` (two search methods, with a control confirming 17 of 67 files match `Subject`/`PIIAttribute`), so the rename introduces no gap in an existing rule. It also means no analyzer guards this surface at all today — relevant to the open question on #3644.

**Scope.** Declaration surface only. Nothing under `Source/Kernel/**/Compliance`, no storage sink, no `WellKnownProperties`, no per-property subject map — verified by `git status`. Whether Chronicle **stores** a document whose protected values belong to different owners is the still-unruled question on #3644, and it is a persisted-format change, so nothing is being built ahead of it.

**Gate.** Debug and Release, `--no-incremental -m:1`: 0 errors, 0 warnings across 7 projects. Full client suite **2457 passed / 0 failed**. Release-declaration coverage 71 → 88 assertions. `Integration/Client` compiles clean in both configs.

**Docs** updated: `Documentation/compliance/read-models.mdx` plus its snippets, `git mv`'d to `subject-from*.md`; no stale `release-under` references remain. Deliberately left alone: `IMP-24-DESIGN-NOTE.md` mentions `[ReleaseUnder]` five times but is a historical record of design questions as they stood — rewriting it would falsify the record.

**Not verified:** the container-baked integration specs did not run, so their runtime behavior under `[SubjectFrom]` is unconfirmed (they compile). True binary compatibility was not proven against the shipped v16.20.0 assembly by compiling a consumer and running it against the new DLL — the one shape that would break is `typeof(ReleaseUnderAttribute).GetProperty("PropertyName", BindingFlags.DeclaredOnly)`, since `PropertyName` now lives on the base; judged vanishingly unlikely, not proven absent.

Issue #3711, opened 2026-08-15
Provide a real scalar clear for read model members


woksin:
## Describe the feature

Chronicle has no way to clear a **scalar** read model member. `[ClearWith<TEvent>]` clears a whole `[Nested]` single-object property, but there is no equivalent for a `string`, a `DateTimeOffset?`, a concept, or any other non-object member. This is the feature half of #3641, which was closed by shipping the diagnostic half only (CHR0047 in v16.29.0). The closing comment states the feature "deserves its own issue if wanted" — this is that issue.

## The gap today

Three declarations look like a scalar clear. None of them is one.

1. **`[SetValue<TEvent>(null)]`** — does not even compile as written: the attribute parameter is `object`, so a bare `null` raises CS8625 and anyone shipping this writes `null!`. Projection construction then guards every set-value mapping with `if (value is not null)` and no `else` (`ModelBoundProjectionBuilder.cs:358-367` and `:506-516`, `ChildrenDefinitionExtensions.cs:233-242` and `:320-334`), so the null is dropped and no mapping is emitted at all. The member is nonetheless counted as explicitly mapped (`ChildrenDefinitionExtensions.cs:343`), which suppresses CHR0024 too.

2. **`[ClearWith<TEvent>]` on a property or a parameter** — `AttributeUsage` permits `Class | Property | Parameter`, but the only consumer reads *class-level* attributes off the nested type (`NestedDefinitionExtensions.cs:135` → `:251`). Three permitted targets, one read.

3. **The fluent `.Set(m => m.Value).ToValue(null)`** — worse than inert. `SetBuilder.ToValue` maps a null to `ValueExpression.Null`, which builds the expression `$value(null)`. The kernel's `ValueExpressionResolver` matches that against its `\$value\((?<value>…)\)` regex and hands the captured text to `EventValueProviders.Value`, so the projection writes the four-character **string `"null"`** into the member. This one is not covered by CHR0047 and has no diagnostic at all.

In every case the domain shape pays for it: consumers invent type-specific sentinels — an empty string standing in for "not set" — which is not equivalent to null and leaks storage mechanics into the model.

## Proposed mechanism

- A `$null` well-known expression in both `WellKnownExpressions` classes (kernel `Concepts` and client), sitting alongside `$value(...)`, `$increment`, `$count`.
[...]

Issue #3697, opened 2026-08-13
Unique value constraints silently keep only the last RemovedWith declaration


woksin:
## Problem

`IUniqueConstraintBuilder` permits repeated `RemovedWith<TEventType>()` calls, but `UniqueConstraintBuilder` stores a single `_removedWith` value and `UniqueConstraintDefinition.RemovedWith` is singular. Each call silently overwrites the previous one.

A lifecycle with multiple terminal release facts therefore looks valid in fluent code but releases on only the last declared event. Ada reproduced this with invitation email uniqueness: failure and revocation remained reserved while provisioning, the final declaration, released.

## Requested framework seam

Either support multiple removal event types end to end for value uniqueness constraints, or reject a second `RemovedWith` declaration with a clear diagnostic/exception so consumers cannot ship a false constraint definition.

This is a framework contract issue, independent of Ada policy.

## Acceptance

- A focused builder spec proves repeated removal declarations cannot silently overwrite.
- If multiple releases are supported, the definition/contract/storage/index updater/reindex paths preserve and honor every declared event type.
- A kernel spec reserves one value, releases it independently with each configured terminal event type, and permits the value again.
- Existing single-removal constraints remain source/binary compatible.

woksin:
Fixed in #3713, released in **v16.34.0**. A unique constraint can now declare several removal events and every one of them frees the value; the attribute form honours `[RemoveConstraint]` on more than one event type; and constraints sharing a name carry every removal declared across them. Pinned end to end, including a wire-level spec for a payload from a client predating the plural form.

Issue #3690, opened 2026-08-12
Encryption-certificate rotation: decide the mechanism, then implement it and document the backup/restore ordering


woksin:
Carried forward from #3607 (asks 3 and 4, and acceptance criterion 6), which is closed as split.

`EncryptionCertificate` is one path and one password (`Source/Kernel/Core/Configuration/EncryptionCertificate.cs:14,19`). The single protection call is `ProtectKeysWithCertificate(certificate)` (`Source/Kernel/Server/Authentication/ServiceCollectionExtensions.cs:59`); `UnprotectKeysWithAnyCertificate` appears nowhere in the repository. All three consumers — Data Protection, OpenIddict encryption/signing (`OpenIddictServiceCollectionExtensions.cs:97`) and `IEncryption` (`Core/Security/Encryption.cs:70`) — load that same file, so replacing it makes every previously protected value undecryptable. There is no rollover path today.

@einari — one line unblocks this:

> Should `EncryptionCertificate` accept an ordered list — active first, previous kept decryption-only via `UnprotectKeysWithAnyCertificate` — or is Chronicle's supported rotation a documented re-encryption window with downtime?

Then the work is: a rotation spec (protect with A, add B retaining A, switch writes to B, restore from backup, read both generations); a bounded diagnostic that names the required key id without exposing protected material; and a backup/restore ordering document covering Mongo data, the database-backed Data Protection key ring, the encryption certificate and the optional compliance-key store. The ordering depends on the answer above, which is why it lives in this issue rather than the documentation one.

woksin:
Fixed in #3713, released in **v16.34.0**. The mechanism chosen is an additive ordered key ring: the active certificate protects new writes while previous ones stay for decryption, with ciphertext carrying the key id so a value needing a retired certificate is reported as unreadable rather than as corrupt. Adds `GET /diagnostics/encryption-certificates`, boot-time ring logging, and the backup/restore ordering — including that a certificate must stay in the backup set longer than it stays in the ring.

Issue #3643, opened 2026-08-10
[Resolved in Chronicle v16.34.0] Reactor delivery identity for retry-aware idempotency


woksin:
## Resolution

**Resolved in Chronicle v16.34.0 by #3713.** `ReactorDelivery` is available as a handler parameter and provides a stable per-delivery idempotency key across recovery re-delivery. It is not an exactly-once-delivery guarantee.

This resolves the requested delivery-identity seam. A Chronicle-managed durable receipt remains deliberately out of scope: it would introduce retained state and a retention policy, so it requires a separate issue and maintainer ruling if wanted.

## Problem / motivation

This issue promotes a previously recorded upstream item from a production Cratis consumer. The complete evidence, proposed seam and limits follow.

**Status:** Resolved in Chronicle v16.34.0 by [#3713](https://github.com/Cratis/Chronicle/pull/3713)
**Product:** Chronicle reactors
**Class:** Side-effect idempotency seam

## Gap and Ada cost

`[OnceOnly]` correctly suppresses deliberate replay, not failed-partition retry. A handler that performs an
external side effect and then fails can therefore repeat that effect on recovery. Ada must make each integration
self-idempotent, usually with provider-specific keys or log-authoritative guards, and explicitly teach that
`[OnceOnly]` is not exactly-once delivery.

## Requested seam

Expose stable delivery/attempt identity and a durable consumer-controlled completion/idempotency receipt around
external effects, or document and provide a standard pattern that composes with provider idempotency keys. Recovery
must still re-run a handler that failed before performing its effect.

## Explicit non-goal and limits

Do not make `[OnceOnly]` skip retries wholesale; that would turn pre-effect failures into permanent loss. Exactly
once across an arbitrary external system is impossible without cooperation. This is a design conversation backed
by repeated Ada integration design cost, not a framework correctness defect.

woksin:
Investigated against current `origin/main`. **The documentation half already shipped**; the seam does not exist and needs a ruling from you, @einari.

**Shipped in v16.13.2** (`9b2a81a65`): `Documentation/reactors/once-only.mdx` now carries a "What it does and does not guarantee" section stating plainly that this is *not* an exactly-once delivery guarantee, and that recovering a failed partition re-delivers the event as an ordinary observation, so a `[OnceOnly]` handler runs again. The teaching this issue asked for exists.

[...]

PR #3713, opened 2026-08-15
Close six upstream consumer reports: delivery identity, certificate rotation, scalar clear, key erasure, constraint removals, first-append concurrency

woksin:
## Added

- Reactors can take a `ReactorDelivery` parameter on a handler method to get a stable identity for that delivery of that event — the same value when a failed partition is recovered and the event is delivered again, and different for every distinct delivery. Use `delivery.Id` as the key for your own record of completed side effects (#3643)
- Read model members can be cleared. `[ClearWith<TEvent>]` on a scalar property or record parameter, `[SetValue<TEvent>(null)]`, and the fluent `.Clear(m => m.Property)` all write the member back to no value every time the event is observed, replay included — at the root, on a child item, and on a nested object's members (#3711)
- The fluent projection builder gained `.Clear(m => m.Property)`, available everywhere `.Set(...)` is. It is the recommended way to clear; `.Set(...).ToValue(null)` means the same thing and keeps working (#3711)
- The projection declaration language gains a `clear` statement. `clear <property>` takes a value away, alongside `increment`, `decrement` and `count`, and accepts a property path so `clear owner.note` clears a member of a nested object. `property = null` still compiles to exactly the same mapping (#3711)
- `CHR0048` reports a clear declared for a member that cannot hold null — for `[ClearWith]`, a null `[SetValue]`, `.Clear(...)` and `.Set(...).ToValue(null)` alike — as a warning. It is scheduled to become an error in the next major; building the projection refuses the declaration outright, so an unaddressed warning fails at startup rather than at build time (#3711)
- A unique constraint can be released by more than one event. Declare `RemovedWith<T>()` once for each way the lifecycle ends — accepted, revoked, expired — and every one of them frees the value on its own. The same holds for the attribute form: several event types may carry `[RemoveConstraint]` for one constraint name (#3697)
- Encryption certificates can be rotated without downtime: `encryptionCertificate.previous` holds an ordered list of previously active certificates, kept for decryption while the active one protects everything new (#3690)
- `GET /diagnostics/encryption-certificates` reports the certificate ring, which certificate is active, and how many stored Data Protection keys still depend on each — including keys that depend on a certificate no longer in the ring (#3690)
- The encryption-certificate ring is logged at every boot, one line per certificate, with a warning for any that has expired (#3690)
- `IEventStore.PII.AllowNewEncryptionKeyFor(subject)` authorizes a fresh encryption key for a subject whose key was erased, so a later lawful lifecycle can protect their data again under an independent key (#3600)
- Encryption key stores record erasures beside the keys — a revision floor, the fingerprints of the destroyed key material, and whether a new lifecycle is authorized — so "erased" is a state Chronicle can tell apart from "never provisioned" (#3600)
- Erasing a subject and authorizing a new encryption key for one are recorded in Chronicle's log — the completion, the namespace, and every event store reached, with the subject as a one-way binding rather than by name and never any key material. An erasure that did not reach every event store is logged as an error saying so (#3600)
- `ConcurrencyOptions.CheckFirstAppendIntoAScope` opts in to checking the first append into a concurrency scope. With it off, the default, the first append into a narrowed scope is unchecked as before; with it on, the kernel rejects the append if a matching event appeared between resolving the scope and the append arriving (#3635)
- `ConcurrencyScopeBuilder.ExpectingNoMatchingEvent()` asks for that check on a single append without changing the application-wide default (#3635)
- `IAppendResult.ConcurrencyCheckPerformed` reports whether the concurrency check actually ran, so a skipped check can be told from a passing one in a test rather than only in the server log. It has a default implementation, so existing implementers are unaffected (#3635)

## Changed

- **Re-run the erasure for every subject erased before upgrading.** A pre-upgrade erasure left the same absence as "never provisioned" — that is the defect — so nothing can recognize it retroactively and those subjects stay resurrectable until erased again. The call is idempotent and costs one round trip per subject (#3600)
- **Upgrade every silo together while an erasure is in force.** The erasure fence is enforced by the key store, so a silo that predates it provisions straight over one (#3600)
- `DeleteEncryptionKeyFor` erases across every event store in its namespace instead of the one it was resolved from, which is exactly how far a subscription can copy a subject's key; it never crosses a namespace (#3600)
- Appending a `[PII]` value for an erased subject fails with `EncryptionKeyErased` instead of silently minting a new key; authorize a new key first when the person has a lawful basis to be protected again (#3600)
- An erasure records itself in every event store before destroying key material in any of them, and reports every store it could not reach as `EncryptionKeyErasureIncomplete` (#3600)
- Values encrypted by Chronicle carry the key id of the certificate that protected them, so a value needing a retired certificate is reported as unreadable and names the key id required, instead of failing as if it were corrupt. Values written before this are still read (#3690)
- OpenIddict and ASP.NET Core Data Protection are configured with the whole certificate ring, so tokens and keys written under a previous certificate keep working across a rotation (#3690)
- A unique event type constraint violation no longer carries the event source id, in the default message or in the violation details. It reaches callers as a validation result, and an event source id is the stream identity and compliance subject and cannot be encrypted. The event type and the conflicting sequence number are unchanged, and a caller that needs the id already holds it — it is the event source it just attempted to append for (#3632)
- Reading `RemovedWith` on `UniqueConstraintDefinition` or `UniqueEventTypeConstraintDefinition` returns a collection rather than a single event type. Constructing them is unaffected — the previous signatures are kept as obsolete constructors — but code compiled against the old property will need recompiling (#3697)
- `[SetValue<TEvent>]` accepts a nullable value, so a clear is written as a plain `null` instead of `null!` (#3711)
- `[ClearWith<TEvent>]` on the property holding a `[Nested]` object clears that object, the same as declaring it on the nested type — useful when the nested type is shared and should not know the owner's events (#3711)
- `CHR0024` no longer treats a null `[SetValue]` as a mapping source, so a member that one event clears and nothing ever populates is reported instead of silently accepted (#3711)
- A clear round-trips as a clear. Generating a declaration from a projection definition writes `clear <property>` whichever spelling it was authored in, and generated code renders `.Clear(...)` in fluent form and `[ClearWith<TEvent>]` in model-bound form (#3711)
- `ReactorScenario` gives every event it feeds a reactor its own event sequence number instead of `Unavailable`, so a reactor keyed on the delivery identity behaves in a scenario the way it behaves in production (#3643)

## Fixed

- An erasure no longer leaves the subject's key readable in the other event stores a subscription had copied it into (#3600)
- A forwarded event no longer restores an erased subject's original key material, which used to make already crypto-shredded personal data readable again (#3600)
- A composed encryption key store no longer heals a surviving copy of an erased key back into the store it was erased from, and no longer serves that copy (#3600)
- Repeated `RemovedWith` declarations no longer overwrite each other. A constraint declaring several terminal events previously released on only the last one declared; the others registered and did nothing, so the value stayed claimed forever while every append succeeded (#3697)
- Constraints declared with `[RemoveConstraint]` on more than one event type honor all of them, instead of the first one found (#3697)
- Unique event type constraints declared under one shared name carry every removal event declared across those declarations (#3697)
- The fluent `.Set(...).ToValue(null)` wrote the literal string `"null"` into the member instead of clearing it (#3711)
- A `[SetValue<TEvent>(null)]` was discarded when the projection was built, leaving the member with whatever it last held, across replay (#3711)
- A `[ClearWith<TEvent>]` on a property or record parameter bound to nothing at all (#3711)
- A property path through a nullable nested object never resolved during projection-declaration validation, so any dotted mapping onto a nested read model member was rejected as "property not found" (#3711)
- An unrecognized projection mapping silently skipped property validation and then failed as an unhandled exception; it is now reported as a compiler error on the declaration, with a line and column (#3711)
- An encryption certificate configured with a path that holds no file no longer starts silently: it used to leave Data Protection writing its keys unencrypted and OpenIddict reporting that no certificate was configured when one was. It now stops startup naming the path (#3690)

## Removed

- `CHR0047` is retired: both declarations it reported as inert now work. Remove any suppression of it — the identifier stays reserved and will never be reused (#3711)

## Security

- Crypto-shredding is irreversible within a Chronicle cluster: an erased key cannot be re-provisioned, copied between event stores, healed between composed stores, or restored by writing the same key material back — at any revision, and including after a new lifecycle is authorized (#3600)
- A key store that cannot record an erasure fails loudly when one is attempted, instead of reporting a reversible erasure as complete (#3600)
- A misconfigured encryption-certificate ring fails closed rather than degrading quietly: a missing file, a previous entry without a path, previous certificates without an active one, the same key pair in two positions, and a certificate without a private key are all rejected at startup (#3690)
- Encryption-certificate diagnostics and failure messages name certificate thumbprints and paths only — never protected values or key material (#3690)


woksin:
Reviewer context — deliberately kept out of the PR body.

## Why this is `minor` and not `major`

It started out `major`. Every breaking surface was removed deliberately, and each removal is spec'd:

| Was breaking | Resolution |
|---|---|
| First append into a narrowed scope started being **checked**, rejecting appends that succeed today | Now **opt-in** via `ConcurrencyOptions.CheckFirstAppendIntoAScope`, default `false` in a single named constant so the next major flips one value. Two layered specs pin the default — one on the behaviour, one on the constant |
| `IAppendResult.ConcurrencyCheckPerformed` abstract member | Default interface implementation returning `false` — conservative direction, under-reports and never over-reports |
| `AppendResult` / `AppendManyResult` class → record | Reverted. The agent **packed `Testing.csproj` and read the nupkg** rather than trusting `IsPackable`: `Cratis.Chronicle.Core.dll` (3,009,024 bytes) ships embedded via `PrivateAssets=all` |
| `OptimisticConcurrencyStrategy` gained an optional constructor parameter | Single-argument overload kept — default arguments are baked in at the call site, so already-compiled callers would have broken |
| Four public positional records changed `RemovedWith` from `EventTypeId?` to a collection | `[Obsolete]` compatibility constructors on all four, verified by dumping the constructor tables via `MetadataLoadContext` and diffing against `origin/main` parameter-for-parameter |
| `CHR0048` shipped as a build **error** | Warning, matching the CHR0047 precedent, with the escalation to error announced in its docs |
| `IPIIManager.AllowNewEncryptionKeyFor` abstract member | Default implementation throwing `AllowNewEncryptionKeyNotSupported` |

**Two residual breaks, both judged and neither hidden:**

1. Reading `RemovedWith` on those four records now returns a collection, so pre-compiled code that *reads* the property breaks. The fix would be renaming the plural and keeping a lossy scalar accessor — which would answer "one removal event" and silently hide the rest, reintroducing the exact defect #3697 is about. Construction is the plausible external use and is shimmed; reading a definition's removal event externally is not, since the supported way to observe constraints is the builder and the attributes.
2. `PIIManager`'s constructor gained a parameter. It is an Orleans grain, activated by the runtime and only reached via `grainFactory.GetGrain<IPIIManager>(...)`; constructing it directly yields a grain with no Orleans context whose `GetPrimaryKey` fails. A two-argument overload would have to leave `IStorage` null, and erasure would then quietly not fan out across the namespace — the silent degradation this issue exists to remove.

Two interface additions deliberately have **no** default: the kernel `IPIIManager` is an Orleans grain interface and `ICompliance` is protobuf-net codegen input. In both, a default body executes on the caller's side and never reaches the implementation, so it would be a silent no-op across a wire boundary.

## Verification

Run with CI's own filter (`FullyQualifiedName!~Cratis.Chronicle.Integration&FullyQualifiedName!~Cratis.Chronicle.Kernel.Integration`), which is how `dotnet-build.yml` gates:

- `dotnet build Chronicle.slnx -c Release` — 50 projects, **0 errors, 0 warnings**
- **8,226 passed, 0 failed, 1 skipped** across 17 assemblies (the skip is the pre-existing Orleans TestKit one)
- Protobuf schemas compile with a real `protoc` — 23 modules

A first full-solution run locally showed failures in the integration and MongoDB assemblies. Those are **not** regressions: each passes when run alone (MongoDB specs 728/728 on this branch, 718/718 on pristine `origin/main` as a control), and CI runs those tiers separately with dedicated Docker orchestration. The machine was hosting 14 containers and 41.8GB of images from parallel worktrees.

Behaviour-level evidence rather than build-level, per finding: multi-store erasure ran `for_PIIManager` **27/27 against a real silo and real MongoDB**; the reactor delivery identity has an acceptance spec (charge → receipt → throw → recover → charged once, handler ran twice) **with a control** proving the same two deliveries through `[OnceOnly]` charge twice; the constraint work was mutation-checked by reintroducing the overwrite in the builder and separately in the kernel; the scalar clear was falsified by removing the line that emits the mapping.

## Deliberately not in this PR

**The proto generator fix.** `Tools/ProtoGenerator` validates assembly-wide `ReservedProtoFields` types against each package's schema, so it fails 22 of 23 packages — **and exits 0**, which is why it was never noticed. Fixing the scoping was implemented and then **reverted**, because it unmasks a second defect: the regenerated `eventsequences.proto` is invalid protobuf (`EventType` defined twice — enum values are siblings of their type). Measured three ways with a real `protoc`: pristine `main` compiles (23 modules), this branch compiles (23), the regenerated output does not. Since `generate-protos` is a `needs:` dependency of `publish.yml` feeding the Kotlin/TypeScript/Elixir client publishes, fixing the scoping alone would make publish *succeed* at producing a schema those clients cannot compile. Full write-up on #3712, including that the proper fix requires renaming enum values — a breaking change to the generated client APIs and its own decision. The four hand-applied `.proto` fields in this PR are marked inline and reference #3712.

## Follow-ups the work surfaced

- `#3712` — proto generator: reserved-fields scoping, exit code, and the duplicate-symbol defect
- `#3711` — filed for the scalar clear feature, which #3641 left unowned after shipping only its diagnostic
- `#3675` — `XUnit.Integration.Specs` `for_the_embedded_kernel_closure` fails on an undeclared package; pre-existing, reproduced on pristine `main`, and that project is one of four missing from `Chronicle.slnx`
- `AllowNewEncryptionKeyFor` has no authorization gate or audit log; docs tell operators to keep their own record. If Chronicle should gate or log it, that is a product call
- SQL stores erasure fingerprints comma-joined in an unbounded string column; a `maxLength` audit would be cheap insurance against silent truncation
- The fluent `.ToValue(null)` path has no nullability gate, so a clear on a non-nullable member there neither warns nor throws. A fluent sibling analyzer would close it
- Encryption-certificate rotation has no Aspire builder method for the previous list; the `WithBindMount` + `WithEnvironment` escape hatch is documented instead

woksin:
## For review — the public API surface, in full

Computed by extracting every `public`/`protected` declaration from each of the 140 changed production files at `origin/main` and at this branch, and diffing the sets. **146 declarations added, 17 removed or changed.** The additions are listed in the release notes; the 17 below are the ones worth your attention, because each is either a signature that moved or a type that went away.

### Signatures that changed in place

| Declaration | Change | Why it is not a break |
|---|---|---|
| `record UniqueConstraintDefinition(...)` ×2 assemblies | `RemovedWith`: `EventTypeId?` → `IEnumerable<EventTypeId>` | `[Obsolete]` compatibility constructor carries the old positional signature, verified by dumping constructor tables via `MetadataLoadContext` and diffing against `origin/main`. ⚠️ **Reading** the property still breaks — see below |
| `record UniqueEventTypeConstraintDefinition(...)` ×2 assemblies | same | same |
| `class OptimisticConcurrencyStrategy(IEventSequence)` | gained an optional `ConcurrencyOptions?` | Single-argument overload kept; default arguments are baked in at the call site, so removing it would have broken already-compiled callers |
| `class PIIManager(IEncryptionKeyStorage, IEncryptionKeyCacheClient)` | gained `IStorage` | ⚠️ **Not shimmed.** Orleans grain — activated by the runtime, only reached via `grainFactory.GetGrain<IPIIManager>(...)`; direct construction yields a grain with no context whose `GetPrimaryKey` fails. A 2-arg overload would leave `IStorage` null and erasure would quietly not fan out across the namespace, which is the failure this PR exists to close |
| `class Encryption(IOptions<ChronicleOptions>)` | now takes the certificate ring | Internal composition; resolved through DI |
| `record EncryptionKey(byte[] Public, byte[] Private)` | gained a body with a computed `Fingerprint` | Positional constructor unchanged — purely additive |
| `SetValueAttribute<TEvent>(object value)` | `object` → `object?` | Annotation-only; binary and source compatible. This is what lets a clear be written as plain `null` rather than `null!` |
| `EventValueProviders.Value(string)` | `string` → `string?` | Annotation-only |
| `AddOpenIddictIfEnabled(...)` | gained a parameter | Kernel is not packed on its own |
| `ConcurrencyScope.IsIncomplete` / `ShouldBeValidated`, `EventSequenceNumber.IsActualValue` | bodies widened for the new sentinel | Same signature, same meaning for every previously-representable value |
| `class InertClearDeclarationAnalyzer` | **deleted** | CHR0047 retired; both shapes it reported as inert now work. The id stays reserved and its docs page remains as a retired page, so an existing suppression cannot silently carry onto a new rule |

### The one residual break, stated plainly

Reading `RemovedWith` on those four records now returns a collection, so **pre-compiled code that reads the property breaks**. The fix would be renaming the plural and keeping a lossy scalar `RemovedWith` — which would answer "one removal event" and silently hide the rest, reintroducing exactly the defect #3697 fixes. Construction is the plausible external use and is shimmed; reading a definition's removal event externally is not, since the supported way to observe constraints is the builder and the attributes. **If you would rather carry the accessor anyway, say so and it is a small change.**

Two interface additions deliberately have **no** default implementation: the kernel `IPIIManager` is an Orleans grain interface and `ICompliance` is protobuf-net codegen input. In both, a default body runs on the caller's side and never reaches the implementation, so it would be a silent no-op across a wire boundary. Every other new interface member has a default.

## Documentation — 43 pages

**New:** `reactors/delivery-identity.mdx`, `projections/model-bound/clearing.mdx`, `code-analysis/CHR0048.mdx`, `hosting/encryption-certificate.md`, `compliance/key-lifecycle.md`, `compliance/erasing-a-subject.md`, plus 14 compiled client snippets.

**Rewritten, and worth reading rather than skimming — each replaces guidance that is now wrong:**

- `events/concurrency.mdx` — previously documented the skipped first append as intended behaviour, with a decision diagram. Now documents the opt-in check, both levels of opt-in, the planned default flip, and what a mixed-version deployment does.
- `code-analysis/CHR0047.mdx` — rewritten as a retired page rather than deleted.
- `hosting/production.md` + `hosting/encryption-certificate.md` — the rotation procedure and the backup/restore ordering, including the rule that a certificate must stay in the backup set longer than it stays in the ring.
- `compliance/key-lifecycle.md` — the erasure fence, its threat model, its limits, and the upgrade story including re-running pre-upgrade erasures.
- `reactors/once-only.mdx` / `replay.mdx` — now point at the delivery identity for the recovery case they do not cover.

All client snippets compile (`validate-client-snippets.py`, 0 warnings) and `markdownlint-cli2` is clean across 838 files.

## The two operational instructions

These are in the release notes as `## Changed` bullets because they are things an operator must *do*, not things to be aware of:

1. **Upgrade the kernel before any client.** A client from this release validates its full service schema at connect, so it throws `IncompatibleServerException` against an older kernel — verified experimentally, not inferred: webhook integration specs failed 15/24 against a stale kernel image and passed 24/24 once rebuilt.
2. **Re-run the erasure for every subject erased before upgrading.** A pre-upgrade erasure left the same absence as "never provisioned" — that is the defect — so nothing can recognise it retroactively.

## What I would most like a second opinion on

- **The non-nullable clear ruling** — clearing a member that cannot hold null is refused (`CHR0048` + a builder throw) rather than defined as the type default, on the grounds that writing `""` or `0` is a different fact the read model cannot distinguish from a real value.
- **The concurrency default staying off.** The checked first append is correct and is what #3635 asked for; it ships opt-in so this release stays `minor`. The default flip is announced in the docs for the next major.
- **`AllowNewEncryptionKeyFor` has no authorization gate or audit log.** Docs tell operators to keep their own record. Whether Chronicle should gate or log it is a product call.
- **The erasure record stores the raw identifier.** Argued in the docs as no more revealing than the subject every appended event already carries in the clear; a pseudonymous binding could be layered later.

woksin:
**CI is fully green — 44 checks pass, 0 fail**, including all **25 integration shards** (client, clustering and kernel tiers, against real containers), plus `check-compatibility`, `kotlin-build` and `typescript-build`.

That last group matters for this PR specifically, because it carries wire changes: two new proto fields and a pluralised `RemovedWith`. The Kotlin and TypeScript client builds passing is the signal that the non-.NET clients still generate and compile against the changed schemas.

Worth recording for whoever reviews: a first full-solution run **locally** showed failures in the integration and MongoDB assemblies. Those were environment, not regressions — each assembly passes when run alone (MongoDB specs 728/728 on this branch; 718/718 on pristine `origin/main` as a control), and CI's own gate excludes the integration tiers from the unit step and runs them separately with dedicated Docker orchestration. The machine was hosting 14 containers and 41.8 GB of images from parallel worktrees at the time. CI, with clean infrastructure, is the authority — and it is green.

Not merging on green: this is held for @einari's review of the API surface and docs.

woksin:
## Wire break found in review and fixed — force-free, and now measured

@woksin caught that `Constraint.RemovedWith` had been moved off proto field 3 (reserved) onto a new field 6. That is a **silent** break, and worse, it sits on the upgrade path this release documents.

**The failure path:** an older client writes `RemovedWith` on field 3 → the upgraded kernel has field 3 reserved, so protobuf-net skips it → `ToRemovedWith()` reads the empty field 6 → **the constraint registers with no removal event**. The unique value is claimed and never released, so a value that should become re-usable after its terminal event stays claimed forever. `CompatibilityValidator` compares services and RPC method signatures only, never message field shapes, so the client connects cleanly and loses its removal events without a word. And the #3600 note tells operators to *upgrade the kernel before any client*, which is exactly that window.

**The fix is smaller than the break.** `RemovedWith` keeps **field 3** and only widens its type, `string?` → `IList<string>`. No reservation, no field 6, no legacy property, no fallback branch. The entire proto diff against `origin/main` is now one line:

```diff
-   string RemovedWith = 3;
+   repeated string RemovedWith = 3;
```

In proto3 a singular field and a `repeated` field of the same non-packed type share a wire representation. Measured with real `protoc`-generated types and actual bytes, then reproduced independently through protobuf-net:

| Case | Result |
|---|---|
| Old client writes scalar → new kernel reads repeated | `['ItemRemoved']` ✅ |
| New writes one → old kernel reads scalar | `'ItemRemoved'` ✅ |
| New writes three → old kernel reads scalar | `'Expired'` — degraded to one, **never empty** |
| Single-value encoding vs. before | **byte-identical** |

Existing single-removal-event payloads therefore serialise to exactly the same bytes as they always did.

## The spec that would have caught it, and the one that would not

Three converter specs were written for this, run against the **broken** contract as a check — and **all three passed**. They construct a `Constraint` in memory and call `ToChronicle()`, which pins what the converter does with a field but says nothing about which number it lives on. Necessary, not sufficient.

`and_the_payload_came_from_a_client_predating_the_plural_form` is the one that holds the line: it declares a `[ProtoContract]` stand-in with the pre-plural shape, serialises it to real bytes, deserialises those bytes as the current `Constraint`, and converts. Re-running the mutation, that spec **fails on 2 facts** while the other 13 stay green — verified independently by moving `RemovedWith` back off field 3 and watching exactly those two go red.

## Audit of the rest of the wave

Every `[ProtoMember]` in all six changed contract files was compared between `origin/main` and this branch:

| Contract | Field | Change |
|---|---|---|
| `AllowNewEncryptionKeyRequest` | 1, 2, 3 | added (new message) |
| `AppendResponse` / `AppendManyResponse` | 6 | added |
| `ConcurrencyScope` | 7 | added |
| `Constraint` | 3 | **widened in place** — the fix above |

Everything else is additive at fresh field numbers, which degrades safely in both directions. `Constraint` field 3 was the only wire change in the wave, and it is now non-breaking.

## One pre-existing instance found, filed separately

The same audit turned up the identical shape already on `main`: `UniqueEventTypeConstraintDefinition` has `[ReservedProtoFields(1)]` with `EventTypeIds` moved to field 2. An older client's field-1 value deserialises to **null**, and `ToChronicle` dereferences it unguarded — an NRE that aborts constraint registration for the whole event store. Loud rather than silent, but on the same upgrade path. Not touched here, because kernels are already deployed with `reserved 1` and moving it back is a three-population compatibility decision. → **#3715**

## Re-verification after the fix

- `dotnet build Chronicle.slnx -c Release` — 50 projects, **0 errors, 0 warnings**
- **17 spec assemblies green** under CI's own filter, 0 failing
- `protoc` over the full schema set — **23 modules**, exit 0
- Working tree clean (a generated Workbench proxy re-emitted with a fresh timestamp header; content identical, reverted)

woksin:
**Removed the "upgrade the kernel before any client" bullet from the release notes** — @woksin is right that it restates the standing compatibility contract rather than describing anything this release changes.

`Source/Clients/Connections/CompatibilityValidator.cs` validates *"each service the client expects"* against the server's schema. So:

- **client behind the kernel** — expects fewer methods than the server offers → **passes**
- **client ahead of the kernel** — expects a method the server lacks → **refused**

That is the existing policy: a client may lag on minors, never lead. Adding an RPC in a minor is the ordinary case that policy is designed for, so it is true of every such release and does not belong in these notes. Leaving it in would also frame a routine release as carrying a special upgrade constraint.

Replaced with the constraint that **is** specific to this release: **upgrade every silo together while an erasure is in force**, because the fence is enforced by the key store and a silo predating it provisions straight over one. The re-run-your-pre-upgrade-erasures bullet stays, for the same reason.

Two things this sharpens, both worth recording:

1. **The right framing of the wire fix is the opposite of a new constraint.** `Constraint.RemovedWith` moving off field 3 would have *broken* the standing "client may be behind" guarantee — an older client's removal event silently dropped. Widening field 3 in place restores it. This release does not add an upgrade constraint; it removes a violation of one.
2. **#3715 is a live violation of that same guarantee, on `main` today.** `UniqueEventTypeConstraintDefinition` reserved field 1 and moved `EventTypeIds` to field 2, so an older client's value deserializes to null and `ToChronicle` dereferences it — a client that is *legitimately behind* crashes constraint registration for the whole event store. Under the stated policy that is a defect, not an expected upgrade cost.

Related gap: I could not find the "client may be behind on minors, never ahead" rule stated anywhere user-facing in `Documentation/`. It is enforced in code and assumed in review, but a consumer has no way to read it. Worth documenting on its own — happy to open a follow-up if wanted.

Documentation/projections/model-bound/clearing.mdx
```diff
@@ -0,0 +1,79 @@
...
+## In the Declarative Language
+
+A clear is the `null` literal in a mapping:
+
+```text
+projection Notes => MbClearingProjectNotes
+    from MbClearingProjectNoted
+        note = note
+    from MbClearingProjectNoteCleared
+        note = null
```
einari: I would love to see something that does not involve the concept of null. 
Clearing is a much better term. So something like

```text
projection Notes => MbClearingProjectNotes
    from MbClearingProjectNoteCleared
        clear note
```

Documentation/projections/model-bound/clearing.mdx
```diff
@@ -0,0 +1,79 @@
...
+`ToValue(null)` builds the same clear:
+
+```csharp
+public class MbClearingProjectNotesProjection : IProjectionFor<MbClearingProjectNotes>
+{
+    public void Define(IProjectionBuilderFor<MbClearingProjectNotes> builder) => builder
+        .From<MbClearingProjectNoted>(_ => _
+            .Set(m => m.Note).To(e => e.Note))
+        .From<MbClearingProjectNoteCleared>(_ => _
+            .Set(m => m.Note).ToValue(null));
```
einari: Would like to see a formalized Clear, rather than .Set().ToValue(): 

```csharp
.From<MbClearingProjectNodeCleared(_ =>
   .Clear(m => m.Note));
```

einari (dismissed):

Documentation/projections/model-bound/clearing.mdx
woksin: Agreed, and this one is being done now — `.Clear(m => m.Note)` is going in as a formal builder method at root, child and nested level, over the same `$null` expression the attribute form already resolves. So it is a new spelling on machinery that exists rather than a second mechanism.

It also closes a gap worth naming: the fluent path currently has **no** nullability gate at all, so `.Set(...).ToValue(null)` on a non-nullable member neither warns nor throws today. A typed `.Clear(...)` can carry that constraint in its signature, which turns the non-nullable case into a compile error instead of a diagnostic.

`.ToValue(null)` will keep working — it was writing the literal string `"null"` into the member, and that fix stays — but the docs will lead with `.Clear(...)` as the way to express it.

Documentation/projections/model-bound/clearing.mdx
woksin: I agree with the principle, and `clear note` reads far better than `note = null` — a mapping that assigns a value and a statement that removes one are different acts, and spelling both with `=` hides that.

⚠️ **It cannot be done on this PR.** The projection declaration language is parsed in the **Cratis/Screenplay** repository — `Source/DotNET/Screenplay/Parsing/ProjectionParser.cs` — and Chronicle consumes it as a package (`Cratis.Screenplay` 2.1.0). A `clear <property>` statement needs a grammar change there first, then a Screenplay release, then a package bump here before the syntax can be accepted or generated.

So the chain is: Screenplay grammar + parser → Screenplay release → Chronicle bump → the declaration language and all three code generators learn `clear`.

**Your call on how to sequence it:**

1. **Follow up** — merge this with `note = null` as the declaration-language spelling, and land `clear` as its own change once Screenplay ships. `note = null` keeps working either way, so `clear` arrives additively rather than as a breaking rename.
2. **Block** — hold #3713 until the Screenplay change is released and bumped, so the feature lands with the spelling you want and consumers never see the `null` form documented as the way.

I lean towards (1): the six reports in this PR are unrelated to the spelling, and holding them behind a cross-repo release chain delays a set of security and correctness fixes — the multi-store key erasure most of all. But if you would rather the `null` form was never the documented syntax, (2) is the cleaner story and I will queue the Screenplay work.

Either way the fluent `.Clear(...)` from your other comment is going in now, so the two spellings will not be inconsistent for long.

woksin: Ruled — `clear` is the rule, not an option, and both halves are being done.

**The language half is in progress in Screenplay.** It fits the existing keyword-mapping family exactly: `increment`, `decrement`, `count`, `add … by`, `subtract … by` already parse to their own syntax nodes, so `clear <property>` becomes a `ClearMappingSyntax` beside `IncrementMappingSyntax` — parser, printer and syntax walker, with `clear owner.note` working for dotted paths too. A distinct node rather than sugar over the null literal, so the printer round-trips `clear note` back to `clear note` instead of silently rewriting it to `note = null`. PR to follow on Cratis/Screenplay.

`note = null` keeps parsing exactly as it does today — a released Chronicle compiles it into a clear, so removing it would be a breaking language change. `clear` arrives additively and becomes the documented spelling.

**Sequencing.** The chain is Screenplay merge → Screenplay release → Chronicle bumps `Cratis.Screenplay` → Chronicle's `DeclarativeCodeGenerator` learns to emit `clear`. That is more than #3713 can carry, so my proposal stands: merge #3713 with `note = null` still documented, and land the `clear` spelling as its own follow-up once Screenplay ships. If you would rather the `null` form were never documented as the way, say so and I will hold #3713 until the chain completes — but that delays the multi-store key erasure and the certificate rotation behind a spelling change.

Your other comment — the formal `.Clear(m => m.Note)` in the fluent API — is being implemented on this branch now and will land in #3713 itself, since it needs no Screenplay change.

woksin:
**Both review comments are now implemented, and `clear` is the spelling end to end.** @einari — the sequencing question I raised is moot; no follow-up release is needed.

- **The fluent half** landed here: `.Clear(m => m.Property)` at root, child and nested, from one addition to `IReadModelPropertiesBuilder` (which all four surfaces compose over). It carries a throwing default implementation, so adding it breaks no existing implementor.
- **The language half** landed in [Cratis/Screenplay#116](https://github.com/Cratis/Screenplay/pull/116), released as 3.1.0, and this branch now bumps to it and teaches Chronicle to read *and emit* `clear <property>`.

**On the bump.** 2.1.0 → 3.1.0 crosses Screenplay's 3.0.0, whose breaking changes are all in the reaction/trigger surface. Chronicle consumes only the projection and capture syntax and holds no reference to any of it, so the full solution built clean on the bump alone, before a line of code changed. Worth knowing for anyone auditing dependencies: `Cratis.Screenplay` is not a declared dependency of `Cratis.Chronicle.Testing` — `PrivateAssets=all` suppresses it — but its DLL ships embedded in `lib/`, so the embedded copy moves from 2.1.0 to 3.1.0.

## Two pre-existing bugs found on the way, both fixed

1. **A property path through a *nullable* nested object never resolved.** `ProjectionValidator.TryResolveProperty` compared `prop.ActualSchema.Type == JsonObjectType.Object`, but NJsonSchema emits `"type": ["object","null"]` for a nullable nested object, so the equality never matched. Every dotted mapping onto a nested read model member — `clear owner.note` and equally `owner.note = x` — was rejected as "Read model property not found". Now uses `HasFlag`. Verified by dumping the actual `JsonSchema.FromType<T>()` shape rather than reasoning about it.
2. **`ProjectionValidator.ValidateMappings` had no default arm**, so an unrecognized mapping silently skipped property validation and then surfaced from the visitor as an unhandled exception. Now a compiler error on the declaration with a line and column — same rejection, delivered as a diagnostic. Safe to add because all seven `MappingSyntax` subtypes in Screenplay 3.1.0 are handled, confirmed by decompiling the package.

## Two of the five call sites needed nothing, and that is spec-backed

`DeclarativeCodeGenerator` and `ModelBoundCodeGenerator` already branch on `WellKnownExpressions.Null`, because they read the *definition* rather than the syntax tree — so `clear` reaches them through the mapping the visitor produces. `clear_operation_as_client_code` passes with those files untouched and fails when the visitor arm is removed, which is what makes "needed nothing" evidence rather than an assumption.

## The falsification result is the part worth reading

Breaking the generator's emit line fails **6 facts** across three round-trip specs. But every `should_compile_the_clear_to_the_clear_expression` fact **still passed**, because both spellings recompile to `$null`. A round-trip spec asserting only on the recompiled definition would have been worthless here — the assertions on `CompilerResult.GeneratedDefinition` are what catch it. Sites 1 and 3 were falsified independently (23 facts fail, including the pre-existing `literal_expressions`).

## Verification

`dotnet build Chronicle.slnx -c Release` — 50 projects, 0 errors, 0 warnings. **8,285 specs passed, 0 failed** under CI's own filter (1 pre-existing skip). No generated-proxy drift.

## Deliberately not done

- **No nullability check on a clear target in the declaration language.** The client refuses a non-nullable member via CHR0048, but the validator only checks that the property exists. Adding one would make `clear x` stricter than `x = null`, and the whole point of the ruling is that they mean the same thing. Making *both* stricter is a coherent follow-up but is a behavior change to the released `= null` spelling.
- **The Workbench projection editor's help panel does not list `clear`** — it is already missing `nested`, `all` and `literal`. Frontend, outside the declaration language; worth its own issue.
- `Documentation/projections/architecture.mdx` carries a stale PDL example unrelated to clearing.

woksin:
**PII lifecycle acts are now logged — and I was overruled on how, correctly.**

I asked for the subject identifier to be written into the log line, reasoning that it is already in the clear on every appended event and in the fence itself, so it exposes nothing new. That was wrong, and the agent refused rather than complying silently. Two reasons, the second of which I had not considered:

1. **It would have reversed a committed decision.** `d15baa820` (2026-08-07), *"Stop naming the data subject in the key-propagation log"*, says: *"the feature whose purpose is to make a person's data unreadable was writing that person's identity, unencrypted, into a log the crypto-shred does not reach"* — and, pointedly, *"the design note beside it treats 'is a durable record that a named subject was erased itself personal data?' as a question Chronicle has to answer before a tombstone can be designed. **Shipping such a record by accident does not answer it.**"* Naming the subject in the erasure log while the propagation log deliberately omits it would have made one feature incoherent with itself.
2. **The fence argument does not carry over to logs.** The fence lives inside the key store — within the deployment's data boundary, destroyed when the store is. A log line leaves that boundary for an aggregator with its own retention and access, and genuinely outlives the erasure in a way the fence does not.

So the log carries a **stable one-way binding** of the subject instead. An operator investigating a known person computes the binding and greps for it; what it prevents is a name sitting in a log index forever. It is a pseudonym, not anonymization — a low-entropy identifier space is brute-forceable — and the documentation says exactly that.

## What is logged

Four messages in `PIIManagerLogMessages`, in the repo's `[LoggerMessage]` shape. **Both** operations, symmetrically — logging the un-erase while leaving the erase silent would read as an oversight rather than a decision. Each carries the namespace, the count of event stores reached and their names; the incomplete variants add the failure count. `Information` for the two completions, matching the repo's use of Information for statements of fact; `Error` for the two incomplete outcomes, matching `ErasedKeySurvivedInAnotherStore`, because a partial erasure needs operator action. Completion is logged after the fan-out *and* the eviction, and the incomplete path returns before the success message is reachable, so the two can never both appear for one act.

## Why this was worth doing before merge

Erasure used to be one-way: unlogged, but also unreversible, so the absence of a record was survivable. This PR makes crypto-shredding **reversible by authorization**. Someone can re-permit a subject whose personal data was erased under a legal obligation, and until this change Chronicle recorded nothing at all — `PIIManager` had no logger. Reversibility without an audit trail is worse than irreversibility without one.

Scope was held to logging. **No authorization gate** — *who may call it* is the host's concern, and adding a Chronicle-side gate would be a design decision nobody has taken.

## The no-key-material spec is deliberately non-vacuous

The shared given now has the store return real key material and an erasure carrying its fingerprint, so the assertion is scanning for strings that genuinely exist in the scenario — raw bytes, base64 and fingerprint, across every rendered message. It goes red the day someone adds a key to a log line. A `RecordingLogger<T>` captures level *and* rendered message; the existing precedent in `for_Jobs` records only levels, which cannot assert content.

## Still open for @einari, and now shipped rather than hypothetical

The erasure **fence** does store the raw subject identifier, argued in the docs as no more revealing than the subject every appended event already carries in the clear. That is a deliberate answer to the question `d15baa820` raised — but it is an answer, and it is worth a second opinion now that it ships.

Verification: 50 projects, 0 errors, 0 warnings; **8,307 specs green** under CI's filter; `for_PIIManager` integration 27/27 against a real silo and MongoDB.

woksin:
@einari — both of your review comments are addressed, and `clear` is the spelling end to end rather than in one API and not the other.

**Your first comment — `clear note` in the declaration language, not `note = null`.** The parser lives in Cratis/Screenplay, so it went there: [Cratis/Screenplay#116](https://github.com/Cratis/Screenplay/pull/116), merged and released as 3.1.0. `ClearMappingSyntax` sits beside `IncrementMappingSyntax` in the existing keyword-mapping family, with parser, printer, syntax walker, docs, Monaco and the VS Code grammar. This branch bumps to 3.1.0 and teaches Chronicle to read **and emit** it, so a clear round-trips as a clear rather than degrading back to an assignment. `note = null` keeps parsing exactly as it did — the new spelling is additive, never a rename.

**Your second comment — a formal `.Clear()` rather than `.Set().ToValue(null)`.** `.Clear(m => m.Property)` is on the fluent builder at root, child and nested level, from one addition to `IReadModelPropertiesBuilder` which all four surfaces compose over. It carries a throwing default implementation, so it breaks no existing implementor. `.ToValue(null)` still works — it was writing the literal string `"null"` and that fix stands — but the docs now lead with `.Clear(...)`.

Implementing them turned up three defects that had nothing to do with the spelling:

- A property path through a **nullable** nested object never resolved during validation — `ProjectionValidator` compared `Type == JsonObjectType.Object`, but NJsonSchema emits `["object","null"]`, so every dotted mapping onto a nested member was rejected as "property not found". Long-standing, and it affected `owner.note = x` just as much as `clear owner.note`.
- `ProjectionValidator.ValidateMappings` had no default arm, so an unrecognised mapping silently skipped property validation and then surfaced as an unhandled exception. Now a compiler error with a line and column.
- The fluent path had no nullability gate at all, so `.ToValue(null)` on a non-nullable member neither warned nor threw. `CHR0048` now covers all four spellings.

Merging on the maintainer's call rather than waiting for a re-review — nothing here is unfixable later, and the follow-ups are tracked: #3712 (proto generator), #3715 (a pre-existing wire break on the same upgrade path), #3716 (flipping the concurrency default in the next major).

Two things in the comments above are worth your eye when you have a moment, because they are now shipped rather than hypothetical: the **erasure fence stores the raw subject identifier**, which is a deliberate answer to the question `d15baa820` deliberately left open; and the **non-nullable clear is refused rather than defined as the type default**.

PR #3714, opened 2026-08-15
Release on push so fork contributions can publish

woksin:
# Summary

A pull request from a fork cannot cut a release, and the run reports success anyway — so a merged, labeled contribution silently publishes nothing.

## Changed

- Releases are cut on push to `main` instead of on the pull request closing, so contributions from forks publish like any other.
- The Publish run now fails when a merge releases nothing because the version label was missing, instead of reporting success.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,12 +1,12 @@
 name: Publish
 
-# Keyed by the pull request, never by a value that several simultaneously-closing pull requests share.
-# Merging a consolidation closes every pull request whose commits it carries, each of those fires its own
-# pull_request:closed event, and on a shared group cancel-in-progress means the last run to start cancels
-# the rest — including the one that was supposed to cut the release.
+# Never cancel a release in flight. On push every merge to main shares one group, so cancelling would let the
+# next merge kill a run that may already have tagged and cut the release but not yet finished publishing.
+# This also retires the hazard the pull_request trigger had, where a consolidation merge closed several pull
+# requests at once and their simultaneous runs cancelled each other: one push is one run.
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
-  cancel-in-progress: true
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: false
 
 env:
   NUGET_OUTPUT: ./Artifacts/NuGet
@@ -36,9 +36,13 @@ on:
         - info
         - warning
         - debug
-  pull_request:
-    types: [closed]
-    # Only a pull request into main releases. On "**" every base branch released, so merging one pull
+  # Releasing on push rather than on the pull_request closed event is deliberate. A pull request from a fork
+  # runs with a read-only GITHUB_TOKEN and no secrets even on merge, so it cannot create the release or reach
+  # the publishing credentials - which is how a merged, labeled fork contribution silently released nothing.
+  # A push to main always runs with a full-permission token, and the release action finds the merged pull
+  # request and its label from the commit.
+  push:
+    # Only main releases. On "**" every base branch released, so merging one pull
     # request into another one's branch - the ordinary way to stack work - cut and published a version
     # from a branch that was still in review. That is how Cratis.Fundamentals v7.17.0 came to be
     # published from the head of an open pull request, carrying every unmerged change on that branch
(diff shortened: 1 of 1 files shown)


woksin:
Reviewer context. Part of an org-wide migration; `Cratis/cli` is already migrated and has cut a release this way ([v2.8.1](https://github.com/Cratis/cli/releases/tag/v2.8.1)).

**Why.** A fork pull request runs with a read-only `GITHUB_TOKEN` and no secrets even on merge, so the release job cannot tag and the publish jobs have no credentials — while the run still reports success. `Cratis/cli`#72 merged with a `minor` label and released nothing. On push the token is full-permission, and the action resolves the pull request and its label live from the commit instead of from the event payload, which also closes the label-added-just-after-merge race.

**Not `pull_request_target`.** NuGet rejects it — `GitHubTokenPolicyValidator` checks `event_name` against `BannedGitHubActionsEvents`. `push` is not on that list, and cli published on push successfully.

**`cancel-in-progress` must be `false`.** Under `pull_request` each run had its own `github.ref`; on push every merge shares `refs/heads/main`, so leaving it `true` lets the next merge cancel a release possibly mid-publish.

**Depends on** [release-action v1.2.0](https://github.com/Cratis/release-action/pull/176), which added the `reason` output. `@v1` already points at it.

**The `paths` filter is kept deliberately**, against the action README's advice: the new check fails an unlabeled merge, and docs-only pull requests carry no label by design, so without the filter every docs merge would go red on main.

**Dropped the `github.event.pull_request.merged` guard** on the release job — it is never true under push, so leaving it would have stopped the release job from running at all.

**Rekeyed the existing `verify-published`** from `github.event_name == 'pull_request'` (never true under push) onto the `reason`, so it fires on `no-label`/`error` only and stays quiet for a direct push or a Dependabot merge.

**Verification.** Parses as YAML; trigger resolves to `push` on `main`; no `github.event.pull_request` or `event_name == 'pull_request'` references remain; the release job exposes `reason`. Not merged — over to you.

woksin:
Closing this without merging. The problem it identifies is real, so this is a record of what it found and where the fix now lives.

## What it set out to do

A pull request from a fork runs with a read-only `GITHUB_TOKEN` and no secrets even on merge, so under the `pull_request: [closed]` trigger it cannot create the release or reach the publishing credentials — and the run reports success anyway. A merged, correctly labeled fork contribution therefore publishes nothing, silently. Moving the trigger to `push` on `main` addresses that, because a push always runs with a full-permission token.

## Why it cannot land as written

Since this was opened, #3771 adds a `no-release` label — the fourth accepted answer in `verify-semver-label`, meaning "this changes nothing a consumer compiles against or runs, so there is nothing to version". Its exemption in the `verify-published` tripwire reads:

```yaml
if: always() && github.event_name == 'pull_request' && needs.release.result == 'success' && needs.release.outputs.publish != 'true' && !contains(github.event.pull_request.labels.*.name, 'no-release')
```

Both pull-request-shaped clauses go dead under a `push` event, and neither outcome is acceptable:

- **Keep that `if:` as it stands** and `github.event_name == 'pull_request'` is false for every push, so `verify-published` never runs at all — the guard against a merge that silently releases nothing stops existing.
- **Take this pull request's replacement `if:`**, which drops that clause, and the other one fails instead: `github.event.pull_request` is not populated on a `push` payload, so `github.event.pull_request.labels.*.name` yields nothing, `contains(…)` is false, and `!contains(…)` is therefore always true. The exemption never applies. A `no-release` merge publishes nothing *on purpose*, and would fail the Publish run on `main` every single time.

That is the crux, and it is worth stating plainly: **`no-release` expresses release intent as a pull request label, and a `push` event cannot read pull request labels.** Any move to a `push` trigger has to carry release intent in something a push can see — the merge commit, an API lookup of the associated pull request, or a different mechanism altogether.

There is a second interaction. This pull request also sets `concurrency.group` to `${{ github.workflow }}-${{ github.ref }}` with `cancel-in-progress: false`. Under `pull_request` that expression is per-pull-request, because `github.ref` is `refs/pull/<N>/merge`; under `push` it is `refs/heads/main` for every merge, which makes it a constant group. #3771 documents why a constant group must not be introduced: GitHub keeps only one pending run per group and cancels the earlier pending one, so a burst of merges would silently drop a release rather than fail loudly.

## On the overlap with #3756

#3756 rewrites the same `concurrency:` block to exactly these two lines, and #3771 already carries that change — with the comment corrected to describe what the group actually does under the `pull_request` trigger. That half is handled; nothing is waiting on this pull request for it.

## The goal is folded in, not dropped

Both halves of what this was after — letting fork contributions publish, and not racing the release job — are the point of **task 2.2, "Release train with a soak", in `PLAN-reliability-program.md`**: decouple merge from release so merges accumulate, and a scheduled or manually dispatched job takes the current `main`, soaks it through the out-of-process matrix plus the boot and consumer-smoke gates, then publishes once. That job runs on the repository's own ref with full permissions, so a fork-originated merge in the batch publishes like any other; and only one release run exists at a time, so there is nothing left to race. The trigger question stops being load-bearing once the per-merge release model is gone.

The branch `chore/publish-on-push` stays in place, so none of this work is lost. A note under task 2.2 now records that the trigger choice and the `no-release` exemption are the same decision, so whatever the release train does about triggers has to solve label-based release intent at the same time.

PR #3719, opened 2026-08-17
Gate ephemeral OpenIddict keys on the DEVELOPMENT symbol again

einari:
## Fixed

- The `*-development` and `*-development-slim` container images start again without an encryption certificate. Since 16.33.1 they aborted at startup with `An encryption certificate is required in production for OpenIddict key security`, because the check for a development build had regressed to the `DOTNET_ENVIRONMENT` / `ASPNETCORE_ENVIRONMENT` variables the images never set; it is a compile-time `DEVELOPMENT` check again, matching how the development TLS certificate is decided.


einari:
Superseded by #3721, which carries this fix together with the Workbench look.

Issue #3712, opened 2026-08-15
generate-protos.sh regenerates nothing and exits 0, so the checked-in .proto files drift from the contracts unnoticed


woksin:
## Problem

`Source/Kernel/Protobuf/generate-protos.sh` cannot regenerate **any** of the checked-in `.proto` files. It aborts for 22 of the contract packages — including `Cratis.Chronicle.Contracts.EventSequences`, `Cratis.Chronicle.Contracts.Events` and `Cratis.Chronicle.Contracts.Events.Constraints` — with:

```
Error generating proto for package Cratis.Chronicle.Contracts.<X>: 'Cratis.Chronicle.Contracts.Events.Constraints.UniqueEventTypeConstraintDefinition' carries ReservedProtoFieldsAttribute, but no 'message UniqueEventTypeConstraintDefinition' was found in the generated schema, so its retired field numbers would not be reserved.
```

The reservation check in `ProtoSchemaHelper` looks for a `message UniqueEventTypeConstraintDefinition` in the schema being emitted for the package it is currently generating. `UniqueEventTypeConstraintDefinition` is not reachable from most of those packages' service graphs, so no such message is emitted, and the check treats "the type is not in this schema" the same as "the reservation could not be applied" — and aborts the whole package.

**The script still exits 0.** That is why this has gone unnoticed: `.github/workflows/generate-protos.yml` runs the generator, sees a success exit code, and uploads whatever is on disk — which is the previous, unchanged content for every failed package.

## Reproduction

On a clean checkout of `main`:

```bash
./Source/Kernel/Protobuf/generate-protos.sh; echo "exit code: $?"
```

Observed:

- 22 lines of `Error generating proto for package …`
- `exit code: 0`
- `git status Source/Kernel/Protobuf/` reports **no changes** — nothing was regenerated

## Impact

1. **The `.proto` files silently drift from the contracts.** They are the source of truth for the Kotlin, TypeScript and Elixir clients (`Source/Clients/{Kotlin,Elixir}/generate-protos.sh`, `publish-kotlin-client.yml`, `typescript-build.yml` all consume `Source/Kernel/Protobuf/`). A new `[ProtoMember]` on a contract reaches the .NET client but never reaches the other three.
2. **Contributors have to hand-edit generated files.** #3635 needed three new contract fields and had to hand-apply them to `eventsequences.proto`, with a comment in the file pointing here.
3. **CI cannot catch either of the above**, because of the exit code.

## Suggested fix

[...]

Issue #3725, opened 2026-08-17
Projections a client no longer registers are never retired — orphaned definitions keep consuming events, and a renamed read model races its successor on the same container (source of every remaining failed partition in prod)


einari:
## Summary

Chronicle never retires a projection that a client stops registering. When a read model is **renamed or deleted** in the client, the old projection definition stays stored, its observer stays `Active` and keeps consuming every new event, and its sink keeps writing — into a container that, for a *renamed* read model, is the **same container** the successor projection writes to (container name is derived from the type name, which did not change). The result in production today: two projections with different definitions racing on the same documents (fields disappear), plus permanently failed partitions from the stale definition.

Chronicle **16.32.0**, Studio event store. Found via `cratis chronicle failed-partitions list` — every failed partition in the store except the one in #3722 belongs to an orphaned projection.

## Evidence (namespace `Cratis` unless noted; every orphan is at `next=1668` = the current tail)

| stored projection (no longer in any client) | removed/renamed in client | observer state | container |
|---|---|---|---|
| `Studio.Projects.Applications.Domains.EventModels.Brainstorming.BrainstormingBoard` | renamed 2026-07-27 to `…Domains.Brainstorming.BrainstormingBoard` | Active, next=1668, last=1638 | **`brainstormingBoards` — same as the successor** |
| `Studio.Projects.Users.ProjectUsers` | deleted 2026-08-05 | Active, next=1668, last=1054 | `projectUsers` |
| `Studio.Settings.AI.Models.Listing.AIModel` | renamed 2026-08-11 (→ AI Providers) | Active, next=1668, last=1015 | `AIModels` |
| `Studio.Settings.AI.Models.Resolving.ConfiguredAIModel` | renamed 2026-08-11 | Active, next=1668, last=1015 | `configuredAIModels` |

`cratis chronicle projections list -e Studio` returns 72 definitions; the client registers fewer. `projections show` for the old BrainstormingBoard still returns the pre-rename declaration (root `from EventModelCreatedForDomain`, `children stickyNoteRegionMemberships … from BrainstormingStickyNoteAddedToRegion key stickyNoteId parent eventModelId`), while the successor is `from BrainstormingCreatedForDomain` with `parent $eventSourceId`.

**Failed partitions caused by the orphan** (`failed-partitions list`, observer `…EventModels.Brainstorming.BrainstormingBoard`):
[...]

PR #3734, opened 2026-08-17
Superseded by #4009: Fail proto generation loudly and fix the schema defects it was hiding

woksin:
> **Superseded — not a merge candidate.**

The generator/schema work was superseded by merged #3768. The remaining real-protoc validation was included in merged #4009 and shipped in v18.1.0. Do not merge this obsolete branch into current main.

Closure is pending the repository's package-cleanup access safeguards. The branch is retained to avoid losing historical work.

---

## Added

- Generated proto schemas are validated with a real protoc before they feed the Kotlin, TypeScript and Elixir client publishes (#3712)

## Fixed

- Proto generation no longer reports success when packages fail to generate — the run exits non-zero and names every failure, so stale schemas can no longer ship silently (#3712)
- The reserved-fields reservation check is scoped to the whole generation run instead of every package's schema, so all 23 packages generate again (#3712)
- Enum values that collide with a message name in the same schema are prefixed, making the regenerated `eventsequences.proto` valid protobuf (#3712)
- Client proto schemas are regenerated from the current contracts, restoring the sequence-query surface and other drift the broken generator withheld (#3712)


woksin:
Reviewer context (not release notes): this re-lands the two commits reverted out of #3713 on Aug 15 (scoping fix + regen) and resolves the blocker documented in the #3712 comment — the enum-value/message collision. Verification: 23/23 packages generate with exit 0 locally; all 23 schemas compile with a real protoc (grpcio-tools); ProtoGenerator.Specs 20/20 green incl. two new specs pinning both enum-collision modes; deliberate failure exits non-zero. The renamed enum value (EVENT_SEQUENCE_QUERY_SORT_BY_EventType) has never appeared in a shipped schema — verified absent from the checked-in eventsequences.proto — so no published client API changes. The projections.proto FullSet field is contract drift from #3725 that accumulated in the two days since the revert.

woksin:
Holding this one rather than merging it with the rest of today's batch, deliberately.

It regenerates the checked-in `.proto` files, which are the source the Kotlin, TypeScript and Elixir clients are generated from — a published wire surface. Einar is mid-investigation on exactly that surface (per-minor breaking-change sweep, moving the protocol compat check into the server), so regenerating it underneath him is the wrong move even though the regeneration itself is verified: all 23 packages generate, every schema compiles with a real protoc, and the three hand-applied fields keep their field numbers.

Worth flagging for that investigation, since it is directly relevant: the hand-applied fields in `eventsequences.proto` were **not** rogue editing. `9eddffe28` added them with a seven-line note explaining the generator was silently broken (#3712 — 22 of 23 packages failing while exiting 0), that the fields match what protobuf-net would emit, and to re-run the generator and drop the note once fixed. This PR is that fix and does exactly that.

The one behavioral change to look at: `EventSequenceQuerySortBy.EventType` gets prefixed to avoid colliding with the `EventType` message, because proto3 enum values are siblings of their type. I verified that enum has never appeared in a published schema, so no client API changes — but it is an enum rename, which is the category his sweep flags, so he should be the one to confirm it.

Ready whenever he has looked; nothing else depends on it.

woksin:
Recording where this stands: #3768 supersedes this, and it will be closed once #3768 merges. Leaving it open until then.

#3768 regenerates every one of the six `.proto` files this branch fixes — `eventsequences`, `eventstores`, `jobs`, `namespaces`, `projections` and `sequencequeries` — as part of a wider regeneration that also covers `clients`, `host`, `identities`, `recommendations` and `security`, and it revises the same `ProtoGenerator` sources (`Program.cs`, `ProtoSchemaHelper.cs`). On top of that it adds a real descriptor-set generator (`DescriptorSetWriter`, emitting `chronicle.desc`), embeds the descriptor set in the shipped clients, and builds the `WireCompatibilityChecker` gate over it — so the contract is compared as protobuf descriptors rather than as generated `.proto` text.

Nothing here is being discarded on the strength of that alone, which is why this stays open: if #3768 stalls, the loud-failure fix and the schema defects it uncovered are still wanted. Close this once #3768 lands.

PR #3735, opened 2026-08-17
Boot-test every Docker image before publishing

woksin:
## Added

- Every published Docker image variant (production, development, development-slim, workbench) is booted the way a consumer runs it before being pushed — the container must start, stay up, and listen on its port, or the publish fails with the container logs (#3446)

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -438,6 +438,23 @@ jobs:
           echo -e "$TAGS" >> $GITHUB_OUTPUT
           echo "EOF" >> $GITHUB_OUTPUT
 
+      # A published image that cannot start must fail here, not at the consumer - 15.37.0-15.38.2 shipped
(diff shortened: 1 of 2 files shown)


woksin:
Reviewer context: verified locally against the production image on arm64 — healthy image boots and passes (Mongo replica-set machinery included); rebuilding it with the historical exec-bit defect reproduced (bit stripped on the artifact + chmod removed) fails the gate with exit 126; a container that exits at startup is caught and reported with its logs. The gate builds the native-arch image with load:true before the multi-arch push, so buildx layer cache keeps the extra cost to the amd64 leg. Note: a variant failing its gate blocks that job's push; other variants' jobs run in parallel, so a partial publish is still possible — far better than today's publish-broken-everything, and the release train (plan Task 2.2) later removes the partial-publish window. The development-variant zero-config path exercises fully on the first real publish run.

PR #3736, opened 2026-08-17
Add a watchdog that files an issue for persistently red workflows

woksin:
# Summary

Internal CI only: a scheduled workflow-health watchdog that files a GitHub issue when any workflow has failed its last three completed runs on main, so persistent red becomes an actionable object instead of background noise. No user-facing changes.

### DIFF
--- a/.github/workflows/workflow-health.yml
+++ b/.github/workflows/workflow-health.yml
@@ -1,0 +1,113 @@
+name: Workflow Health
+
+# A workflow that fails every day and is never fixed teaches everyone to ignore it, and once
+# ignored it is indistinguishable from a workflow that does not exist. This watchdog turns
+# "persistently red" into an actionable object: a GitHub issue, filed once, that somebody has
+# to close by fixing the workflow or deleting it.
+
+on:
+  schedule:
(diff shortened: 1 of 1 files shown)


woksin:
Reviewer context (Opus-implemented, Fable-reviewed): the decision logic was executed read-only against this repo's live run data with issue creation stubbed — no writes were made. Today it would flag three workflows as persistently red: **integration.yml** (all 23 of its completed runs on main have failed — it has never been green — and it stopped running entirely on 2026-06-10), **dotnet-build.yml** (nightly red for 6 consecutive runs since 2026-08-12, last green 2026-08-11 — looks like a live regression, not rot), and **benchmarks.yml** (11 consecutive failures since 2026-08-14). Streak counts only decisive runs (success/failure); cancelled/skipped are dropped so a cancellation cannot reset a streak. Dedupe is by exact issue title; an existing open issue suppresses re-filing. YAML validated with actionlint (0 findings). Note: Dependabot Updates is flapping (fail/success/fail) and deliberately reads healthy under the 3-streak rule — a failure-rate detector would be a separate follow-up.

PR #3737, opened 2026-08-17
Run the orphaned spec projects and fail test jobs that execute zero tests

woksin:
# Summary

Internal CI and test-coverage hardening only: spec projects that were missing from the solution (and therefore never ran anywhere) are added and now execute in CI, and every CI test job now fails when it executed zero tests instead of reporting an empty run as a pass. No user-facing changes.

### DIFF
--- a/.github/actions/assert-tests-ran/action.yml
+++ b/.github/actions/assert-tests-ran/action.yml
@@ -1,0 +1,24 @@
+# Copyright (c) Cratis. All rights reserved.
+# Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+name: Assert tests ran
+description: >-
+  Fails the job when the TRX files a test step produced report zero executed tests.
+  A run that matches no tests still exits successfully, so without this a broken filter,
+  a project missing from the solution, or a test host that never started all read as a pass.
+
+inputs:
+  path:
+    description: Directory searched recursively for *.trx result files.
+    required: true
+  name:
+    description: Name of the test run, used in the log and failure message.
+    required: false
+    default: The test run
+
+runs:
+  using: composite
+  steps:
+    - name: Assert the executed test count is not zero
+      shell: bash
+      run: python3 "${GITHUB_ACTION_PATH}/assert-tests-ran.py" "${{ inputs.path }}" "${{ inputs.name }}"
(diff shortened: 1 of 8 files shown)


woksin:
Reviewer context (Opus-implemented, Fable-reviewed): Connections.Specs (57 specs) and Services.Specs (46 specs) had never run anywhere — not in the solution, not named by any workflow. Both now pass under the exact CI invocation (settings + filter). Storage.AzureKeyVault was already compiled transitively via Server.csproj; adding it to the solution is for visibility. Two projects deliberately NOT added, follow-up issues incoming: XUnit.Integration.Specs (65/66 pass; the failing packaging-closure spec flags Cratis.Arc.ProxyGenerator.Build as an undeclared dependency, a false positive — the parser cannot evaluate the MSBuild condition that disables it; needs a packaging-owner ruling) and Storage.AzureKeyVault.Specs (needs Docker plus a trusted emulator root CA; its NU1903 SSH.NET restore blocker is fixed in this PR the same way four sibling projects do it). The zero-test guard is load-bearing: verified that dotnet test EXITS 0 when a filter matches nothing, writing executed="0" — the guard turns that into a red job. Guard verified three ways locally (real run 57 → pass; empty dir → fail; no-match filter TRX → fail). The integration shards already emit TRX via run-tests-with-retry.sh into ./coverage-results, so only guard steps were added there. This PR's own CI run exercises all six guards for real.

woksin:
Added on review: the solution-membership gate (`assert-solution-membership.py`) — a spec project missing from Chronicle.slnx never runs and emits no TRX, so the zero-test guard is structurally blind to it. Hard-fails on unlisted orphan spec projects (the two known ones are allowlisted with their tracking issues #3738/#3739), warns on unreachable non-spec projects, and holds the spec folder-structure line (one pre-existing deviation grandfathered in a named constant: Core.Specs/Services/Observation/when_resolving_grpc_observation_services lacks a for_* parent — fix the folder, delete the entry). Chronicle.slnx added to the workflow's paths trigger. Verified: passes at head (52 on disk / 51 in slnx / 51 reachable); dropping an allowlist line, adding a dummy orphan spec project, and adding a misplaced when_ folder each fail with a named ::error::.

PR #3740, opened 2026-08-17
Provision the namespace database before its first table read on SQL backends

woksin:
## Fixed

- Appending events to a second, non-default namespace on PostgreSQL or SQL Server no longer fails with "database does not exist" — the per-namespace database is now provisioned before its event-sequence tables are first read, the same way read-model contexts already do

### DIFF
--- a/Integration/Client/for_ReadModels/given/a_projection_with_events.cs
+++ b/Integration/Client/for_ReadModels/given/a_projection_with_events.cs
@@ -31,16 +31,22 @@ public class a_projection_with_events(ChronicleFixture chronicleFixture) : Speci
     }
 
     /// <summary>
-    /// Polls the materialized read model until it holds at least <paramref name="expectedCount"/>
-    /// instances, or the default timeout elapses.
+    /// Polls the materialized read model until at least <paramref name="expectedCount"/> instances have
+    /// caught up with every event appended to their event source, or the default timeout elapses.
     /// </summary>
-    /// <param name="expectedCount">The minimum number of instances to wait for.</param>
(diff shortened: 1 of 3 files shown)


woksin:
Reviewer context (Opus-diagnosed and implemented, Fable-reviewed): this resolves the nightly dotnet-build red streak (6 consecutive nights since Aug 12). Two independent causes. (1) Real kernel regression fixed here: Storage.Sql's GetOrCreateTableDbContext never called EnsureDatabaseExists, so the first table read of a not-yet-provisioned namespace database failed with 3D000/login errors — surfaced by the PIIManager namespace-scoping spec added Aug 15. (2) Test defect fixed here: the GetInstances specs asserted First()/Last() over sink results, but neither sink orders — MongoDB returns insertion order, SQL Server clustered-key order — a divergence exposed when #3610 made GetInstances read the sink; the wait predicate also only counted documents, racing the second event. Verification against real backends: baseline reproduction with specs reverted gives the exact CI failure signature; with fixes, mssql when_getting_instances 13/13, for_PIIManager 27/27 on both postgres and mssql, for_ReadModels 122/122 postgres, sqlite 149/149, mongodb in-process 149/149 (no PR-path regression), full solution Release build 0W/0E. Why PRs stayed green: the matrix runs mongodb only on pull_request (all-providers is schedule-only) — making one SQL provider PR-blocking is queued in the plan. Follow-ups being filed: GetInstances pages an unordered sink (correctness hazard under concurrent writes); benchmarks.yml red streak has an unrelated one-line cause (git pull --rebase with unstaged BenchmarkDotNet output).

PR #3748, opened 2026-08-18
Make the existing CI checkers actually enforce

woksin:
# Summary

Internal CI only: gates that already existed but could not fail anything are now blocking — the gRPC wire-contract comparison, the CodeQL configuration file, the AI corpus validator, and the declared license-header rule — plus a ledger for tests a retry rescued. No user-facing changes.


woksin:
Reviewer context: five checkers existed but enforced nothing.

1. **gRPC compatibility** — the compare step opened with `set +e` and its `has-breaking-changes` output was consumed by nothing, so a removed service/rpc or changed signature could not fail a PR. Now fails, with `grpc-breaking-change-approved` as the deliberate escape hatch (implies `major`). The orphaned `update-pr-description.sh` and the README section documenting an integration that never existed are removed.
2. **CodeQL** — `codeql-config.yml` (incl. the deliberate CA1031 exclusion) was never passed via `config-file`, so every scan ran the unfiltered default suite. Dead manual-build-mode step dropped.
3. **AI corpus validator** — `validate-ai-setup.sh` is a strict exit-1 checker that no workflow ran. Wiring it surfaced real drift, all fixed here: `terminal-commands.md` had no `applyTo`, two prompts lacked the `.prompt.md` suffix their adapters resolve through, four agents had no Copilot adapter, two Claude command adapters were missing.
4. **License header** — `.editorconfig` declares `file_header_template` but IDE0073 (the rule that reads it) had no severity. Now `warning` → error in Release for `Source/**`. Verified zero violating source files. (One curiosity: `Source/Clients/DotNET.Specs/.../for_UniqueConstraintBuilder/when_adding_on_using_generics.cs` is a **directory** with a `.cs` suffix — harmless, but worth renaming separately.)
5. **Flake ledger** — a retry-rescued test previously left a green check and no trace. Now each is named in a `::warning::` and appended to `flaky-tests.log` in the results dir CI already uploads. This is what makes timing decay (category G) visible.

Also records the enforcement-audit backlog as PLAN §8 (8.a–8.h) with evidence and per-item DoD, and brings the status board current.

Verification: full Release solution build in this branch after merging main — 50 projects, 0 errors, 0 warnings; all three workflow YAMLs parse; `bash -n` clean on the retry script; validator exits 0 (3 advisory warnings only).

woksin:
Holding this one rather than merging it with the rest of today's batch, deliberately.

Everything else from this session is merged (#3734 aside — see below), but this PR makes the gRPC compatibility gate **blocking**, and Einar is actively working on protocol compatibility right now: a per-minor breaking-change sweep, moving the compat check into the server, and a client compatibility battery. A newly-blocking gate landing mid-flight could start failing his PRs for reasons he did not ask for and has not seen the design of. The escape hatch (`grpc-breaking-change-approved`) is discoverable from the error text, but discovering it by being blocked is not a good first introduction.

It is also worth him seeing what this gate does and does not catch before it becomes load-bearing: it detects **removed service, removed rpc, and changed rpc signature only**, and parses services by bare name with no package qualification — so a service moving proto package is not detected, and there is no message, field or enum analysis. Of the 53 breaking changes his sweep found in 16.34.0, this would have caught some and missed others. His own tool looks more thorough, and the two should be reconciled rather than stacked blindly.

The other four items in this PR are uncontroversial on their own (CodeQL config actually applied, AI corpus validator wired, license-header rule enabled, flake ledger). If the gRPC part is the only sticking point, I am happy to split it out so the rest can land.

No rush from my side — nothing depends on it.

woksin:
Closing this without merging — it has been overtaken on both halves.

**Four of its five items already landed on `main`.** #3762 carried the CodeQL configuration that was never being passed, the AI corpus validator (and the corpus fixes it caught), the license-header rule, the flake ledger that records the tests a retry rescued instead of losing them, and §8 of the reliability program plan. None of that is still outstanding here.

**The one remaining item, the gRPC compatibility gate, is superseded by #3768.** This branch made `grpc-compatibility.yml` fail on a breaking change, but the comparison underneath it is `compare-grpc-schemas.sh` diffing two generated `.proto` files by name — it notices a service or a method disappearing, and not much else. #3768 replaces that with a real `WireCompatibilityChecker` over protobuf descriptor sets, with specs covering removed fields, changed field types, fields widened in place, altered call shapes and changed enum values. That is strictly more than a name-level text comparison can see, and keeping both would leave two overlapping gRPC gates in CI with the weaker one generating the noise.

**The one orphan worth keeping has been carried across.** The deletion of `.github/scripts/update-pr-description.sh` is now in #3771. The claim that it was dead code was verified before acting: no workflow invokes it, and the only mentions anywhere in the repository were its own usage line and its entry in `.github/scripts/README.md`, which has been pruned along with it.

The `fix/wire-existing-gates` branch is deliberately left in place so the work stays recoverable.

PR #3749, opened 2026-08-18
Keep replay results and child removals on the in-memory sink

woksin:
## Fixed

- Completing a replay against the in-memory sink no longer discards everything the replay produced — the replayed documents are promoted the way the persistent sinks swap in their replay container, while an empty replay is still left as a no-op
- Removing a child from all read models that match on identity now applies on the in-memory sink instead of being silently ignored

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -438,6 +438,23 @@ jobs:
           echo -e "$TAGS" >> $GITHUB_OUTPUT
           echo "EOF" >> $GITHUB_OUTPUT
 
+      # A published image that cannot start must fail here, not at the consumer - 15.37.0-15.38.2 shipped
+      # unable to start because nothing ever booted what was pushed. Build the native arch locally and boot it.
+      - name: Build boot-test image
+        uses: docker/build-push-action@v5
+        with:
+          builder: ${{ steps.buildx.outputs.name }}
+          context: .
+          file: ./Docker/Production/Dockerfile
+          platforms: linux/amd64
+          load: true
+          tags: chronicle-boot-test:production
+          build-args: |
+            VERSION=${{ needs.release.outputs.version }}
+
+      - name: Boot the image before publishing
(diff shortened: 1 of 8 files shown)


woksin:
Reviewer context: first slice of plan Task 1.1 — both defects were found by diffing the in-memory sink against MongoDB/SQL, not by a failing test, which is exactly the category-C signature (a divergence does not fail, it makes specs pass vacuously).

**Replay (D-8):** `EndReplay` only set `_isReplaying = false`; the rewind collection holding every replay write was never promoted, so the sink kept serving pre-replay documents. Now promotes, with MongoDB's zero-write guard preserved (an empty replay must not empty the read model — that guard exists because of a real prior incident), and `BeginReplay` clears stale rewind state so a second replay cannot inherit the first's documents.

**ChildRemovedFromAll (G1.3):** the change switch had no case and fell through silently; MongoDB (Sink.cs:211) and SQL both handle it. Because it spans documents it runs across the collection at the same point MongoDB runs it, not inside per-document change application.

**Verification — every spec checked red-without/green-with:** replay specs 3/3 pass with the fix, 2/3 fail without; child-removal 2/2 pass with, 2/2 fail without. Worth noting: the first version of the child-removal spec passed *either way* — it asserted on the read model being written, whose state is rebuilt from the changeset. That vacuous assertion was replaced with one on a model the write never touches plus a check that non-matching children survive. Full suite 147/147; solution Release build 50 projects, 0 warnings.

**Deliberately not changed:** `InMemorySink.Remove(containerName)` ignores its argument and clears rewind state (audit item D-9). MongoDB drops the named collection, but the in-memory sink has no container map, so the correct semantics are a design decision rather than a convergence — left for the contract-suite work.

**Next in Task 1.1:** extract the three hand-copied `when_applying_changes_guarded_on_watermark` suites (verified byte-identical apart from namespace) into one abstract suite in `Storage.Specs` with a per-backend sink factory, then run these replay/child cases through it against MongoDB and SQL too.

PR #3750, opened 2026-08-18
Run one sink contract against every storage implementation

woksin:
# Summary

Internal specs only: the sink behavioral contract is written once and executed against each storage implementation, replacing three hand-copied copies of the same cases. No user-facing changes.


woksin:
Reviewer context: plan Task 1.1 — this is the seam, delivered on the case that proves it is mechanical.

**What moved:** the four `when_applying_changes_guarded_on_watermark` cases existed three times, verified byte-identical apart from the namespace. They now live once in `Storage.Specs/Sinks/for_ISink/` as abstract classes generic over an `ISinkHarness`, and each backend contributes one harness plus four one-line drivers. `Storage.InMemory.Specs` and `Storage.Sql.Specs` now reference `Storage.Specs`. **No solution-file change** — the contracts live in an existing project deliberately, to avoid colliding with #3737 which owns `Chronicle.slnx`.

**Definition of done, demonstrated:** replacing the in-memory watermark guard with `false` turns the *shared* suite red on the in-memory run (2 of 4 fail) while SQL stays green. That is the whole point — one divergence, one failure, no copy to keep in sync.

**A finding worth reading, because it is this program in miniature:** my first version discovered **zero tests** and reported success. The repo style writes `[Fact] void should_x()` with no access modifier, i.e. private — fine on a concrete class, invisible on the closed subclass xUnit collects. The suite passed having run nothing. The contract's assertions are now `public`, with a comment saying why. Note the zero-test guard from #3737 would have caught this in CI; here it was caught by checking the count by hand.

**Verification:** in-memory 4/4 and SQL 4/4 through the shared contract; full suites 233 / 147 / 90 green; solution Release build 50 projects, 0 warnings.

**Deliberately deferred:** MongoDB keeps its own copy for now — its cases receive a container fixture through the constructor, so it needs a harness that can be handed one rather than `new()`-constructed. That plus lifting the replay and child-removal cases from #3749 into the contract is the next increment.

PR #3751, opened 2026-08-18
Run the replay contract against the SQL sink too

woksin:
# Summary

Internal specs only: the sink replay cases move into the shared contract so they run against the SQL sink as well as the in-memory one. No user-facing changes.

Builds on #3749 and #3750 — merge those first.


woksin:
Reviewer context: **stacked on #3749 and #3750 — merge those first.**

The point of this one was to answer a question I could not answer by reading: does the SQL sink have the replay bug I fixed in memory in #3749? **It does not** — SQL promotes correctly via its rename swap. But putting it under the same contract found two things anyway, both in the specs rather than the product:

1. **The read model definition took the container name in the display name's position.** `ReadModelDefinition` is `(Identifier, ContainerName, DisplayName, …)`, and all three hand-copied contexts passed the container in slot 3. Every backend was writing to a container named after the display name — harmless while each copy only had to agree with itself, wrong the moment two implementations must agree with each other. Exactly the class of thing a shared contract exists to expose.
2. **The second-replay case began a replay against a read model that had never been written.** No engine does that, and the SQL sink renames the primary container aside when a replay ends, so it needs one to exist. Arrangement corrected to write first.

**Not claimed as a product bug, but worth someone's judgment:** if Chronicle can ever reach `EndReplay` on a SQL-backed container whose primary table does not exist yet, `PerformRenameSwap` will fail on the rename. In production `EnsureTableExists` normally creates it first, so I could not show it is reachable — flagging rather than filing.

**Verification, each checked both directions:** in-memory 7/7 and SQL 7/7 through the shared contract; disabling the in-memory promotion turns the in-memory run red (2/3) while SQL stays green — one divergence, one failure, and the backend that is correct stays quiet. Full suites 233 / 147 / 93; solution Release build 50 projects, 0 warnings.

**Still open on Task 1.1:** MongoDB, which needs a harness that can be handed its container fixture rather than `new()`-constructed; and the child-removal case from #3749, which should run everywhere too.

PR #3752, opened 2026-08-18
Remove the container the in-memory sink was asked to remove

woksin:
## Fixed

- Removing a read model's container from the in-memory sink now removes the read model instead of being ignored, and removing a revert container no longer discards a replay that is still in progress

Builds on #3749, #3750 and #3751 — merge those first.


woksin:
Reviewer context: **stacked on #3749, #3750 and #3751 — merge those first.**

This is audit item D-9, and it is the first defect the shared contract found rather than confirmed. The audit *suspected* `InMemorySink.Remove` ignored its argument; the contract proved it, in the run where the case was written: SQL passed both new cases untouched, in-memory failed the Remove one.

**Why it matters beyond the in-memory sink:** `Remove` had **no spec on any implementation** — the sink replay lifecycle (`Remove`, `PrepareInitialRun`, `ResumeReplay`, `BeginReplay`, `EndReplay`) was entirely uncovered, which the audit called the highest defect-per-spec ratio in the repo. A method that answers to nothing keeps answering to nothing.

**The behavior:** the sole production caller passes a revert container name after a completed replay. Ignoring the name meant removing the read model's own container left it readable, and removing a revert container while a replay was in flight discarded the live replay state — opposite outcomes in both directions. It now drops whichever container it is handed, matching MongoDB (drops the collection) and SQL (`DROP TABLE IF EXISTS`).

**Verification:** the new `when_removing_a_container` and `when_preparing_an_initial_run` cases run through the shared contract on both backends — 9/9 each across the four sink behaviors now covered. Full suites 233 / 149 / 95; solution Release build 50 projects, 0 warnings.

**Still uncovered on the lifecycle:** `ResumeReplay` has no case yet (it needs an interrupted-replay arrangement), and MongoDB is not yet wired into the contract.

Issue #3743, opened 2026-08-17
GetNextSequenceNumberGreaterOrEqualThan returns the highest match on MongoDB, so failed-partition retry skips events


woksin:
Found by the storage-divergence audit (read-only, HEAD 050ed6069).

The three implementations disagree on `IEventSequenceStorage.GetNextSequenceNumberGreaterOrEqualThan`: MongoDB sorts **descending** and returns the *highest* matching sequence number (Storage.MongoDB/EventSequences/EventSequenceStorage.cs:713-718), while InMemory (MinBy, :361-365) and SQL (OrderBy().First(), :574-580) return the *lowest*.

Live callers: `Core/EventSequences/EventSequence.cs:191` and `Core/Observation/Jobs/RetryFailedPartition.cs:127` — on the default MongoDB backend, a failed-partition retry can jump to the tail and **silently skip every event in between**. No spec covers this method on any backend.

Fix: make MongoDB ascending like the others; add the method to the shared storage contract suite so all three implementations are pinned together.

woksin:
Correction to this report's severity, from investigating the call sites while fixing it in #3753.

The divergence is **real and confirmed** — MongoDB sorted descending and answered with the last qualifying event (the variable was named `highest`), where SQL and in-memory answer with the nearest. A spec written against the old implementation fails with exactly the wrong values.

But the claim I opened this issue with — that failed-partition retry silently skips events — **is not accurate**. `RetryFailedPartition` (Source/Kernel/Core/Observation/Jobs/RetryFailedPartition.cs:127) only uses the result as an existence check (`return nextSequenceNumber.IsActualValue`), so the ordering does not change its behavior. The other consumer is the kernel's own event sequence grain, which returns the value onward.

So: a latent correctness divergence in the storage contract rather than a live data-loss path. Still worth fixing — one method meaning two different things across implementations is how missing data becomes unaccountable later — but nobody should be reading this as an active production incident.

woksin:
Verified fixed on main: `Storage.MongoDB/EventSequences/EventSequenceStorage.cs` now sorts ascending (`SortByAscendingSequenceNumber()`) with an in-code comment describing exactly this bug. Closing as completed via org-wide issues sweep.

PR #3753, opened 2026-08-18
Answer with the nearest event at or after a sequence number

woksin:
## Fixed

- Asking a MongoDB-backed event sequence for the next sequence number at or after a position now answers with the nearest qualifying event instead of the furthest, matching the SQL and in-memory implementations (#3743)

### DIFF
--- a/Source/Kernel/Storage.MongoDB.Specs/EventSequences/for_EventSequenceStorage/when_getting_the_next_sequence_number_greater_or_equal_than/and_several_events_qualify.cs
+++ b/Source/Kernel/Storage.MongoDB.Specs/EventSequences/for_EventSequenceStorage/when_getting_the_next_sequence_number_greater_or_equal_than/and_several_events_qualify.cs
@@ -1,0 +1,37 @@
+// Copyright (c) Cratis. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Cratis.Chronicle.Concepts.Events;
+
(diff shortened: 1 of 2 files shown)


woksin:
Reviewer context: resolves #3743. **Independent of the #3749→#3752 stack** — different files, merges in any order.

**The defect:** MongoDB's `GetNextSequenceNumberGreaterOrEqualThan` sorted descending and returned the *last* qualifying event; the variable was literally named `highest`. SQL and in-memory both return the nearest. The name says which is right, and the backend the server actually runs on was the odd one out.

**Severity — correcting my own earlier overstatement.** When I filed #3743 I repeated the audit's claim that failed-partition retry "silently skips events". Reading the call sites, that is not accurate: `RetryFailedPartition` only uses the result as an existence check (`return nextSequenceNumber.IsActualValue`), so highest-vs-lowest does not change its behavior. The other consumer is the kernel's own event sequence grain, which returns the value to its caller. So this is a **latent divergence, not a reported production fault** — worth fixing because the storage layer disagreed with itself about what the method means, not because data is being lost today. I will note the correction on #3743.

**Verification:** the new spec fails against the old implementation with exactly the wrong values (asking for next ≥ 1 returned 2; next ≥ 0 returned 2) and passes after the fix. Full MongoDB suite 730/730 — these are real container-backed specs, so the fix is exercised against an actual replica set. Solution Release build 50 projects, 0 warnings.

**Care taken:** two other call sites in the same file sort descending correctly (tail lookups and `TryGetLastEventBefore`) and are untouched — only the block whose method promises "next at or after" changed.

**Follow-up:** this case belongs in a shared `IEventSequenceStorage` contract alongside the sink contract from #3750, so all three implementations are held to it rather than MongoDB alone.

PR #3754, opened 2026-08-18
Hold MongoDB to the sink contract as well

woksin:
# Summary

Internal specs only: the MongoDB sink now runs the shared sink contract instead of keeping a private copy of it. No user-facing changes.

Builds on #3749, #3750, #3751 and #3752 — merge those first.


woksin:
Reviewer context: **top of the stack — merge #3749 → #3750 → #3751 → #3752 → this.** Completes plan Task 1.1's core claim: nine cases, written once, now executed against all three sink implementations.

**MongoDB passes all nine unchanged**, which is the expected and useful result — it is the reference implementation, so the contract's job here is to keep the others honest against it rather than to find fault with it. The four watermark cases it had been keeping privately are now the shared ones.

**Design note:** harness creation became overridable (`protected virtual THarness CreateHarness() => new()`) because a backend needing infrastructure receives it through the constructor — MongoDB takes a container fixture — so the contract cannot construct it. In-memory and SQL stay one-line drivers; MongoDB's are four lines each.

**Deliberately left MongoDB-only:** the three bulk-window cases. Bulk mode is a MongoDB feature (`Storage.Sql` and the in-memory sink implement `BeginBulk`/`EndBulk` as no-ops), so promoting them to the contract would assert behavior the other sinks do not claim to have. That is a real divergence, tracked in the audit, but it is a product decision rather than something to force through a spec.

**Verification:** MongoDB contract cases 9/9, and confirmed discovered rather than silently skipped by running one case by fully-qualified name. Full suites 233 / 149 / 95 / 733; solution Release build 50 projects, 0 warnings.

**One flake worth naming:** the first full MongoDB run had 8 failures, all `MongoNotPrimaryException` in `for_EventSequenceStorage.when_appending_many` — replica-set election, nothing to do with sinks. A re-run was 733/733. This is exactly the kind of rescued flake the ledger added in #3748 is meant to record instead of losing.

PR #3755, opened 2026-08-18
Start the packages beside Arc before publishing them

woksin:
# Summary

Internal publish pipeline only: the packages about to be published are installed into a throwaway application alongside the current Arc packages and started, so a combination that cannot boot is caught before the version exists. No user-facing changes.

### DIFF
--- a/.github/scripts/verify-consumer-smoke.sh
+++ b/.github/scripts/verify-consumer-smoke.sh
@@ -1,0 +1,104 @@
+#!/usr/bin/env bash
+# Copyright (c) Cratis. All rights reserved.
+# Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+# Installs the Chronicle packages about to be published into a throwaway application alongside the
+# published Arc packages that build on them, and starts it.
+#
+# Chronicle never references the Arc packages that integrate with it, so nothing in this repository
+# exercises the combination an application actually installs. Every gate here can be green while that
+# combination fails on the first line it runs - and it fails at startup, in the call every service
+# makes, rather than anywhere a test would look. The only way to see it is from outside, with the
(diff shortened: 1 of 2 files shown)


woksin:
Reviewer context: this is plan Task 0.2, and it comes directly out of the Arc compatibility scare — including the part where the scare turned out not to reproduce.

**The blind spot it closes is structural.** Chronicle references `Cratis.Arc` but never `Cratis.Arc.Chronicle`, so nothing in this repository ever composes itself with the Arc packages that build on it. Every gate we have can be green while that combination fails on the first line an application runs — and it fails inside `AddCratisArcCore()`, on startup, rather than anywhere a spec would look.

**Why it runs rather than restores.** My first version only restored, and it passed — which is how I found out that the NU1109 downgrade errors I had been reasoning from were Arc's own central-package pins, not anything a consumer sees. Restore resolves cleanly and the packages still have to agree afterwards, because type discovery loads them by name and only finds out then. So the gate starts the application.

**Verified both directions, because a gate that cannot fail is worse than none:**
- Current pairing (Chronicle 16.35.3 + Arc 21.19.1): starts, registers 1545 services, exit 0.
- Deliberately incomplete pairing: exits 1 with the `::error::` and the `ContractToImplementorsMap.Feed` → `AddCratisArcCore` stack trace.

Placed before `dotnet nuget push`, since after it the version exists and cannot be withdrawn.

**Cost:** one restore plus one process start against packages that are already built — seconds, on the publish path only.

**Known limit:** Arc floats to whatever is current, so this catches Chronicle moving out from under a released Arc. It does not catch the mirror case of Arc releasing against an older Chronicle — that check belongs in Arc's pipeline, and Arc/#2559 just closed a nine-minor gap of exactly that kind.

PR #3756, opened 2026-08-18
Release one version at a time

woksin:
## Fixed

- Publishing several merges at once no longer loses releases to a race on the next version number: runs queue behind each other instead of computing the same version in parallel

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,12 +1,21 @@
 name: Publish
 
-# Keyed by the pull request, never by a value that several simultaneously-closing pull requests share.
-# Merging a consolidation closes every pull request whose commits it carries, each of those fires its own
-# pull_request:closed event, and on a shared group cancel-in-progress means the last run to start cancels
-# the rest — including the one that was supposed to cut the release.
+# One release at a time, and never cancel one.
+#
+# Deciding the next version is a read followed by a create: the run reads the latest released version
+# and creates the one after it. Two runs doing that at once read the same answer and try to create the
+# same version, so all but one fail - and a burst of them trips GitHub's secondary rate limit, which
+# comes back as a 403 on the create rather than anything that names the real problem. Merging eleven
+# pull requests in two minutes produced exactly that: five runs, all computing v16.35.7, four losing.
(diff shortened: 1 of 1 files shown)


woksin:
Reviewer context — **stacking rather than merging this one: it changes release behavior, which is an operational call, not mine to make unilaterally.**

**How it surfaced, honestly:** I merged eleven PRs in about two minutes this afternoon. Five publish runs failed. I went looking because I had just merged two gates *into* the publish path and assumed I had broken it — I had not; the gates were fine and the failure is in the `release` job, upstream of them.

**The actual mechanism**, from run 32144142667's log:
```
Latest released version: 16.35.6
New version is '16.35.7'
GET  /releases/tags/v16.35.7 - 404      (not there yet)
POST /releases                - 403 Resource not accessible by integration
```
Read-then-create. Every concurrent run read `16.35.6` and tried to create `16.35.7`; one won. The 403 is GitHub's secondary rate limit on the burst, which is why the error names a permission problem that does not exist and sent me looking in the wrong place first.

**Not new, and not caused by my gates:** the same pair of failures happened on 2026-08-13 at 07:01, two runs in the same minute. My burst just made it five instead of two. The per-PR concurrency key means N merges = N parallel publishes by design.

**Nothing was lost:** every merged change shipped — the sink-contract work in v16.35.7, the SQL namespace fix in v16.35.6. Only the intermediate version numbers never came into being.

**The fix and the trade-off it revisits:** the group was keyed per pull request deliberately, because a shared group *with cancellation on* meant the last run to start killed the rest, including the one meant to cut the release — there is a comment in the file about it. That fixed cancellation and bought this race. Sharing the group with `cancel-in-progress: false` avoids both: runs queue, none are cancelled, one version is decided at a time.

**Cost:** releases serialize. With a normal merge cadence that is invisible; with a burst it means waiting rather than failing. That is the trade I would make, but it is the part worth a second opinion — and it is a smaller version of plan Task 2.2 (release train), which proposes going further and batching releases outright.

**Not verified end to end:** concurrency behavior only proves itself on a real burst. The next multi-merge day is the test.

PR #3758, opened 2026-08-18
Give the production image the certificates it requires to start

woksin:
## Fixed

- The image boot gate no longer fails every production publish: it now supplies the TLS and token-encryption certificates the production image requires, instead of starting it in a configuration it deliberately refuses

### DIFF
--- a/Docker/boot-test.sh
+++ b/Docker/boot-test.sh
@@ -23,6 +23,8 @@ TIMEOUT="${BOOT_TEST_TIMEOUT:-180}"
 cleanup() {
     docker rm -f "$SERVER" "$MONGO" > /dev/null 2>&1 || true
     docker network rm "$NETWORK" > /dev/null 2>&1 || true
+    [ -n "${CERT_DIR:-}" ] && rm -rf "$CERT_DIR"
+    return 0
 }
 trap cleanup EXIT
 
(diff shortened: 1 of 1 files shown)


woksin:
Reviewer context: **this is my regression and it is blocking every Docker publish right now — merging it as soon as CI is green.**

**What I got wrong.** The boot gate I added in #3735 starts the production image with storage configured and nothing else. But the production image requires two certificates and refuses to start without either, deliberately and with clear messages:
- a TLS certificate, because the single Chronicle port serves gRPC and HTTP/1.1 together and can only do that over TLS;
- an OpenIddict token-encryption certificate, which is mandatory in production.

So the gate was never testing the image. It was re-proving a documented requirement, and it could not have passed. Eight publish runs in a row failed on it.

**Why I did not catch it before merging:** I verified the gate locally against the *production* image on arm64 and it passed — because at that point I was testing an image I had built from a branch, and I only exercised the failure mode I was hunting (the missing exec bit). I never ran it against a real published production image. That is the gap.

**The fix:** write a throwaway PKCS#12 certificate and point both settings at it. Also made the host port overridable (`BOOT_TEST_HOST_PORT`), because a machine already running Chronicle on 35000 cannot run the test at all — which is exactly why I could not reproduce the CI failure locally until I worked around it.

**Verified against published images, not ones I built:**
- `cratis/chronicle:16.35.3` production → starts and stays up.
- `cratis/chronicle:16.35.3-development` → starts and stays up (it generates its own certificate; left alone).
- A deliberately broken image → still exits 1, so the gate has not been softened into uselessness.

**Note on the other publish failures in the same runs:** `coverage-statistics` and the npm/Hex publishes also failed, from a separate cause — concurrent publish runs racing on the next version number. That is #3756, stacked separately.

PR #3759, opened 2026-08-18
Fail waiting for observer completion when it cannot be determined

woksin:
## Added

- `CannotWaitForObserverCompletion` names the case where completion cannot be established

## Changed

- Waiting for observer completion on an append that carries no observer surface now fails by name instead of reporting success, so an in-process spec can no longer assert that downstream work finished when none of it was started

### DIFF
--- a/Documentation/events/observing-appends.mdx
+++ b/Documentation/events/observing-appends.mdx
@@ -50,6 +50,13 @@ When you need to wait until observers affected by an append operation have compl
 
 If any affected observer fails while catching up, the returned result includes failed partitions.
 
+Waiting only means something when there are observers behind the connection. The in-process testing
+surfaces (`EventScenario` and `EventStoreForTesting`) run the event sequence without a silo, so an
+append there never reaches a projection, reducer, or reactor. Rather than report that everything
+completed, `WaitForCompletion()` throws `CannotWaitForObserverCompletion` on an append result that
+carries no observer surface — assert on the scenario's own surface, or move the check to an
+out-of-process integration spec where observers genuinely run.
+
 ### Append
 
 <ChronicleClientTabs snippet="events/observing-appends/wait-for-completion-append" />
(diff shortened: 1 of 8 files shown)


woksin:
Reviewer context (Opus subagent implemented, I reviewed and re-ran the gates myself).

**The lie being removed.** `AppendResult.WaitForCompletion()` returned `new(true, [])` — success — whenever the result carried no observer surface. In-process that is always: `InProcessEventSequence` builds the kernel `EventSequence` grain by reflection and never calls `OnActivateAsync`, so `_appendedEventsQueues` stays unset and **no projection, reducer or reactor ever runs from an in-process append**. `InProcessServices.Observers` throws `NotSupportedException`, the client caught it and returned null, and null was read as "nothing to wait for". A spec could await completion, assert success, and have verified nothing.

**Why throwing rather than a not-success result** — this was the judgment call and I agree with it: an `IsSuccess == false` is indistinguishable from *an observer genuinely failed*, so it would send a spec author hunting a phantom failed partition, and a spec that reasonably asserted failure would pass for the wrong reason. The exception names the limitation at the moment the caller asks for something the connection cannot answer.

**Real clients are untouched.** Every `AppendResult`/`AppendManyResult` construction site was audited: out-of-process always carries the real `IObservers`; the only production site without one is `UnitOfWork.Rollback`, whose empty sequence numbers make the tail unavailable, which still completes trivially. The `TailSequenceNumber.IsUnavailable` early return is deliberately left alone — an append that never reached the log genuinely has nothing to wait for.

**Specs: three added, each run red against the pre-fix code** (`Assert.IsType() Failure: Value is null — Expected CannotWaitForObserverCompletion, Actual null`, i.e. the wait reported completion) and green after. Each also asserts the append itself succeeded, so it cannot pass by the append merely failing. **No existing spec relied on the false success** — nothing in the repo awaited completion on an in-process append, which is precisely how it survived.

**Also audited, deliberately not changed:** the other fields the reflection leaves unset. `_knownConstraints` and `_statePersistenceInterval` have field initializers and are fine; `_constraintsVersionCheckInterval` being zero is already deliberate; the `null!` meter is safe only because it is dereferenced solely in `OnActivateAsync`, which never runs — that invariant is now documented rather than papered over with a null-object.

**My verification, independent of the agent's:** solution Release build 52 projects 0 warnings 0 errors; `Testing.Specs` 438/438; `DotNET.Specs` 2526/2526.

**Labeled `minor`** for the new public exception type. Making observers actually run in-process was explicitly out of scope and remains open.

PR #3760, opened 2026-08-18
Superseded by #4009: Apply PII compliance in the in-process testing harness

woksin:
> **Superseded — not a merge candidate.**

The useful in-process PII/compliance work was ported to the current generated-service architecture, corrected for generation-specific schemas and provider failures, and shipped through #4009 in v18.1.0. Read-model sink encryption and erasure were not claimed as supported. Do not merge the old branch.

Closure is pending the repository's package-cleanup access safeguards. The branch is retained to avoid losing historical work.

---

## Fixed

- `[PII]` values are now encrypted at rest and released on read in the in-process testing scenarios, instead of passing through in plaintext, so a spec can verify its own compliance markings
- A read model carrying `[PII]` now reports compliance as a substituted layer, so `WithStrictFidelity()` no longer passes a model whose compliance never runs


woksin:
Reviewer context (Opus subagent implemented, I reviewed and re-ran the gates myself). **Stacking rather than merging — see the blast-radius item at the end.**

**The audited defect was only a third of it.** `[PII]` was inert in-process because *three* independent stand-ins each short-circuited compliance, and fixing only the first would have left the specs red:
1. The handler set was empty — `JsonComplianceManager` built over `KnownInstancesOf<...>()` with nothing in it, at 9 construction sites.
2. `EventStoreForTesting` built its schema generator with an empty `ComplianceMetadataResolver`, so **no generated schema carried compliance metadata at all**.
3. `InMemoryEventTypesStorage` answered every schema lookup with `new JsonSchema()`, and the kernel gates compliance on the schema (`Apply` short-circuits on `!schema.HasComplianceMetadata()`).

Each is a textbook instance of the repo's own "a stub that silently succeeds is a bug" rule, and the second and third were not in the audit.

**Key storage choice, deliberate:** `Storage/Compliance/InMemoryEncryptionKeyStorage` over `Storage.InMemory/Compliance/EncryptionKeyStorage` — its `GetOrAddFor` is atomic under a lock (the other is check-then-act outside one), it is the one with spec coverage, and it is what the kernel's own compliance specs exercise. One instance is shared per scenario because the grain encrypts on append and the read path releases: a second key store mints a second key that cannot decrypt the first one's ciphertext.

**A finding worth reading:** `ReadModelScenario` runs **no compliance at all** — there is no `EncryptChangeset` equivalent in `ProjectionReadModelProcessor` and no release, so there was nothing to wire. Building that pipeline is a feature, not a harness fix. Rather than leave it silently absent, the substituted-layer detector now reports it, which is what stops `WithStrictFidelity()` giving a false all-clear on a PII read model. Same gap in `ReactorScenario`'s materialization.

**Specs: 4 files, 9 facts, each verified red-without/green-with by reverting one fix piece at a time and restoring between** — empty handler set → both round-trip specs fail on plaintext found in the store; empty metadata providers → all four fail; empty schema → both round-trip specs fail; detection removed → both read-model specs fail. The value observed at rest is a `CENV` envelope blob.

**The one thing I want a second opinion on:** the harness now hands the kernel the *real* event schema on every in-process append, so schema validation and schema-driven coercion actually run where they previously no-opped. All 443 specs pass, and nothing in this repo consumes `Cratis.Chronicle.Testing` outside `Testing.Specs` — but **downstream apps and Arc's `CommandScenario` do**, and their specs could start failing on markings that were never checked before. That is a correct outcome and still a behavior change worth landing deliberately, which is why I have not merged it. A narrower variant (real schema only when it carries compliance metadata) also passes; it was rejected as a subtler trap.

**Verified independently of the agent:** solution Release build 52 projects, 0 warnings, 0 errors; `Testing.Specs` 443/443 in Release (434 before).

**Still not working in-process and out of scope:** `EventStoreForTesting.PII` crypto-shredding, which resolves an Orleans grain the testing factory cannot provide — so erasure behavior remains unspec'd.

woksin:
Converted to draft during a worktree sweep — this is in-progress work rather than something awaiting merge.

State as of now: the branch is **284 commits behind main** and was last updated 2026-08-26. GitHub still reports it mergeable, but that is stale information — it needs a rebase onto current main and a fresh CI run before the result means anything.

The work itself (4 commits, 24 files, ~529 insertions: wiring real PII compliance into the in-process testing harness, giving the harness real event-type schema and compliance metadata, reporting compliance as a substituted layer for read models, plus specs) is intact and unmerged. Nothing here is lost — it just needs picking back up or deliberately closing.

Marking draft so it stops reading as review-ready.

woksin:
## Closing as superseded — work preserved

The useful PII-testing changes were ported to the current generated-service architecture, corrected for generation-specific schemas and provider activation failures, and shipped in **v18.1.0 through #4009**. The replacement has broader regression coverage. Keeping this older branch open would invite duplicate or obsolete changes. This does not claim in-process read-model sink encryption or erasure support.

The source branch [`fix/harness-compliance-wiring`](https://github.com/Cratis/Chronicle/tree/fix/harness-compliance-wiring) is retained at `43301096de46befe2a51b3e4fbe8054ebb3c7a9c`. No source history is being rewritten or deleted.

The owner approved closing this PR and its automatic cleanup of **11 historical prerelease NuGet versions**. Those exact packages were downloaded and checked for archive integrity, package identity, and SHA-256 hashes before closure. No stable-release packages or container images were identified among the cleanup targets. Cleanup completion will be checked separately rather than assumed from the PR state.

PR #3761, opened 2026-08-18
Assert that every build-affecting path triggers a build

woksin:
# Summary

Internal CI only: the set of paths that must trigger a build and test run is now a committed contract that CI asserts, and the build trigger is widened to cover the thirteen that were not covered. No user-facing changes.

### DIFF
--- a/.github/build-affecting-paths.txt
+++ b/.github/build-affecting-paths.txt
@@ -1,0 +1,77 @@
+# Paths that must trigger a build and a test run.
+#
+# THE RULE: changing this list requires saying in the pull request what stops being covered.
+#
+# This file is a committed contract, not documentation. `.github/scripts/assert-build-affecting-paths.py`
+# reads it, reads the `paths:` filter on `.github/workflows/dotnet-build.yml`, and fails the build when a
+# path listed here is not covered by any pattern in that filter. Narrowing the filter for speed is what
+# this exists to make visible: on 2026-04-17 the trigger was scoped to `.cs`/`.csproj` (47a82ad45), and on
+# 2026-08-05 a `Directory.Packages.props` dependency bump - described in its own pull request as the
+# riskiest change in the repository - merged with no build at all (c9fbc8d99), because a change confined
+# to that file matched none of the remaining patterns. Benchmarks broke while landing green for the same
+# reason (51590103c).
+#
+# FORMAT: one entry per line. Either a literal repository-relative path, or a directory followed by `/**`.
+# Blank lines and `#` comments are ignored. Every entry is verified to exist in the repository, so the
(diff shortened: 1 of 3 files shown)


woksin:
Reviewer context (Opus subagent implemented, I reviewed and re-verified myself). Plan task 8.f.

**The hole this closes is not hypothetical.** Commit `47a82ad45` narrowed the build trigger to `.cs`/`.csproj` for speed. On 2026-08-05 a `Directory.Packages.props` dependency bump — described in its own PR as "the riskiest change in the repository" — merged **with no build at all**. Benchmarks broke while landing green for the same reason. Those were patched by hand; this makes the filter list a committed contract instead, so the next well-meaning narrowing has to say what it drops.

**At HEAD the checker failed on 13 of 17 required paths** — only `Integration/**`, `Directory.Packages.props`, `Directory.Build.props` and `Chronicle.slnx` were covered. Uncovered: `Source/**`, `Benchmarks/**`, `Docker/**`, both per-TFM package props, `Directory.Build.targets`, `global.json`, `.globalconfig`, `.editorconfig`, `specs.runsettings`, `.github/workflows/**`, `.github/scripts/**`, and the contract file itself. **All 13 were closed by widening the trigger, not by weakening the check** — each widening backed by a step in that same workflow. No build or test logic changed; the diff is the `paths:` list plus one step.

**Every entry is evidence-based, and two suggestions were rejected on inspection:** `nuget.config` does not exist in this repo (sources are added imperatively in the workflow), and `integration.runsettings` exists but nothing references it. The contract also rejects listing a path that is not on disk, which is how the first was caught.

**Matching is real glob semantics, not string equality** — proven by re-narrowing `Source/**` back to `Source/**/*.cs` + `*.csproj` (the exact `47a82ad45` shape): the pattern list *grew* and the check still failed. The agent's own test suite caught a genuine bug in its first version, where a `!Docker/Production/**` exclusion carved a hole mid-subtree that a prefix probe stepped over.

**My own verification, not the agent's:** solution Release build 52 projects 0 warnings; checker exits 0 at branch head; removing `Docker/**` from the filter → exits **1** with a message naming the path, the probe that matches nothing, and both remedies; restored → exits 0.

**Two deliberate costs, worth knowing:** `.github/workflows/**` now triggers the full build+integration matrix on any workflow edit (previously two named files) — that is the class of change that removes coverage, so it is the one that should pay for a build. And only `dotnet-build.yml` is asserted; `publish.yml` still lacks `global.json`, `Directory.Build.targets` and the per-TFM props, but widening what triggers a *release* is a heavier call and was left alone.

The plan's Status Board row for 0.4 is not updated here — I will do that separately rather than have every PR touch the plan.

PR #3762, opened 2026-08-18
Make the CodeQL config, corpus validator and header rule actually enforce

woksin:
# Summary

Internal CI only: four checkers that existed but enforced nothing now do — the CodeQL configuration file is applied, the AI corpus validator runs on pull requests, the declared license-header rule has a severity, and a test rescued by a retry leaves a record. Also records the enforcement-audit findings in the program plan. No user-facing changes.


woksin:
Reviewer context: **this is #3748 with the contentious part removed.** Same four commits, minus the one that makes the gRPC gate blocking, plus the plan update that was stuck behind it.

I held #3748 because a newly-blocking gRPC gate could start failing Einar's PRs mid-way through his protocol-compatibility work, and he should see what it does and does not catch before it becomes load-bearing. That reasoning does not apply to the rest, which was sitting behind it for no reason:

- **CodeQL config applied** — `.github/codeql/codeql-config.yml` exists, including a deliberate CA1031 exclusion, and the workflow never passed `config-file`, so every scan ran the unfiltered default suite.
- **Corpus validator runs** — `validate-ai-setup.sh` is a strict exit-1 checker no workflow ran. Wiring it surfaced real drift, fixed here: a rule with no `applyTo`, two prompts missing the `.prompt.md` suffix their adapters resolve through, four agents with no Copilot adapter, two missing Claude command adapters.
- **License header enforced** — `.editorconfig` declares `file_header_template` but IDE0073 had no severity, so it was decoration. Zero violating source files.
- **Flake ledger** — a retry-rescued test previously left a green check and no trace; it is now named in a `::warning::` and appended to `flaky-tests.log` in the results directory CI already uploads.
- **Plan §8** — the enforcement-audit backlog with evidence and per-item definitions of done.

#3748 now carries only the gRPC gate and stays open for Einar.

**Verified after the split, not inherited from the original PR:** solution Release build 52 projects 0 warnings 0 errors; corpus validator exits 0; both touched workflow YAMLs parse; `bash -n` clean on the retry script.

PR #3763, opened 2026-08-18
Catch specs deleted while the code they cover survives

woksin:
# Summary

Internal CI only: a pull request that deletes a spec folder while its subject type stays in the tree now fails a check, so a change cannot quietly remove the coverage that would have caught it. Deliberate removals carry the `specs-removal-intended` label. No user-facing changes.

### DIFF
--- a/.github/scripts/assert-spec-deletions-covered.py
+++ b/.github/scripts/assert-spec-deletions-covered.py
@@ -1,0 +1,196 @@
+# Copyright (c) Cratis. All rights reserved.
+# Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+"""Fails a pull request that deletes a `for_*` spec folder while the type it specifies stays in the tree.
+
+A stale worktree once pushed a change that silently reverted a sibling branch's merged hardening.
+CI stayed green, because the specs that proved the hardening were deleted along with the code it
+guarded - nothing was left to fail. `git log -- <file>` hid it too, since the release merge was
+TREESAME on those paths. Parallel agent branches editing the same files make this a recurring
+shape rather than a one-off, and no existing gate can see it: a deleted spec does not fail, it
+simply stops being counted.
+
+The signal this looks for is the asymmetry that made the incident detectable in hindsight - a
+`for_<TypeName>` folder disappearing entirely while `<TypeName>` itself is still in the tree.
+Deleting a type together with its specs is ordinary housekeeping and is silent here; deleting only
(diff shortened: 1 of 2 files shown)


woksin:
Reviewer context (Opus subagent implemented, I reviewed). Plan task 1.4, the tripwire half.

**The incident it replays.** A stale agent worktree pushed a change that silently reverted a sibling branch's merged compliance hardening (`9404b2b42`, restored by `b1b0e1553`). **CI stayed green — because the specs proving the hardening were deleted along with the code**, so nothing was left to fail. `git log -- <file>` hid it too. Parallel agent branches over shared files make this a class, not a one-off.

**Replayed against the real commit:** the check exits 1 and names it —
`Storage.Specs/Compliance/for_CompositeEncryptionKeyStorage` deleted in full while `CompositeEncryptionKeyStorage.cs` stayed in the tree. The *same commit* also deleted `for_ComplianceEncryptionKeyStorageExtensions` **together with** its type, and the check is correctly silent about that. It separates the accidental revert from the deliberate removal using nothing but the rule. The restore commit exits 0.

**The false-positive analysis is the part worth reading**, because a noisy tripwire gets disabled and then protects nothing. Swept the last **60 merged PRs**, replaying each exactly as the check would see it:

| | |
|---|---|
| PRs examined | 60 |
| deleted ≥1 spec file | 8 |
| reached the final condition | 3 |
| **would have failed** | **0** |

The zero is not vacuous — it reached the last condition on three real PRs and declined all three for the *verified correct* reason: the subject type was deleted alongside its specs. Plus 10 constructed scenarios covering C# and TypeScript flagship shapes, folder moves, renames, partial deletions.

**The design call:** subject resolution is **path-anchored**, not a repo-wide filename search. Chronicle has many same-named types across projects (storage providers per backend, contracts mirrored kernel↔client), so bare-name matching would fire whenever specs for one were retired while siblings remained. The trade is a known false negative — a cross-project spec project whose subject lives elsewhere is not caught — accepted deliberately: a false negative costs one detection, a false positive costs the gate.

**Escape hatch:** `specs-removal-intended`, read from the event payload, documented in the failure text, and the workflow re-runs on label changes. Malformed label JSON **fails closed**. I created the label in the repo so the hatch works on arrival.

**Deliberately not caught,** each a conscious trade: specs gutted rather than deleted (bodies emptied, `Skip` added), partial deletions inside a surviving folder, and non-`for_` test layouts.

PR #3764, opened 2026-08-18
Cover SQLite alongside MongoDB on every pull request

woksin:
# Summary

Internal CI only: pull requests now run the integration suite against SQLite as well as MongoDB, so a storage assumption that only holds on MongoDB fails at review time instead of in the nightly. The scheduled all-providers run is unchanged. No user-facing changes.


woksin:
Reviewer context (Opus subagent implemented, I re-ran the matrices and specs myself). **Do not merge yet — there is a prerequisite, below.**

**The honest answer to the question this task was built on.** I asked whether SQLite would have caught the two defects that broke the nightly for six nights. **It would not have caught either**, and the evidence is specific:
- *Row order*: SQLite scans in rowid (insertion) order, so it agrees with MongoDB. Only **mssql** — clustered-key order — surfaces it.
- *Namespace database*: the fix sits behind `EnsureDatabaseExists`, which returns early for anything that is not PostgreSQL or SQL Server. SQLite creates its file on connect, so the buggy path never executes.

**What it does catch is measured, not theoretical** — and this is the part that matters more: the first nightly after the Aug 17 merge storm has **two deterministic SQLite failures that MongoDB passes**, and the three prior nightlies had zero SQLite failures. So these are **fresh non-Mongo regressions that merged green on Aug 17**, which is precisely the class this gate exists for:
- `Projections 3of4` — `when_renaming_read_model.and_reregistering_the_full_set` (also red on postgresql)
- `for_Reactors 1of2` — `should_have_reactor_observer_be_in_disconnected_state` (red on all three retry attempts)

**The prerequisite: those two shards are red on `main` right now**, so merging this turns every pull request red until they are fixed. I have put an agent on diagnosing them; this should land after they are green, not before. Landing a gate that is red on arrival is how gates get disabled.

**Cost:** +17 jobs per PR (34 → 51), ≈ +54 runner-minutes, ≈ +2 minutes wall clock on the integration stage — measured from real runs, and not the critical path (`dotnet-build` gates at ~10.7m).

**What was deliberately rejected:** adding mssql, the only backend that demonstrably catches the row-order class. It is the slowest leg (median 4.0m, max 9.0m), needs a second container plus SQL Server init and EF migrations, and failed on 3 of the last 4 nightlies — a slow, chronically-red required check gets ignored within a week. Also rejected: an allowlist of "namespaces currently green on sqlite", which rots silently and hides regressions.

**Implementation is a new explicit option, not an inverted flag:** `--databases mongodb,sqlite` names an arbitrary set and fails loudly on an unknown backend; `--all-providers` semantics unchanged and still wins. The generator had **no specs at all** despite being CI-critical — 14 were added, and confirmed non-vacuous by reverting the default and watching the right one fail.

**My own verification:** PR matrix 51 jobs `{mongodb: 34, sqlite: 17}`, scheduled 85 and byte-identical to before, generator specs pass.

**Cheap follow-up left undone:** the pre-pull step runs `docker pull mongo` unconditionally, so each SQLite job pulls an image its fixture never starts — 17 needless pulls per PR, one conditional to fix.

woksin:
Your prerequisite is met, so I have landed this. The evidence, since "green now" is worth being specific about:

**Both shards you named are green on `main`, across every backend.** The nightly all-providers run this morning ([32210036217](https://github.com/Cratis/Chronicle/actions/runs/32210036217), 2026-08-19 02:51 UTC) is **99/99 jobs successful** — `Projections 3of4` and `for_Reactors 1of2` both pass on all five configurations, not just the two this PR adds.

**What the two failures actually were**, since they turned out to be different problems rather than one:
- `Projections 3of4 / when_renaming_read_model` was **`JobStateConverter`** — it did not own derived job state types. Fixed in `683ff0075` and released in **v16.36.1**.
- `for_Reactors 1of2 / should_have_reactor_observer_be_in_disconnected_state` is the observer `RunningState` flap, now **#3778**. It is not fixed — it is understood, filed, and parked behind task 2.1, because the direct fix trades a false positive for a false negative: `WaitTillActive` polls for `Active`, so reporting only settled states hands clients a stale `Active`. That attempt is preserved on `fix/observer-running-state-flap`.

So the second one is a known flake rather than a closed defect. I judged that acceptable to land on: it is green on all three retries across all backends now, it is tracked, and it is a pre-existing condition that this PR neither causes nor worsens — MongoDB is equally exposed to it.

**This branch itself:** merged `origin/main` (it was 27 behind), then **65/65 integration jobs green** at `6ebae9daa`, matrix at the 51 jobs you measured. The `verify` check that was red was not a real failure — the branch predated #3771's `verify-semver-label.yml`, so it did not yet recognise `no-release`.

Your "would it have caught the original two" finding is recorded in the plan under §8-h, including that SQLite misses the namespace-provisioning class because it creates its file on connect. I also noted your leftover `docker pull mongo` on SQLite jobs as an open follow-up.

PR #3765, opened 2026-08-18
Freeze spec sleeps behind a per-file ratchet

woksin:
# Summary

Internal specs and CI only: adds the missing rule that a spec awaits a fact rather than a duration, and freezes the sleeps that predate it per file so the count can only go down. No existing sleep is removed and no timing behavior changes. No user-facing changes.

### DIFF
--- a/.ai/rules/specs.csharp.md
+++ b/.ai/rules/specs.csharp.md
@@ -96,6 +96,26 @@ From `Cratis.Specifications`: `.ShouldEqual(expected)`, `.ShouldBeTrue()`, `.Sho
 - Single-statement assertion lambdas use expression-body form (RCS1021) — `[Fact] void should_x() => result.ShouldEqual(...)`, not a `{ … }` block. Don't break long `should_` lines; don't add blank lines between `should_` methods.
 - **Prefer a concept's sentinel over a raw empty primitive.** When a `ConceptAs<T>` / `EventSourceId<T>` value is expected — as an argument, a mock setup, or an assertion — use its `NotSet` (or `Empty` / `New()`) static rather than a `string.Empty` / `Guid.Empty` / `0` that implicitly converts. It states intent and survives a sentinel-value change (see [concepts.md](./concepts.md)). Reserve raw `string.Empty` for genuinely non-concept `string` values (e.g. asserting a decrypted-to-empty JSON payload).
 
+## Never sleep — await a fact, not a duration
+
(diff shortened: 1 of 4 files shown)


woksin:
Reviewer context (Opus subagent implemented, I re-ran the checker myself). Plan task 8.d — the stopgap for the largest regression category, not the cure.

**Why this class matters:** fix-induced regressions in the timing-coupled observer/projection core are **20 of 73 incidents** in the year's analysis, the single largest bucket. Commit `63cf43e38` says outright that the suite depends on "the implicit 50ms settling buffer", and two rollbacks deliberately reinstated known bugs because a fix's timing shift broke a sibling path. Until now **no rule in the corpus forbade sleeping in a spec** — so this adds the rule and freezes the debt. The cure is deterministic completion signals (task 2.1); this only stops the class growing.

**Nothing was removed or rewritten.** Deliberately: changing timing behavior is 2.1 and a human design decision.

**The inventory was read, not grepped-and-guessed.** All 92 in-scope sites inspected with ±9 lines of context, classified by one test — *does the sleep stand in for a completion signal the code cannot ask for?*
- **67 governing**: 11 bare settles (`Task.Delay(500)` then assert) and 56 poll intervals (`while (!condition) await Task.Delay(50)` — awaiting a fact, but re-checking on a timer instead of a signal).
- **25 legitimate**: slowness knobs in test doubles meant to be observed mid-flight, infrastructure readiness backoffs, deadline arms, and `Task.Delay(1)` interleaving nudges.

The baseline annotates the 21 entirely-legitimate entries, so **an unannotated entry reads as remaining work for 2.1**. There is deliberately no write mode — the ratchet cannot be silently reset.

**Where 2.1 should start**, from the data: per-file counts are flat (max 3), so the concentration is by area. `for_ReadModels` has 13 governing sites (all poll-for-projection), `for_EventStoreSubscriptions` holds 4 of the 11 bare settles, and about a dozen near-identical hand-rolled `WaitTillHandledEventReaches` loops across reactors/reducers/event-sequence would collapse into one signal-based helper.

**All four failure modes verified with transcripts:** passes at HEAD; a sleep added to a baselined file fails naming the file, both counts and the new line numbers; a sleep in a new file fails with the three legitimate exceptions spelled out; removing a sleep passes *and* emits a `::notice::` telling the author to lower the entry. Tree restored clean after each.

**One honest note:** the corpus validator exits 1 on this branch — but byte-identically to a run on the pristine tree with the rule edit stashed. All 7 findings are pre-existing (a rule missing `applyTo`, two prompts lacking the `.prompt.md` suffix, four missing Copilot agent adapters). **#3762 fixes exactly those**, so this goes green once that lands.

**My own verification:** checker exits 0 after merging current main, and the path-affecting-paths contract from #3761 still passes alongside it.

PR #3766, opened 2026-08-18
Bring the program status board up to date

woksin:
Documentation only: records what landed today against the reliability program, what is deliberately held and why, and two lessons worth carrying forward.


woksin:
Documentation-only, so no semver label — a docs change should cut no release, and the missing label is the correct state rather than something to fix.

Issue #439, opened 2022-08-30
Add branch protection


einari:
We want branch protection for **main**, but at the same time allow for our GitHub action to be able to commit and push changes to the `CHANGELOG.md` file.

Figure out if we can use the feature to allow certain actors to override the protection:
![image](https://user-images.githubusercontent.com/134365/187371091-0a2cb6bc-91d8-4cdd-b9f9-a053c67ff7ad.png)

PR #3767, opened 2026-08-18
Run the full out-of-process matrix on hot-core changes

woksin:
# Summary

Internal CI only: a change touching the parts of the kernel where fixes historically break sibling paths now runs the full out-of-process integration matrix, and those paths gain code owners. A change touching none of them passes the gate immediately. No user-facing changes.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,51 @@
+# Copyright (c) Cratis. All rights reserved.
+# Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+# Review ownership of the hot core.
+#
+# Only the hot core is listed. Ownership over the whole repository would put a maintainer in front of
+# every documentation typo and would be routed around within a week; this file exists so that the
+# areas where a fix breaks a sibling path - and where an autofix bot has twice committed a mechanical
+# rewrite that broke the build outright (`c237e5e74`, `aedafe1bc`) - cannot change without a
+# maintainer looking at it.
+#
+# The paths below mirror `.github/hot-core-paths.txt` one for one, and
+# `.github/scripts/detect-hot-core-changes.py --verify` fails the build if they ever stop matching.
+# That script runs on every pull request, so the two contracts cannot drift in silence.
+#
+# PLACEHOLDER-OWNER: the owner below is a placeholder and does not resolve. A CODEOWNERS entry whose
+# owner is not a real user or team that has read access to this repository is ignored by GitHub - the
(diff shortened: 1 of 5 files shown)


woksin:
Reviewer context (Opus subagent implemented, I reviewed and re-verified). Plan tasks 1.3 and the in-repo half of 0.6. **Stacking rather than merging — two reasons at the end.**

**Why this gate:** fix-induced regressions in a handful of kernel areas are **20 of 73 incidents**, the largest category, and twice a rollback deliberately reinstated a known bug because the fix broke a sibling path. The out-of-process matrix is the only oracle that has reliably caught it. This runs that matrix when — and only when — a PR touches those areas; everything else passes the gate immediately.

**A finding worth keeping: the plan's own path list was stale.** `Source/Kernel/Grains/Observation/**` does not exist — there is no `Source/Kernel/Grains/` at all. The observation grains live in `Source/Kernel/Core/Observation/` (214 files, including `Observer.Handling.cs`). Had this been implemented from the plan without checking, the gate would have watched a directory that isn't there and reported green forever — the exact failure this program keeps finding. Every one of the seven patterns is now evidence-backed, and the plan is corrected.

Scope: ~415 of 10,209 tracked files (~4%). Spec trees deliberately excluded — they change constantly and cannot break a sibling runtime path, and including them would make the expensive matrix the default.

**Detection verified on real merged diffs, both directions:** `712312afa` (KeyResolvers) → fires, and correctly counts 3 hot-core files while ignoring the 8 changed spec files; `d872a5cec` (the timing ratchet, `.ai` + `.github` only) → passes immediately; `b3e03dbb8` ("Revert Observer.Handling…") → fires. It also does not self-trigger on its own diff.

**Two reasons I am not merging it:**
1. **CODEOWNERS enforces nothing yet.** The owner is `@Cratis/kernel-maintainers`, a placeholder — GitHub silently ignores a rule whose owner is not a real team with read access, so the file parses and requires nothing. No individual usernames were invented. The detector emits a `::warning::` on every PR until the marker is removed, so it cannot rot unnoticed. It also needs "Require review from Code Owners" in branch protection, and `hot-core-gate` marked required — both repo settings I cannot make (#439).
2. **The heavy path has never executed.** The gate does not fire on its own diff, so CI here proves the skip path only. The build job and the matrix invocation are unproven until the first real hot-core PR.

**My own verification:** `dotnet-build.yml` and `integration.yml` are byte-identical to main (`git diff --quiet`); `hot-core-gate.yml` parses and passes actionlint with no findings; the detector's `--verify` exits 0 and emits exactly the one placeholder warning; all four existing assertion scripts still pass alongside it.

Issue #2908, opened 2026-03-23
Refactor to use Arc Commands, Queries and command->Event Hook up (Functional)


einari:
Depends on #2884 

## Goal

The goal is to get rid of the API project  and most of the gRPC Service implementation we have and replace it with Commands and Queries written using the model-bound approach from Arc. This should be done side-by-side to existing code in Core, but hooked up in a way using the gRPC Code Generator done in #2884 - the PR #2885 (https://github.com/Cratis/Chronicle/pull/2885).

While doing the transformation we want

Arc Commands: https://www.cratis.io/docs/Arc/backend/commands/model-bound/index.html
Arc Command Validation: https://www.cratis.io/docs/Arc/backend/commands/validation.html#fluent-validation
Arc Queries: https://www.cratis.io/docs/Arc/backend/queries/model-bound/index.html

The exception we have are the Reducers and the Reactors, they need custom implementations in their services. But for everything else, be it Commands and Queries.

All Command responses should a derivative of the CommandResult - gRPC version.
We should do the same for Queries as well; QueryResult - gRPC version.

## Scope

- Switch the gRPC Service layer in favor of Commands & Queries based on the Arc approach
- Add FluentValidation based CommandValidator<> for all commands
- Generate Contracts project from Commands & Queries
- Some gRPC services like Reducers and Reactors remain as they are, since they are streaming from the server to the client.
- Get rid of API project

## High Cohesion

- Commands and queries are placed in the location of where they belong - next to Grains and internals (Core)

PR #3768, opened 2026-08-18
Gate wire compatibility and fix the contracts that carried no data

einari:
### Added

- gRPC wire compatibility gate that runs before anything else builds, checking the contract at HEAD against every released minor of the current major and reporting each one. A major with no declared floor is compared in full. (#3873)
- `CheckCompatibility` on the connection service, so the kernel performs the one compatibility check for every client language instead of each SDK reimplementing it. A refused connection names the client version, the protocol version and the server version.
- The canonical descriptor set now ships inside the .NET, Kotlin, TypeScript and Elixir contracts packages, so a client hands the kernel exactly what it was built against.
- `ProtocolVersion` on the contracts, and `ServerVersionInfo.ProtocolVersion` on the server.
- `AllPatterns` and `AllPatternScopes` queries, so a namespace's behavior patterns can be browsed without choosing a scope first. (#3857)
- Cross-cutting existence checks for event store, application and user: a command carrying one of those concepts is rejected when the thing it names does not exist, with no per-command wiring. A command that *creates* the entity opts its own property out.

### Changed

- Every gRPC service is now derived from Arc `[Command]`/`[ReadModel]` artifacts in Core rather than hand-written, completing the migration #2908 started. `Source/Clients/Api` and `Source/Kernel/Core/Services` are gone; only `Observers` and the streaming services, whose server-to-client lifetime no command or query describes, remain hand-written.
- Patterns is served by artifacts like everything else. Its five queries — `MatchingPatterns`, `UsualActions`, `PatternsForScope`, `AllPatterns`, `AllPatternScopes` — read storage and the mining engine directly, and the contract is generated from them. (#3857)
- Query parameters that name a concept are declared as that concept rather than as `string`, so the value is validated on the way in instead of being cast inside the method. The wire is unchanged — a concept still travels as its primitive.
- The two event-sequence lookups that answered with a bare `bool` and `ulong` now answer with `EventSourceEvents` and `EventSequenceTail`. The value travels as the same primitive, one level in.
- The Workbench's surfaces are served by Arc artifacts in the kernel, with proxies generated beside the features that use them rather than into a separate `Api` folder.

### Fixed

- `DateTimeOffset` reached the wire as an empty message, so ten fields — job creation times, user timestamps, event query date ranges — carried no value to Kotlin, TypeScript and Elixir clients. .NET was unaffected, which is why it went unseen.
- The proto generator aborted for 22 of its 23 packages while exiting 0, so most generated contracts had silently stopped regenerating and an entire `SequenceQueries` service had never been emitted at all.
- Generated proto files and the descriptor set are now deleted and rewritten on every build, so a hand-applied edit to a generated file cannot survive.
- `AppendMany` was silently dropping the caller-supplied `Occurred` timestamp for every event in the batch.
- Pattern capture re-subscribes when registering event types introduces something new. A server starting against a store no client had connected to yet captured nothing at all until the next restart. (#3857)
- Failed-partition attempts carry their failure kind again, so a congestion timeout can be told apart from a subscriber that genuinely failed.
- An optional concept parameter generated an unguarded cast, so an absent value on the wire did not stay absent in the domain.

> [!IMPORTANT]
> **This release breaks the wire contract against every 16.x release before it**, not just the narrow `DateTimeOffset` case above. Finishing #2908 moved the remaining hand-written gRPC services (`EventSequences`, `EventTypes`, `Captures`, `ExternalServices`, `Identities`, `Observation`/`Webhooks`, `Recommendations`, `Seeding`, `Patterns`, and the rest of the Web API surface) onto the same generated path that `Security` and `Jobs` moved to in 16.34.0 — same cause, same shape of break: renamed methods, response payloads wrapped in Arc's result envelope, some services moved proto package.
>
> This ships as a **major** release. Upgrading the client package(s) alongside the server is enough for
> application code — nothing about how you call Chronicle changes — but two published packages change
> shape, which is what makes it major rather than minor:
>
> - **`Cratis.Chronicle.Api` is no longer published.** Its 142 public types were the hand-written Web API
>   surface; that surface is now generated by Arc from the `[Command]`/`[ReadModel]` artifacts in the
>   kernel, so the running HTTP API is preserved, but the package itself is gone.
> - **`Cratis.Chronicle.Contracts` is regenerated** — 300 public types become 328. 66 old names disappear;
>   all of them have a renamed or relocated successor except the saved-sequence-query contracts, which
>   have none because that surface is served over HTTP now and its only consumer was `Cratis.Chronicle.Api`.
>
> Every other shipped .NET package is unchanged: `Cratis.Chronicle`, `Connections`, `AspNetCore`, `Aspire`,
> `Testing`, `XUnit.Integration` and `Concepts` have byte-identical public API, with one exception —
> `PatternConverters.ToClient` takes `BehaviorPatternDetailsResponse` where it took `Pattern`.
>
> `WIRE_BASELINE_FLOOR` carried the gate through the rest of 16.x once the conversion above finished; this PR
> removes it. It named the release this change would go out as, and this is that release — the major bump
> means there is no later 16.x to carry the floor toward, so the value is cleared rather than moved again.
> The gate now compares every major in full by default, floor or not, which was always the intended behavior
> once a major actually shipped.
>
> A `v17.0.0` tag already exists — an accidental release cut on 2026-08-25 from a commit 125 behind
> `v16.42.0` — and `--sort=-v:refname` still ranks it first. That tag needs retagging or deleting before this
> PR's release actually publishes as `17.0.0`, or the release step will collide with it; it is unrelated to
> the gate itself, which reads the baseline major from the release version being published, not from tags.


einari:
We already had this tool in place, but it has been removed for some strange reason but something. Annoying. 
Anyways, bringing back what we had with the following issue: https://github.com/Cratis/Chronicle/issues/3873

einari:
OMG.. I closed the wrong PR by accident - I thought I was in the context of something else.. Almost lost all of this work.

PR #3769, opened 2026-08-18
Revert the Docker boot gate that blocked every production publish

woksin:
## Fixed

- Production Docker images publish again — none were published for `16.35.4` through `16.36.0`, and the `latest` tag stayed on `16.35.3` (#3770)
- Job state now round-trips correctly on SQL storage backends (SQLite, PostgreSQL and MSSQL) — a job's request payload was lost when its state was written and reading it back failed, leaving projection retirement silently incomplete. MongoDB was unaffected

## Removed

- The pre-publish Docker boot check


woksin:
Reviewer context, kept out of the body since that publishes verbatim as release notes.

**What was verified**

- `publish-docker-production` failed on all 12 runs since the gate landed, every one at "Boot the image before publishing", with the container reporting `No TLS certificate is configured`.
- Docker Hub confirms the gap: last unsuffixed tag is `16.35.3` (06:51), while `-development`, `-development-slim` and `-workbench` all published through `16.36.0`.
- The image is fine. I published the kernel for `linux-arm64`, built `Docker/Production/Dockerfile` from `main`, and ran the boot script against it — exit 0, listening, stays up.
- The env vars the script passes match `ChronicleContainerImageTags.TlsCertificatePathEnvironmentVariable` and `...EncryptionCertificatePathEnvironmentVariable` exactly, so this is not a key-name mismatch. `CertificateLoader.LoadFromTls` returns null only when TLS is disabled or `File.Exists` is false, and TLS defaults to enabled — so the file is not visible to the process.

**The unproven part**

The likely mechanism is that `mktemp -d` creates a `0700` directory owned by the runner's user while the chiseled runtime image runs as UID 1654 and cannot traverse it; `chmod 0644` on the file does not help if the directory cannot be entered. On a Linux-owned Docker volume that reproduces exactly — UID 1654 gets `NOT READABLE`. But an end-to-end run against the real production image passed with the directory at both `0700` and `0755`, because Docker Desktop on macOS does not enforce Linux mount ownership. So the mechanism is plausible and unconfirmed, which is why this reverts rather than fixes forward.

**Rebuilding it**

The candidate fix is `chmod 0755` on the directory plus a preflight assertion that fails with "boot test is misconfigured, not the image" — I have that written and it rejects `0700`/`0600` and accepts `0755`/`0644`. It is deliberately not in this PR: two versions of this gate have now shipped in a state where they could not pass, and it should not come back until it has been shown to fail against an image that genuinely cannot start, on Linux.

woksin:
## Reviewer context

Two independently verified changes were merged into this one PR deliberately, so that a single merge cuts a single release rather than two.

**1. Revert of the pre-publish Docker boot gate** (`36c393b7f`)
The gate failed on every publish run, so no production image has shipped since `16.35.3` — `16.35.4` through `16.36.0` have no images and `latest` is stuck three versions back. Reverting it unblocks publishing. Touches `.github/workflows/publish.yml`, `Docker/boot-test.sh` and the reliability-program plan note.

**2. `JobStateConverter` owns derived job state types** (`683ff0075`)
The converter only handled the exact `JobState` type, but job grains persist *derived* state types. On SQL storage backends the serializer therefore dropped the `Request` payload on write and threw on read-back. The visible symptom was projection retirement never completing, silently. MongoDB uses a different serialization path and was unaffected. Touches `Source/Kernel/Core/Setup/Serialization/JobStateConverter.cs` plus new specs under `Source/Kernel/Core.Specs/Setup/Serialization/for_JobStateConverter/`.

The two changesets touch disjoint files; the merge (`b2c02585a`) applied cleanly with no conflicts.

**Gates re-run locally on the merge commit**

| Command | Result |
|---|---|
| `dotnet build Chronicle.slnx -c Debug -p:DisableDockerBuild=true -p:DisableProxyGenerator=true` | Build succeeded — 0 Warning(s), 0 Error(s) |
| `dotnet build Chronicle.slnx -c Release -p:CratisProxiesOutputPath= -p:DisableDockerBuild=true` | Build succeeded — 0 Warning(s), 0 Error(s) |
| `dotnet test Source/Kernel/Core.Specs/Core.Specs.csproj -c Debug --filter "FullyQualifiedName~for_JobStateConverter"` | Passed — Failed: 0, Passed: 11, Skipped: 0, Total: 11 |

PR #3771, opened 2026-08-18
Stop releasing changes that ship nothing

woksin:
## Added

- `no-release` is accepted as a fourth release-intent label alongside `major`, `minor` and `patch`. Merging a pull request that carries it publishes nothing, and its checks stay green
- A pull request touching the kernel paths where fixes have historically broken sibling behavior now runs the full out-of-process integration matrix; one touching none of them passes the same check immediately
- Those kernel paths are a committed contract with code owners, so review is requested automatically when one of them changes, and a pull request fails if the paths and the owners stop matching

## Changed

- A pull request that changes nothing a consumer compiles against, runs or observes now carries `no-release` rather than a version label. Documentation, CI, tooling and spec-only changes are named explicitly, with the test stated as outward-facing effect rather than file location
- Exactly one release-intent label is required. Omitting the label is treated as a forgotten label rather than a decision not to release, and `no-release` alongside a version label is rejected
- Related small changes belong in one pull request rather than one each, since the pull request is the release boundary


woksin:
Reviewer context — not release notes.

## What was batched in

The regression-tax dashboard (reliability-program task 2.4) was merged in here rather than opened as its own pull request, because it ships nothing to consumers and the pull request is the release boundary. Five commits came across on a merge commit, no rebase or squash:

- `2449398a` Measure the regression tax from git history and the issue log
- `b8e22bb9` Publish the regression-tax dashboard on the first of the month
- `0fd7712d` Seed the regression-tax dashboard with the year to July 2026
- `d7dd3af5` Record task 2.4 as built and what its definitions had to change
- `d8cba376` Merge instead of rebase when the dashboard commits itself

It adds `.github/scripts/regression-tax.py`, `.github/scripts/regression_tax_metrics.py`, `.github/workflows/regression-tax.yml` and `Metrics/`. None of it is compiled, packaged or run by anyone consuming the framework, which is why this pull request stays `no-release` with the dashboard in it.

## Conflict resolution in `PLAN-reliability-program.md`

Both branches edit that file, in different regions, so git auto-merged it. Both sides were checked against the merge result rather than trusted:

- From the dashboard branch — Status Board row 2.4 moves `todo` → `in review`, plus the new "As built" section under §5 recording the two refined metric definitions and the unreconciled 90-of-304 issue figure.
- From this branch — the §2 rewrite of the release-label working agreements.

Neither supersedes the other and both are present in full.

One follow-up commit was needed on top (`82fe1e2c`): §2 was written by `c07f7045`, *before* `0c8b9b01` introduced the `no-release` label, so it still told a session to leave the label off and to expect a red version-label check. That is now exactly what the rule classifies as an omission rather than a decision, and the check it described passes. §2 now matches `.ai/rules/pull-requests.md`. Drop that commit if the drift is better fixed separately.

## Verification after the merge

- `python3 .github/scripts/regression-tax.py --month 2026-07 --no-issues` — exit 0. Reproduces the verdicts recorded in the plan: commit volume 710 PASS, fix/revert share 9.9% PASS, patch share 56.8% FAIL, reverts 5 FAIL.
- `python3 -c "import yaml; yaml.safe_load(...)"` on `regression-tax.yml`, `verify-semver-label.yml` and `publish.yml` — all three parse.
- Fail-loudly check against a `--depth 1` clone — exit **1**, every metric rendered `n/a` with "the clone is shallow, so check it out with fetch-depth 0" as the stated reason and both sources listed under "Unread sources", plus two `::error::` annotations. No zeros were reported for unread sources.

woksin:
Folded #3767 (`feat/hot-core-review-and-gate`) into this branch so the reliability-program work that ships nothing lands as one pull request rather than several. Both were already labeled `no-release`, so the label is unchanged.

**How it was folded in** — `git merge origin/feat/hot-core-review-and-gate`, no rebase, squash or cherry-pick. The merge was **conflict-free**: `git merge-tree --write-tree --messages` reported `Auto-merging PLAN-reliability-program.md` and nothing else, and the real merge did the same. The two branches' hunks in `PLAN-reliability-program.md` are adjacent but disjoint, and both sides survive — diffing the merge against each parent shows only that parent's counterpart changes:

- from #3767: Task 0.6/1.3 status paragraphs, the corrected hot-core path list (`Source/Kernel/Core/Observation/**`, not `Source/Kernel/Grains/Observation/**`), and Status Board rows 0.6 and 1.3
- from this branch: the §2 release-label working agreements, the Task 2.4 "As built" section, and Status Board row 2.4

**Commits now on the branch** (against `main`):

```
d68aa7648 Add CODEOWNERS over the hot core and its path contract
a06b8a132 Run the full out-of-process matrix on hot-core changes
6ff74bf71 Record hot-core gate and CODEOWNERS status in the program plan
9fdfe993b Own the hot core with a team that actually resolves
2449398ad Measure the regression tax from git history and the issue log
b8e22bb9a Publish the regression-tax dashboard on the first of the month
0fd7712d1 Seed the regression-tax dashboard with the year to July 2026
d7dd3af53 Record task 2.4 as built and what its definitions had to change
d8cba3766 Merge instead of rebase when the dashboard commits itself
e199f9542 Stop releasing changes that ship nothing
c07f70457 Correct the plan's own instruction to label infra work patch
0c8b9b012 Make not releasing a decision the pipeline can tell from a mistake
82fe1e2ce Teach the plan the no-release label it now has
```

**Verification after the merge**

| Check | Result |
|---|---|
| `python3 .github/scripts/detect-hot-core-changes.py --verify` | exit 0 — `7 hot-core patterns all match tracked files and all have an owner in .github/CODEOWNERS.` |
| `yaml.safe_load` on `.github/workflows/hot-core-gate.yml` | parses |
| `yaml.safe_load` on `.github/workflows/verify-semver-label.yml` | parses |
| `yaml.safe_load` on `.github/workflows/regression-tax.yml` | parses |
| `python3 .github/scripts/regression-tax.py --month 2026-07 --no-issues` | exit 0 — 2026-07 row: 710 commits, 9.9% fix/revert share vs 10.6% published (delta -0.7), 5 reverts, 44 releases, 56.8% patch share |
| `.github/CODEOWNERS` ownership | every rule owned by `@Cratis/core`; that is the only owner in the file |
| `PLACEHOLDER-OWNER` in `.github/CODEOWNERS` | 0 occurrences |

#3767 is left open and unmerged — close it once this one lands, since its commits are reachable from here (`git merge-base --is-ancestor 9fdfe993b HEAD` exits 0).

One thing worth a reviewer's eye: the regression-tax dashboard commits (Task 2.4) were already on this branch before the fold and are not described in the body. Left as-is rather than widened here.

woksin:
**Correction to the folded-in #3756 change: it does not serialize releases, and the comment now says so.**

The comment #3756 added claimed every publish run shares one concurrency group and queues behind the others. It does not. `publish.yml` triggers on `pull_request` (`types: [closed]`), and under a `pull_request` event `github.ref` is `refs/pull/<N>/merge` — unique per pull request. The previous key, `${{ github.event.pull_request.number || github.ref }}`, was unique per pull request too. So the group is per-PR both before and after the change, and the only thing that actually changed is `cancel-in-progress: false`. Checked against real runs: the last six `publish.yml` runs all have `event=pull_request` with distinct head branches.

`cancel-in-progress: false` is worth keeping — it guarantees a publish run, once started, always finishes and cuts its version. But it does nothing about the failure this PR was named for: deciding the next version is a read followed by a create, and concurrent runs still read the same latest version and race to create it. That is what produced the 403s from secondary rate limiting on 2026-08-18.

**The obvious fix is a trap, so the comment now warns against it.** A constant concurrency group would be worse than the current state, not better. GitHub keeps at most one *pending* run per group: when a run is in progress and a second is pending, a third arrival cancels the pending second. Merging three pull requests in quick succession would silently cancel the middle one's publish run — its version never cut, and nothing reporting a failure. That trades a noisy 403 for a silently dropped release.

This commit therefore changes **no YAML values** — `group` and `cancel-in-progress` are byte-for-byte what #3756 set. It only rewrites the comment to state what the configuration does and does not do, and records the residual risk under task 2.2 in `PLAN-reliability-program.md`, which owns the real fix: retry version creation with backoff, or decouple release from merge so only one release run exists at a time. The plan's session log no longer claims #3756 serializes releases.

Issue #3715, opened 2026-08-15
An older client registering a unique event type constraint crashes constraint registration on an upgraded kernel


woksin:
## What happens

`Contracts.Events.Constraints.UniqueEventTypeConstraintDefinition` carries `[ReservedProtoFields(1)]`: the scalar `EventTypeId` on field 1 was replaced by the repeated `EventTypeIds` on field **2**.

A client older than that change registers a unique-**event-type** constraint and writes its event type to field 1. An upgraded kernel has field 1 reserved, so protobuf-net skips it, and `EventTypeIds` deserializes as **null**. `Source/Kernel/Core/Services/Events/Constraints/ConstraintConverters.ToChronicle` then does:

```csharp
constraint.Definition.Value1!.EventTypeIds.Select(...)
```

— a `NullReferenceException`, which takes down constraint registration **for the whole event store**, not just the offending constraint.

## Measured, not reasoned

Real `protoc`-generated types, actual bytes, old shape → new shape:

| Encoding | Read as | Result |
|---|---|---|
| old scalar on field 1 | current shape (repeated on field 2, field 1 reserved) | **`null`** |
| old scalar on field 1 | hypothetical repeated **on field 1** | `[ProjectInitialized]` |

The second row is the point: had the field been **widened in place** rather than moved, this would have been fully wire-compatible. In proto3 a singular field and a `repeated` field of the same non-packed type share a wire representation, so a one-element repeated field and a singular field encode to identical bytes.

## Why this is being reported now

The identical mistake was about to ship for `Constraint.RemovedWith` in #3713 — field 3 reserved, plural added at field 6. It was caught in review and fixed by widening field 3 in place instead, which measures byte-identical to the old encoding for the single-value case. Auditing the rest of the contracts for the same shape turned this one up, already on `main`.

⚠️ **It is on the documented upgrade path.** The #3600 release note instructs operators to *"upgrade the Chronicle kernel before upgrading any client"* — which is precisely the old-client/new-kernel window where this fires.

⚠️ **Nothing catches it.** `Source/Clients/Connections/CompatibilityValidator.cs` compares services and RPC method signatures only, never message field shapes, so the client connects cleanly and then fails at registration.

Failing loudly with an NRE is better than the silent empty-collection outcome `Constraint` would have had — but it is still a hard failure on a supported upgrade path, from a change that did not need to break anything.

## What to decide

[...]

PR #3775, opened 2026-08-18
[Recovered, do not merge as-is] The gRPC wire-contract gate that never landed from closed PR #3748

woksin:
## Recovered work — not a live proposal

This is **abandoned work republished for preservation**, opened as a draft so it is discoverable. Nobody is asking for a merge here. It is the one commit from the closed PR #3748 that never reached `main`.

**This PR deliberately does not follow the repository pull request template.** That template feeds release-note generation, and recovered/abandoned work must never generate release notes. There are therefore no `Added` / `Changed` / `Fixed` sections here, by design.

**There is no semver label on this PR, and that is the correct state.** A missing label means no release is cut, which is exactly what should happen for a preservation branch. The `verify`-style check will fail because the label is absent. **That failure is expected and must not be "fixed" by adding a label.**

## Why this one matters more than its size

`main` today has a gRPC contract check that **cannot fail a pull request**.

The `Compare schemas` step in `.github/workflows/grpc-compatibility.yml` on `main` runs under `set +e`, records the verdict into `has-breaking-changes`, and then exits `0` regardless. Nothing consumes that output as a gate on the `pull_request` trigger. A breaking wire change is detected, written to a step output, and merged.

This is not hypothetical for this estate:

- Proto field renumbering has broken clients **twice** (`Constraint.RemovedWith`, since fixed; and Chronicle #3715, still open).
- `CompatibilityValidator` checks **methods, not fields**, so it does not catch that failure mode either.

So the specific class of failure that has actually bitten this repo has **no enforcing gate on `main` right now**. That gap is what the commit below closes.

## What the unique commit does

`0382868af` — *Fail pull requests that break the gRPC wire contract*

- Rewrites the compare step to `set -euo pipefail` and `exit 1` when the schema comparison reports a break, so the job actually fails the PR.
- Adds a deliberate escape hatch: the `grpc-breaking-change-approved` label on the PR downgrades the failure to a warning. Labels are read from the event payload (`github.event.pull_request.labels.*.name`), so no token and no API call are needed.
- Emits `::error::` annotations naming the detected changes and telling the author exactly which label to add.
- Also removes `.github/scripts/update-pr-description.sh` and trims `.github/scripts/README.md` accordingly. **Note this deletion** — it is part of the same commit and is a real change a reviewer should look at, not incidental.

## Verification performed

- Branch head `32babc849`, already on the remote; nothing was force-pushed, rebased or amended.
- Five of the six commits are on `main` (subject match, one hit each): the CodeQL config, the AI corpus validator, the license-header enforcement, the retry-rescued-tests record, and the enforcement-audit plan entry.
- `0382868af` has **zero** subject matches on `main`. A control subject known to be on `main` returned one hit, so the search itself was working.
- The three-dot diff against `origin/main` reduces to exactly three files, because the five landed commits wash out against `main`:

```
M	.github/scripts/README.md
D	.github/scripts/update-pr-description.sh
M	.github/workflows/grpc-compatibility.yml
```

## How far behind main

31 commits behind `origin/main` (`c8d8ce85e`). Merge base `ccdbd63b2`.

## Why it stalled

PR #3748 (*Make the existing CI checkers actually enforce*) was closed on 2026-08-18 without merging. Its other five commits reached `main` by another route; this one did not travel with them.

Bare reference, no closing keyword: #3748, #3715.


woksin:
## Honest assessment

**Still applies — and this is the one item in this preservation sweep worth actually reviving.**

The gap is real and present on `main` today, not a stale observation. I read `origin/main`'s `.github/workflows/grpc-compatibility.yml` directly: the compare step still runs under `set +e`, still writes `has-breaking-changes` to `$GITHUB_OUTPUT`, and still exits `0` on the `pull_request` trigger. Detection without enforcement. A PR that renumbers a proto field goes green.

Combined with the two facts that make this estate specifically vulnerable — proto field renumbering has broken clients twice, and `CompatibilityValidator` inspects methods rather than fields — the enforcing gate is genuinely missing rather than merely duplicated somewhere else.

**What has been superseded:** nothing in the unique commit. The other five commits from #3748 are already on `main`, which is why the diff here collapses to three files. There is no overlap left to untangle.

**Caveats a reviewer should weigh before reviving:**

1. **The branch is 31 commits behind `main`.** The workflow file has changed on `main` since the merge base (`main`'s copy carries newer concurrency-group and base-ref checkout comments that this branch predates). This is therefore **not a clean cherry-pick** — the enforcement logic needs re-applying on top of current `main`, not a merge of this branch as-is. That is the main reason I have left it as a draft rather than suggesting it go straight in.
2. **The script deletion rides along.** `0382868af` also deletes `.github/scripts/update-pr-description.sh` and rewrites `.github/scripts/README.md`. That is unrelated to the wire gate and should be judged separately — quite possibly dropped — rather than inherited silently.
3. **The escape-hatch design is a policy choice, not just code.** The `grpc-breaking-change-approved` label is asserted in a comment to imply the `major` semver label, but nothing in the workflow enforces that pairing. If the gate is revived, someone should decide whether that implication needs its own check or is fine as convention.

**Recommendation:** treat this draft as the specification and the evidence, not as the merge candidate. The most useful outcome is a fresh, small PR against current `main` that re-applies only the enforcement change to the compare step, leaving the script deletion out. Preserved here so the reasoning and the exact workflow text are not lost a second time.

Nothing was deleted, rebased, amended or force-pushed to produce this.

einari:
This is functionality we had, but for some strange reason was removed - bringing it back with issue #3873

PR #3786, opened 2026-08-19
Check pull request bodies and shipped public API coverage before they can regress

woksin:
## Added

- Pull request bodies are checked against the release-note rules before they can publish, since the body becomes the GitHub release note verbatim — a heading outside the sanctioned set, a declared-but-empty section, an issue-closing keyword, or agent-transcript residue now fails the pull request instead of reaching the releases page
- A public type in a shipped package that no spec covers a single line of now fails the build, read from the coverage the specs job already collects rather than from a name grep, with a committed baseline of the types that are uncovered today

## Changed

- `.ai/rules/pull-requests.md` records that the description rules are machine-checked, why the heading rule is an allowlist rather than a denylist, and that internal detail belongs in a pull request comment rather than in a published note
- `PLAN-reliability-program.md` records what tasks 8.b and 8.g actually involve, measured rather than estimated, including why 8.b's specified first tier would have reported fully covered code as uncovered


einari:
@woksin Is this still valid?  2 week old

PR #3841, opened 2026-08-26
Trigger Chronicle releases from main pushes

woksin:
## Fixed
- Trigger Publish from accepted commits reaching `main` instead of relying on a missing `pull_request: closed` event.
- Resolve the merged pull request, release label, and release notes from the merge commit through the existing release action.
- Keep each commit in a distinct non-cancelling concurrency group.
- Preserve explicit failure when a merged PR has no release-intent label.

## Evidence
- Chronicle PR #3833 merged with `patch` and changed `publish.yml`, but GitHub created no Publish run.
- A closed-unmerged path-matching probe PR also created no Publish run.
- A branch-only `validate_only` workflow dispatch instantiated the complete current graph and skipped every release/publish job, proving the graph is valid and the closed event is the missing boundary.
- `actionlint .github/workflows/publish.yml .github/workflows/publish-python-client.yml`

## Effects
- This workflow-only PR is `no-release`. Merging it will exercise the restored main-push trigger, but the release action must resolve the PR's `no-release` label and skip every publication effect.
- Chronicle PR #3833 still requires a separate manual dispatch after the owning human chooses the exact version.
- Chronicle currently has a release-profile split: GitHub/OCI latest is 16.38.2 while NuGet `Cratis.Chronicle` is 17.0.0. Do not guess between 16.38.3 and 17.0.1.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,20 +1,11 @@
 name: Publish
 
-# Publish runs are grouped per pull request, and never cancelled.
-#
-# Under the pull_request trigger github.ref is refs/pull/<N>/merge, so this group is unique to one
-# pull request: simultaneous merges never cancel one another's publish run, and with
-# cancel-in-progress off a run that has started always finishes and cuts its version. It does not
-# serialize releases. Choosing the next version is a read then a create, so two runs still read the
-# same latest version and race to create it - merging eleven pull requests in two minutes on
-# 2026-08-18 gave five runs all computing v16.35.7 and four 403s from secondary rate limiting.
-# A constant group would not fix that and must not be introduced: GitHub keeps only one pending run
-# per group and cancels the earlier pending one, so a third merge silently drops the second's
-# release, which is worse than a noisy 403. The real fix is to stop deriving the version from a
-# read-then-create race - retry the create with backoff, or decouple release from merge so only one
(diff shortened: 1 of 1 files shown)


woksin:
Heads-up before this merges: the missed trigger this evening was **not** the `pull_request: closed` trigger breaking — it was the GitHub Actions platform outage (17:11–19:30 CEST, database primary failure, [githubstatus incident](https://stspg.io/pg14nv9m3095)). Your merge of #3833 at 18:23 **did** trigger Publish — 17 minutes late, once GitHub's queue drained: [run 32989685049](https://github.com/Cratis/Chronicle/actions/runs/32989685049) (`pull_request` event, your actor). It no-op'd only because v16.38.3 had already been released via your manual dispatch.

The thing that actually breaks releases right now is the npm ghost from yesterday's 17.0.0 accident — tracked with a 72h deadline in #3842.

Also worth weighing: with a `push`-to-main trigger, the release action loses the merged-PR event payload it currently reads the semver label from ("Using merged pull request from the event payload"), so label resolution would need a commit→PR lookup that is exactly the path suspected in yesterday's 17.0.0 mishap. Closing this in favor of the existing trigger may be the safer move.

einari:
@woksin There might be something I’m not seeing clearly, but I don’t think we need to change anything in the pipeline. The pipeline has been working perfect up till now (unless something has flown by me).

Issue #3997, opened 2026-09-08
Kernel restart cascades into cluster-wide recovery storm and a permanent, watchdog-unrecoverable observer wedge


einari:
# A single kernel restart cascades into a cluster-wide recovery storm and a permanent, watchdog-unrecoverable observer wedge

## Summary

On a shared Chronicle kernel (16.45.2) hosting ~11 event stores across many
namespaces, restarting the kernel to recover **one** wedged observer instead
destabilized the whole kernel and left ~8 observers in a **permanent** stuck
state that neither the watchdog nor the operator commands (`retry-partition`)
could recover. The only fix was direct MongoDB surgery on observer state.

A kernel restart should be a safe, routine operation. It should not:
- cascade into a cluster-wide startup-recovery storm that saturates the kernel,
- cause the kernel to crash once during boot,
- reject client observation-stream registration while reporting the client as connected, or
- leave observers in a self-perpetuating "preparing catch-up with no catch-up job" loop.

## Environment

- Chronicle kernel **16.45.2**, single silo, shared StatefulSet.
- ~11 event stores (System, Lobby, Studio, StudioAdmin, Catalog, Planner, default, Direct, DirectLobby, Ante, Stagehand) across many namespaces.
- MongoDB storage backend.
- Clients on 16.x. Reactors are `owner: Client`.

## What happened (timeline)

1. One application observer (a reactor) was wedged at an old sequence position and would not advance. To recover it, the kernel StatefulSet was restarted (`kubectl rollout restart`).

2. **Startup-recovery storm.** On boot the kernel began retrying every accumulated failed partition across **all** stores at once, logging hundreds of:
   ```
   warn: Cratis.Chronicle.Observation.Observer
     Starting recovery for partition <part> which has already failed 2 times (max 10).
     This is a startup retry — the subscriber may be stuck
   ```
   (~500 such lines per 90s, spanning cratis-studioissues, cratis-ai, cratis-studio, cratis-chronicle.wolverine, cratis-authproxy, cratis-ante, cratis-stagehand, cratis-cli, cratis-samples, …). The kernel hit ~1.5 cores and **crashed once during boot** (container restartCount 1) before reaching Ready.

3. **Client registration rejected despite "Connected".** While the storm ran, clients logged a successful connection but every reactor observation-stream registration failed:
   ```
   info: Cratis.Chronicle.Connections.ChronicleConnection  Connected to Chronicle
   fail: Cratis.Chronicle.Reactors.Reactors
[...]

Issue #3990, opened 2026-09-07
Do not create databases if not needed


einari:
Chronicle will automatically create databases for all namespaces, even though there is no data being put into it.
This should be lazy and only happen on first insert / upsert of any data.

Typically in Studio and Direct we this for the Lobby services that is basically not a multi tenanted thing, but there are tenant specific namespaces added.

einari:
Fixed in #3999.

Issue #3986, opened 2026-09-07
Deletion of jobs and ignoring of recommendations in Workbench


einari:
It should be possible to select jobs with a checkbox to the left - as a separate column. Then when clicking delete, it will delete all jobs. It should ask if you really want to delete <X> number of jobs.

Then there should be a button for "Select all" at the top.

This same functionality should also be for recommendations, for the ignore button. The ignore button has a weird refresh icon, so that needs to be changed as well.

Worth noting is that ignore in recommendations doesn't work either. Doesn't do anything.

einari:
Fixed in #3999.

Issue #3983, opened 2026-09-07
Displaying data in Projection editor or Read Models should decrypt PII data


einari:
When we look at data in Projections or in Read Models, any data that is encrypted as a consequence of PII should automatically decrypt this.
This should happen on the server before exiting to the client.

<img width="1542" height="1083" alt="Image" src="https://github.com/user-attachments/assets/2b069a98-d685-4bb7-a98b-5b6a75722294" />

einari:
Fixed in #3999.

PR #3999, opened 2026-09-08
Stability fixes: kernel resilience, Workbench UX, and namespace lifecycle

einari:
# Summary

Consolidates six already-open Workbench/Kernel fixes with a further round of stability and Workbench UX work.

## Added

- Server view under System showing running server instances with live CPU/memory (#3982)
- Bulk select and delete for Jobs, and bulk select and ignore for Recommendations, in the Workbench (#3986)
- `ClearFailedPartitions` operator command to recover a wedged observer without direct database access (#3997)

## Changed

- Switching namespace in the Workbench now consistently refreshes the current view, including via URL navigation, browser back/forward, and deep links (#3981)
- Pattern heatmap and causation rendering now recognizes reactors, commands, and ASP.NET requests, and shows identity display names instead of raw subject IDs (#3979)
- "By Silo" renamed to "By Instance" in the Clients view (#3982)
- `RetryPartition` now reports an honest outcome instead of silently succeeding when it did nothing (#3997)
- Chronicle no longer eagerly creates a namespace's database before it has received any data (#3990)

## Fixed

- Navigation flakiness in the Workbench sidebar and routing (#3980)
- Projections list not scrolling (#3978)
- Selecting Seed Data under General no longer also highlights the namespace Seed Data (#3977)
- A silo's startup task no longer hard-crashes on a transient `SiloUnavailableException` from a sibling silo mid-restart (#3967)
- PII-encrypted fields now decrypt correctly when previewing a projection in the Workbench, matching Read Models (#3983)
- Ignore no longer silently fails for Recommendations, and its icon no longer matches Perform's (#3986)
- A kernel restart no longer risks a startup-recovery storm that saturates the kernel (#3997)
- Observers can no longer become permanently wedged in a "preparing catch-up" state that neither the watchdog nor operator commands could recover (#3997)


einari:
Added `Source/Kernel/Core.Specs/Projections/for_ProjectionsManager/when_registering/and_more_definitions_than_the_concurrency_limit_are_new.cs` to `.github/timing-coupling-baseline.txt` as a `knob` entry: it uses `Task.Delay(20)` inside a mocked `Subscribe` call to deliberately widen the interleaving window so the new bounded-concurrency fix (#3997) can be proven to actually allow overlap up to its limit, not just serialize by accident. It's a test double slowed on purpose, not a wait standing in for a missing completion signal.

Issue #3920, opened 2026-09-01
Fence and recover event-sequence content mutations


woksin:
## Goal

Make every in-place event-sequence content mutation durable, fenced, idempotent, and recoverable before Chronicle exposes observer revision barriers.

## Why

Revision, redaction, and event-generation backfill currently mutate historical sequence content without one authoritative operation boundary. A crash or retry can leave partial provider-specific effects, strand repair intent, or allow stale work to overwrite later content. No observer checkpoint can be trustworthy until source-prefix mutation has a durable completion boundary.

## Scope

- Add an internal mutation operation protocol for revision, point redaction, event-source redaction, and generation-content backfill.
- Persist one active operation per event sequence with immutable command data, deterministic operation identity, monotonic ordinal, frozen target boundary, phase, and repair state.
- Fence every effect and verified no-op atomically with the operation ordinal.
- Use canonical storage verification—not cursor exhaustion or filtered counts—to prove source completion.
- Make failures roll-forward-only and recoverable; never clear a partially applied operation.
- Persist repair intent atomically with source commitment and type ambiguous dispatch as repair-unknown.
- Match semantics across MongoDB, SQL, and in-memory storage.
- Add one-way untracked/unsealed/sealed protocol coverage for rolling upgrades.

## Required state

Source operation:

`Reserved -> Applying -> Verifying -> SourceCommitted`

`Applying/Verifying -> Blocked -> previous phase`

Repair:

`Pending -> Dispatching -> Accepted | Unknown`

Coverage:

`Untracked -> Unsealed -> Sealed`

## Acceptance criteria

- [ ] The mutation command is durable before the first content write.
- [ ] At most one unfinished mutation exists per event sequence.
- [ ] The same operation identity and command resumes idempotently; conflicting command data fails typed.
- [ ] Revision/redaction/generation effects and their mutation ordinal are one provider-atomic update.
- [ ] Selective bulk operations stamp every position through a frozen tail as either affected or verified unchanged.
- [ ] Dense-prefix verification proves expected row count and no row below the operation ordinal before source commitment.
- [ ] Appends after the frozen tail are excluded from the operation.
- [ ] Stale workers cannot overwrite equal/newer ordinals or complete a superseded operation.
- [ ] Failures at every phase remain visibly unfinished and recoverable after provider/process recreation.
- [ ] Once application begins, cancellation or failure cannot release the sequence for another mutation.
- [ ] Source commitment persists repair-required/not-required atomically.
[...]

PR #3929, opened 2026-09-03
Superseded by #4013: Add durable event-sequence mutation fencing foundations

woksin:
> **Superseded — not a merge candidate.**

The reviewed and repaired mutation-registry foundations are preserved in draft #4013. They were intentionally excluded from the v18.1.0 reliability release because ordinary revision/redaction/backfill paths do not yet use the registry. The old barrier/capture additions are not release-ready and were not adopted. Original work is preserved on this branch; #3920 remains open. Use #4013 for future foundation work, not this obsolete branch.

Closure is pending the repository's package-cleanup access safeguards. The branch is retained to avoid losing historical work.

---

# Summary

Groundwork for #3920. Revision, redaction, and generation-content backfill mutate historical sequence content today with no single authoritative operation boundary, so a crash or a retry can leave partial provider-specific effects, strand repair intent, or let stale work overwrite later content.

This lands the durable operation protocol and its in-memory provider. **It is deliberately a draft: no runtime path is cut over, and the MongoDB and SQL providers are not written yet**, so the cross-provider parity #3920 requires is not met. Nothing here changes existing behavior — every addition is additive and unreferenced by the current mutation paths.

## Added

- Mutation identity, ordinal and hash concepts, with additive Mongo and SQL schema and migration. SQLite canonical uppercase UUID TEXT checks are hardened against BLOB, case variants, malformed shape and embedded NUL (#3920)
- Canonical mutation identities: strict UTF-8 binary sequence identity with exact display/key round trip, deterministic UUIDv8 mutation ids, and checked definition/receipt digests (#3920)
- A fail-closed mutation state model — an immutable request whose first winning definition freezes the target, a permanent registration lifecycle, and a closed 10-state / 11-edge transition graph with retry and ABA fencing, terminal archive witness and exact archived lineage validation (#3920)
- Registry contracts with exactly four operations — Begin, Transition, Archive, BeginTracking — closed Unknown-zero outcomes, and a fail-loud unsupported registry. Conflict and failure results expose neither command payload nor persisted active object graphs (#3920)
- An in-memory namespace mutation registry making those four operations atomic under one namespace lock, with permanent registrations keyed by mutation id, strict binary sequence keys, and pre-cancellation that writes nothing (#3920)

## Not in this PR

Called out explicitly because #3920's acceptance criteria depend on them:

- MongoDB and SQL permanent registries, and the shared provider conformance suite that proves the three agree
- Cutting Revise, Redact and ReplaceGenerationContent over to the registry
[...]

Issue #3910, opened 2026-08-31
Expose coordinated read-model checkpoints and revision barriers


woksin:
## User need

A consumer that composes one authoritative model from several Chronicle read models must be able to prove that every model represents the same event-sequence revision before publishing a snapshot, export, or generated artifact.

## Problem

Chronicle exposes event-sequence tails and current read-model instances, but no common projection/reducer checkpoint contract or coordinated read barrier across several read models. Reading the underlying Mongo collections in one storage snapshot proves atomic storage visibility, not that independently scheduled observers have all processed the same event tail.

Without a kernel-owned barrier, a consumer can combine a newer application hierarchy with an older event-model projection and produce a plausible but internally inconsistent snapshot. Retrying arbitrary reads or comparing timestamps cannot prove correctness.

This blocks the event-consistent revision boundary required by Cratis/StudioIssues#278 and #289.

## Scope

- Expose the last applied event-sequence position/checkpoint for projected and reduced read models.
- Provide a coordinated read barrier or wait primitive for a requested set of read-model/observer identities against a target sequence position.
- Return typed readiness, timeout, unavailable, failed-partition, replaying, and quarantined outcomes.
- Define behavior across namespaces, event stores, observer generations, replay, failover, and projection registration changes.
- Ensure the checkpoint and read-model state cannot be observed out of order.
- Support in-process/testing implementations with the same semantics.

## Acceptance criteria

- [ ] A caller can read the target event-sequence tail and request that a specified set of read-model observers reach at least that position.
- [ ] Success proves every specified observer has durably applied through the target position and its corresponding read-model state is visible.
- [ ] Timeout/unavailable/failed/replaying/quarantined states are typed and name the blocking observer/partition.
- [ ] The barrier never reports success from wall-clock timestamps, polling read-model contents, or observer running state alone.
- [ ] Namespace, event store, event sequence, observer identity/generation, and target position are explicit in the contract.
- [ ] Replay or observer-generation replacement cannot let a stale checkpoint satisfy a new barrier.
- [ ] MongoDB and in-memory/testing implementations pass the same conformance suite.
- [ ] Specs cover concurrent appends while waiting, lagging observers, failed partitions, replay, cancellation, timeout, namespace isolation, and failover.
- [ ] Documentation explains what revision is proven and what is not.

## Non-goals

- Application-specific model composition.
- Mongo transaction APIs exposed to application code.
- A global Chronicle transaction spanning unrelated event stores.
[...]

Issue #289, opened 2022-05-05
Define properties for indexing using attributes and put them in the model definition


einari:
Depends on #280

einari:
Moving this to Application Model - as we should let things like the MongoDB support we have there use this. This makes more sense as a generic thing and let Chronicle instead leverage it.

PR #4013, opened 2026-09-09
Add reviewed event sequence mutation registry foundations

woksin:
# Summary

Draft kernel-storage foundations for mutation tracking, separated from the reliability fixes merged in #4009. Revision, redaction, and generation-backfill execution are unchanged; this does not complete #3920 and is not ready for release.

## Added

- Add namespace-scoped mutation registration, lifecycle transitions, and retryable terminal receipts for kernel contributors, with in-memory, SQL, and MongoDB providers (#3920).
- Document the registry's storage guarantees and remaining runtime integration boundaries (#3920).


woksin:
## Context: why this exists and why it is a separate draft

The original need came from Studio's authoritative model composition/export work: [StudioIssues #278](https://github.com/Cratis/StudioIssues/issues/278) and [#289](https://github.com/Cratis/StudioIssues/issues/289) needed to compose read models belonging to one consistent model revision. That led to Chronicle #3910 (coordinated checkpoints/barriers), then #3920 (recoverable, fenced historical event mutations as a prerequisite).

GitHub records #3929 as opened by `woksin`, with the persistent-provider implementations subsequently contributed through `einari`'s #3938 into that draft branch. These are account/commit attributions, not a claim about which parts were personally written versus AI-assisted.

During reliability consolidation, the registry was reviewed and repaired, then deliberately excluded from #4009 because ordinary mutation paths do not use it yet. This draft preserves that reviewed work separately instead of shipping unused infrastructure with the immediate fixes.

## What this PR implements

A namespace-scoped **operation ledger**: one active mutation per sequence, immutable registration, increasing ordinals, state transitions, and payload-free terminal receipts. Providers implement mutation-bound compare-and-swap, ordinal-fenced claims, and receipt-first archive recovery so an old operation cannot alter a reused head. Exact retries after head reuse or provider recreation resolve against verified receipts.

This includes the state/contracts, in-memory/MongoDB/SQL implementations, storage migrations, conformance/regression specs, and contributor documentation. It restores the reviewed source through a new additive commit; the original work remains in history and on [`work/consolidate-mutation-registry`](https://github.com/Cratis/Chronicle/tree/work/consolidate-mutation-registry) at `911558cd1f7409b6420c58a28978094668815398`. No history was rewritten.

## What it does NOT do

**It is not yet an event-content fence.** Normal revision, redaction, and generation-backfill paths do not call this registry. Recording `Applying`, `Verifying`, or `SourceCommitted` does not execute or verify those operations.

No public `AppliedThrough`/read-barrier or exact-cut/snapshot API is introduced. It does not establish observer completion, full process-crash recovery, or Studio snapshot consistency.

Before treating #3920 as delivered, we still need:

- Real mutation execution through the protocol, with the immutable request durable before effects.
- Provider-atomic content updates and operation-ordinal fencing, including verified no-ops through a frozen target range.
- Actual dense-prefix verification before source commitment.
- Durable repair intent, dispatch reconciliation, and recovery execution.
- End-to-end stale-worker, retry, crash/restart, and rolling-upgrade tests across supported providers.

**#3920 stays open. Keep this PR draft until its intended delivery boundary is agreed and the remaining work is either completed or explicitly scoped as a foundations-only release.** No release-intent label is being selected merely to make checks green.

## Verification for this restoration

- Source tree matches the previously reviewed combined tree exactly; diff against merged `main` contains only registry foundations.
- Ten affected Debug/Release library builds: zero compiler warnings/errors.
- 595 focused Debug specs passed, none skipped, including actual MongoDB (`mongo:8.0.14`) and SQLite migration/conformance tests.
- Local documentation/frontmatter/TOC/link and whitespace checks passed; no tracked dependency or generated-file drift.

Limits: PostgreSQL/SQL Server execution, Release specs, full Workbench/harness, and the aggregate documentation site were not rerun for this restoration. Earlier broader test runs are historical evidence, not fresh coverage of this PR. Dependency generation also emitted skipped-type/missing-assembly diagnostics; those were retained in local logs rather than represented as compiler warnings or silently hidden.

woksin:
Closing this draft: it was never ready by its own description and is far behind main. I'm keeping the branch for reference for #3920.

PR #4029, opened 2026-09-11
Automatically evolve changed projections and reducers

### LINKED ISSUE #4027: Improving projection and reducers replayability
# Automatic evolution of Reducers and Projections

## Goal

Remove the friction of manually deciding when a changed Reducer or Projection needs to be replayed.

Chronicle should analyze the difference between the registered definition and the new definition, determine the minimum work required to make the read model correct, and perform that work automatically when allowed by the client configuration.

## Change classification

When a definition changes, Chronicle should classify it as one of:

### No action

No replay is necessary when the change cannot affect existing read models.

Examples:

- A new event type is consumed, but no events of that type currently exist.
- A definition change only affects future data.

### Partial replay

Chronicle should determine the smallest safe replay scope for both Projections and Reducers.

A partial replay is possible whenever Chronicle can prove that only part of the historical event set is required to produce the correct new state.

Examples:

- A new event type is added and historical events of that type exist, but its contribution is independent of other events.
- A new event type is added to a Reducer or Projection and, from the first historical occurrence of that event onward, no other event interacts with the same state. Replay can start at that first occurrence rather than rebuilding everything.
- Only streams/read-model keys containing the changed event need rebuilding.
- The mapping for one event type changes and its contribution to the read model is independent.
- A change affects properties that can safely be reconstructed from a subset of events.
- Event groups are independent: if `{A,B}` interact and `{C,D}` interact separately, a change to `A` only requires replaying the `{A,B}` dependency group.
- A Reducer is order-dependent, but only within individual keys/streams. Only affected keys need to be replayed.
- Chronicle has a known-good state before the first affected event and can replay only the suffix from that point.

The definition should provide enough information to derive dependencies such as:

`Event -> operations -> read-model properties / keys`

Replay analysis can then narrow work along several dimensions:

`affected event types -> affected properties/operations -> affected keys -> required history range`

The safe replay scope can therefore range from:

`Specific events -> Specific keys -> Suffix per key -> Entire keys -> Full observer`

Full replay remains the fallback whenever Chronicle cannot prove that a narrower replay is correct.

### Full replay

A full replay is required when Chronicle cannot prove that a partial replay will produce the correct result.

Typical examples include:

- Reducer logic changes where state depends on event ordering.
- Multiple event types interact with the same state in an order-dependent way.
- Identity/key semantics change.
- Filtering changes that may require removing previously projected state.
- Any change where dependency analysis cannot establish that a partial replay is safe.

Correctness wins: when uncertain, fall back to a full replay.

## Client policy

The client should be able to configure how Chronicle handles detected changes.

The default should be to automatically perform whatever evolution Chronicle determines is necessary, including a full replay.

Clients should also be able to disallow automatic full replay, allowing Chronicle to stop at a recommendation when the required operation is considered too expensive or operationally undesirable.

Conceptually, the policy could support:

- `Automatic` - perform no-op, partial replay, or full replay automatically.
- `PartialOnly` - automatically perform safe partial work, but recommend rather than start a full replay.
- `Recommend` - preserve the current recommendation-oriented behavior.

Exact API/configuration naming can be decided during implementation.

## Zero-downtime behavior

Chronicle already supports rebuilding read models side-by-side.

Automatic full replay should use this mechanism:

1. Keep the current generation active for reads.
2. Build the new generation from historical events.
3. Catch up with events produced while the replay is running.
4. Switch to the new generation when it is ready.

This means even a long-running replay can rely on eventual consistency without making the read model unavailable.

## Replay minimization optimizations

Once Chronicle has established that partial replay is safe, it can further reduce the work performed. These should be treated as optimizations on top of the correctness analysis.

### Last relevant event for set operations

For operations that simply set a property, earlier writes to that same property may be irrelevant.

Chronicle can locate the last event that affects the property and replay from that event rather than replaying every historical event that previously set the same value.

For example, if several events set `Name`, but only the latest applicable set determines the current value, only the last relevant event needs to contribute to rebuilding that property.

The same principle can be applied anywhere the operation has overwrite semantics and earlier contributions are fully shadowed by a later event.

### Other optimization opportunities

- Coalesce multiple changes to the same read-model key in memory before writing to the Sink.
- Skip unaffected keys entirely.
- Start from the first affected event rather than the beginning of history when a known-good prefix exists.
- For additive or commutative operations, replay only the contributing event types when their result is independent of ordering.
- Rebuild independent property groups separately when the definition proves they do not interact.
- Stop replaying superseded set operations once a later event completely determines the property value.

These optimizations should never weaken correctness guarantees. If Chronicle cannot prove that an optimization is safe, it should use the broader replay scope.

## Replay performance

Replay should take advantage of Chronicle's ability to process data in memory and the Sink bulk capabilities.

Process replay in batches:

`Read events -> project/reduce in memory -> accumulate mutations -> bulk write to Sink -> checkpoint`

Where multiple events in a batch affect the same read-model key, Chronicle should coalesce those changes in memory where possible and write only the resulting state.

For example, hundreds of Reducer events affecting the same read model may result in a single Sink update for that batch rather than hundreds of individual writes.

Batch size should be configurable/tunable and replay should checkpoint progress so it can recover safely from interruption.

## Observability

Expose what Chronicle decided and why, for example:

- `No replay required - newly observed event type has no historical events.`
- `Partial replay - replaying OrderShipped because only ShippedAt is affected.`
- `Full replay - reducer behavior changed and depends on historical event ordering.`

Progress for partial and full replays should also be available through the existing Chronicle observability/Workbench mechanisms.

## Outcome

Changing a Reducer or Projection should normally require no manual intervention.

Chronicle should automatically determine:

`No action -> Partial replay -> Full replay`

and perform the minimum safe amount of work necessary to bring the read model in line with its new definition.

einari:
## Added

- Chronicle classifies changed projection and reducer definitions as no action, safe partial replay, or full replay and applies the minimum proven work. (#4027)
- Observer configuration supports `Automatic`, `PartialOnly`, and `Manual` definition-evolution policies. (#4027)
- Reducer registrations carry an implementation fingerprint so code changes participate in evolution decisions. (#4027)

## Changed

- Automatic evolution is the default while changes with unproven ordering, key, or property dependencies conservatively fall back to a full replay. (#4027)
- Partial replay applies only newly consumed event types to affected event sources and preserves unaffected observer counts. (#4027)
- Observer configuration and projection registration documentation describe evolution policy, classification, safety boundaries, and operational behavior. (#4027)

Closes #4027


einari:
Consolidated into #4042 (branch feat/automatic-read-model-evolution merged in via a true merge commit — all commits preserved). Closing this one in favor of shipping both together.

PR #4038, opened 2026-09-13
Render migration property names with the configured naming policy

woksin:
# Summary

Event migration maps were written with CLR property names taken verbatim from the declaration, while event payloads are serialized through the configured naming policy. On a client configured with a camelCase policy the two never matched, so `$rename`, `$split`, `$combine` and `$mapValues` silently translated nothing: the map was produced, shipped, accepted and applied against keys that do not exist in the payload.

Silent is the operative word — no throw, no warning. A migration simply had no effect.

No tracking issue exists; found while building an example against `Cratis.Chronicle.Testing`, whose `EventStoreForTesting` configures `CamelCaseNamingPolicy` and therefore always hit this.

## Fixed

- Migration property names are rendered through the configured naming policy where the map is produced, so migrations declared on a camelCase-configured client now match the payload instead of silently doing nothing.

## Notes for review

- No behaviour change under the default policy: `DefaultNamingPolicy.GetPropertyName` is the identity function, so existing PascalCase deployments produce byte-identical maps. A spec pins that.
- Both target keys and the `source`/`sources` values inside each expression are projected.
- The policy is threaded from `EventStore` and `EventStoreForTesting` into `EventTypes` and the migration builders; `PropertyName`/`PropertyExpression` remain the transport types.
- **Worth a maintainer decision:** events appended while the map silently no-opped were stored without their other generations and would need re-migration. Whether a registration change triggers that job is not something I verified.
- String-overload migrators (`pb.Split("FirstName", …)`) are projected too. On a camelCase client that is the fix; on a PascalCase client it is a no-op. Anyone who hand-wrote camelCase strings to work around this would now have them projected again — idempotent for the cases observed, but `ToCamelCase` idempotency for acronym-bearing names (`IPAddress`) is untested and worth a look.
- Enabling `enableEventTypeGenerationValidation` on a camelCase client currently throws `InvalidMigrationPropertyForEventType`; after this it passes.

## Verification

- New specs: camelCase projection across all four expression kinds, and the default policy emitting declared names unchanged.
- `DotNET.Specs` migration suite: 100 tests passing. Client and Testing projects build clean.
- Not covered here: an integration case asserting a generation-1 append reads back upcast on a camelCase client. That belongs alongside `Integration/Client/for_EventSequence/when_appending_event_with_registered_migration/` and would be the decisive end-to-end proof.


woksin:
Consolidated into #4039, preserving this PR’s commits and adding nested-path, constructor-compatibility, and real transport regression coverage. Closing as superseded, without merging to main or deleting the branch. Please review the combined change in #4039.

Issue #4041, opened 2026-09-14
Single-event Append no longer normalizes an unspecified EventSourceType to Default


woksin:
## Summary

Since the 16.x → 18.x gRPC rewrite, the single-event `Append` path no longer normalizes an
unspecified `EventSourceType` to `EventSourceType.Default`. The value is stored verbatim as
`Unspecified` (`""`).

Any client that appends without naming an event-source type and later reads the event back to
compare its route sees `""` where it wrote what it believed was `Default`. `AppendMany` is
unaffected — it still normalizes — so the two append paths persist **different** event-source types
for the same logical input.

This surfaced in `Cratis/Chronicle.Wolverine`, where it broke **all seven** ingress
duplicate-resolution integration tests: every legitimate duplicate was rejected as a route conflict.

## Evidence

**16.38.4 normalized on the way in** — `Source/Kernel/Core/Services/EventSequences/EventSequences.cs:55`:

```csharp
var result = await eventSequence.Append(
    request.EventSourceType is null ? EventSourceType.Default : (EventSourceType)request.EventSourceType,
    request.EventSourceId,
    request.EventStreamType is null ? EventStreamType.All : (EventStreamType)request.EventStreamType,
    request.EventStreamId is null ? EventStreamId.Default : (EventStreamId)request.EventStreamId,
```

**18.1.4 and 18.2.0 cast unconditionally** — `Source/Kernel/Grpc/Sequences/EventSequences.cs` (generated):

```csharp
new Cratis.Chronicle.Sequences.Append(
    ...,
    (Cratis.Chronicle.Concepts.Events.EventSourceType)request.EventSourceType,
    ...)
```

**The client sends an empty string, while computing the resolved value for other purposes** —
`Source/Clients/DotNET/EventSequences/EventSequence.cs`:

```csharp
:108   var resolvedEventSourceType = eventSourceType ?? EventSourceType.Default;
:114       .GetScope(eventSourceId, resolvedEventStreamType, resolvedEventStreamId, resolvedEventSourceType);
...
:135   EventSourceType = eventSourceType?.Value ?? string.Empty,
```

So the **concurrency scope is computed with `Default`** while the **persisted value is `""`**. That
internal disagreement inside one method is the clearest signal this is unintended.

**`AppendMany` still normalizes** — `Source/Kernel/Core/Sequences/AppendMany.cs:62`:

```csharp
EventSourceType.Default,
```

**The concept already treats the two as the same thing** —
[...]

PR #4045, opened 2026-09-16
Return authoritative append metadata from the kernel

woksin:
# Summary

Make Chronicle the authority for append metadata so applications can observe the event that was actually stored, rather than reconstructing routing, timestamps, and audit information in each client.

## Added

- Opt-in append receipts containing persisted routing, sequence numbers, occurrence times, subjects, tags, hashes, and audit metadata. Batch receipts follow input order; rejected appends have no persisted receipts.
- .NET append results and notifications expose the server-acknowledged event context. Malformed acknowledgments are reported explicitly without implying that retrying the append is safe.

## Fixed

- Preserve authorization verdicts across .NET, JVM, JavaScript, Elixir, and Python protobuf clients instead of omitting a successful verdict from the wire.
- Apply one kernel policy to omitted routing metadata across all append endpoints, while preserving explicit routes and existing query-filter semantics. (#4041)
- Match MongoDB append acknowledgments to stored timestamp precision and preserve supplied hashes in in-memory storage.
- Prevent concurrent gRPC generator builds from causing multi-target NuGet publication failures.

## Compatibility

Existing clients continue receiving compact acknowledgments unless they request receipts. Existing wire fields and C# command constructors are preserved. Receipt-aware clients require the new kernel contract; upgrade every kernel node before upgrading those clients. Existing events are not rewritten.


woksin:
Local verification: full Core.Specs 3,687; in-memory storage 207; MongoDB timestamp precision 5; generator specs 97; .NET client 3,023 and Testing.Specs 571 in both Release and Debug; 96 real out-of-process SQLite append/receipt checks. Integration build has zero warnings/errors. Wire compatibility served 18.0.0, 18.1.0, 18.2.0, and 18.3.0. Python generated contracts: 7 tests and lint passed. Multi-target NuGet packaging passed for net8/net9/net10, with generated client implementations present and the generator excluded. Documentation: 1,093 files and 50 navigation files passed. Independent kernel review identified MongoDB timestamp precision; the provider and deterministic read-back specs were corrected. Receipt opt-in preserves legacy response size, existing command constructors, and original grain method signatures. This kernel/contract release must land before the coordinated Kotlin/Java, TypeScript/JavaScript, and Elixir SDK changes.

woksin:
Cross-runtime verification confirmed another prerequisite: C# success envelopes omitted IsAuthorized=true, which proto3 clients decoded as false. Added an observed-red regression for all three envelope shapes and fixed emission of both true/false on the existing field. Field numbers/types are unchanged; all four 18.x wire baselines remain served. Reverification: 3,025 client specs and 96 real append/receipt integration checks passed. Previous CI had one MongoDB projection re-registration/replay shard fail; added its job failure details to the assertion, without weakening it, and the new head is rerunning the complete matrix. Local MongoDB cannot start because its image explicitly rejects this host Linux-kernel version; no guard was bypassed.

woksin:
Added bf5e33093 to preserve the released public equality/hash semantics of EventForEventSourceId while still tracking omitted routes internally. Coverage includes all public values, derived records, base/object references, hash sets, and record copies.

Verification on this head:
- .NET client: 3,055 specs passed in both Debug and Release.
- Testing package: 571 specs passed.
- Explicit multi-target pack: net8.0/net9.0/net10.0, with reference assemblies and generated runtime implementations verified and the build-only generator excluded.
- Independent narrow compatibility review: no blockers.

The previous CI failure was a kernel-container startup error (internal TLS port 35000 already in use), before append assertions. A fresh current-head CI run is underway; this is not a waiver of that gate. Coordinated SDK package pins and releases remain pending the actual kernel/contracts publication.

Issue #4049, opened 2026-09-17
getForEventSourceIdAndEventTypes silently ignores the event source type filter


woksin:
## Summary

`getForEventSourceIdAndEventTypes` takes an event-source-type argument in the clients, but the wire contract it maps to has no field for it, so the value is silently discarded and the read is not narrowed. A caller who passes it believes they filtered and did not.

## Evidence

`ForEventSourceIdAndEventTypesRequest` in `Source/Kernel/Protobuf/sequences.proto` carries `EventStore`, `Namespace`, `EventSequenceId`, `EventSourceId`, `EventTypeIds`, `EventStreamType` and `EventStreamId` - no event source type. The Kotlin and TypeScript clients both expose the parameter; it reaches nothing.

## Why this is worth fixing rather than deleting the parameter

The kernel already supports the filter one layer down. `Source/Kernel/Storage/EventSequences/EventSequenceQueryCriteria.cs` has `EventSourceType`, `HasEventSourceType` and applies it when matching a context, and `EventSequenceQueryCriteriaFactory.Create` maps a narrowing onto it. Only the request contract for this one read path is missing the field, so the capability exists and is simply unreachable from that call.

## Suggested fix

1. Add `EventSourceType` to `ForEventSourceIdAndEventTypesRequest` - additive, next free field number, so every released baseline stays served.
2. Map it through the existing narrowing into the criteria the storage layer already honours.
3. Have the clients send it, and drop the "not applied" notes that currently document the gap in the Kotlin reference and migration guide and in the TypeScript `IEventSequence` docs.

Blank keeps its existing meaning of "do not narrow", so behavior only changes for a caller who was already passing a value expecting it to filter.

## Affected clients

Kotlin and TypeScript document the parameter as inert as of their current majors. The .NET client should be checked for the same call shape before this is closed.

woksin:
Fixed in #4053 (`d7e3af85e`), merged as `af77020c1`.

The scope in this issue was not quite right, and the difference mattered. The kernel did not merely lack a request field: `GetFromSequenceNumber` had no event source type parameter at **any** layer, unlike `GetTailSequenceNumber` beside it. `EventSequenceQueryCriteria.EventSourceType` belongs to the criteria-based presentation reads (`GetPage`, `GetCountMatching`, `GetHistogram`), which this call never touches. So the filter had to be threaded through the storage interface, all three implementations and the replay decorator — not just added to the contract.

[...]

PR #4054, opened 2026-09-17
Apply the event source type filter instead of silently dropping it

woksin:
# Summary

`getForEventSourceIdAndEventTypes` took an event source type in every client and threw it away, so a caller asking for one event source type got every source type back with no way to tell.

The capability was not merely unreachable from the request contract. `GetFromSequenceNumber` had no event source type parameter at any layer — unlike `GetTailSequenceNumber` beside it — so the filter is threaded through the storage interface and all three implementations, each reusing the sentinel guard that method already establishes. The contract field is derived from the Core artifact rather than hand-written, and every released 18.x baseline is still served.

Only a caller who was already passing a value expecting it to filter sees a change: an unspecified or `Default` type still means "do not narrow".

## Fixed

- `GetForEventSourceIdAndEventTypes` now applies its event source type argument instead of ignoring it, so a read narrowed to one event source type no longer returns events of every other type (#4049)

## Added

- `EventSourceType` on `ForEventSourceIdAndEventTypesRequest`, additive and wire-compatible with every released 18.x baseline (#4049)


woksin:
Folded into #4053 so the whole follow-up ships as one release rather than two. The commit `d7e3af85e` is preserved there through a merge commit — nothing was squashed or rebased, and this branch is left intact.

Issue #4058, opened 2026-09-17
A newer client refuses an older kernel: the connect-time compatibility check reads every added field as a removal


einari:
## What happened

On 2026-09-17 Direct 2.63.11 went to production carrying Chronicle client **18.4.1**, against the shared kernel, which runs **18.1.4**. The client refused every connection with `IncompatibleServerException`, naming `ReducerDefinition.Hash` as a field the server no longer serves. The refusal happens inside `ConnectInternal`, before the connection is established, so the host could not start: crash loop, 0/1 ready, production down from ~12:08 to ~12:41 UTC, recovered only by a manual rollback to 2.63.9. Direct is back on Chronicle 18.1.6 / Arc 22.14.0 (Cratis/Direct#1063).

## There was no incompatibility

The entire wire delta between `v18.1.4` and `v18.4.1` is one field:

```
$ git diff v18.1.4 v18.4.1 -- 'Source/Kernel/Protobuf/*.proto' \
    | grep -E '^[+-]' | grep -vE '^(\+\+\+|---)' | sort | uniq -c | sort -rn
  46 +   bool IsAuthorized = 2;
  46 -   bool IsAuthorized = 2; // default value could not be applied: True
   1 +   string Hash = 9;
```

46 of those 47 lines are a proto **comment** being dropped — not something a descriptor set even carries. The one real change is `string Hash = 9` added to `ReducerDefinition` in 33c46f1d7 ("Automate projection and reducer evolution"), first released in 18.3.0. `chronicle.desc` grew by 20 bytes across three minor versions.

`ReducerDefinition` travels **client → server**: the client computes it (`Source/Clients/DotNET/Reducers/Reducers.cs:430`, `ReducerFingerprint.Create`) and sends it in `RegisterReducer`. An 18.1.4 kernel does with field 9 exactly what proto3 says to do — ignores it. The only consequence of the skew is that automatic reducer evolution does not engage against an old kernel: silent, correct degradation.

An 18.4.1 client and an 18.1.4 kernel are wire-compatible. **The outage was manufactured entirely by the compatibility check.**

## Root cause: the check asks the backward question with the roles reversed

`WireCompatibilityChecker` answers exactly one question — *is everything the **older** side expects still present, unchanged, in the **newer** one?* — and its own remarks state that additions are never reported. That is right, and it is right for the release gate, which passes HEAD as `actual` and a released tag as `expected`.

[...]

PR #4059, opened 2026-09-17
Run the wire compatibility gate on pull requests

einari:
# Summary

The wire compatibility gate has never run on a pull request. `workflow_call` was its only trigger and `publish.yml` its only caller, so it could block a release but never a merge — a contract break landed in `main` and the first person to meet it was whoever cut the next release. On 2026-09-17 a client-breaking contract change reached production through that gap (#4058).

It was written for a pull request run that was never wired up: the `major` label it honors is read from `github.event.pull_request.labels.*.name`, which is always `null` under `workflow_call`, and `version` is optional so that a run without a release version falls back to the latest tag — a fallback with no caller. `publish.yml:102` even says *"Unlike the pull request run…"*.

This adds the trigger. Nothing downstream of it changes: baseline discovery, the floor, the withdrawals, the step summary and the exit codes all already worked, once per release instead of once per pull request.

## Changed

- The wire compatibility gate now runs on pull requests as well as at publish, so a change that breaks the contract for clients already running against the current major is caught before it merges rather than by whoever cuts the next release (#4058)

## Notes for review

**`paths-ignore`, not `paths`.** An include list fails closed — a contract change on a path nobody thought to list is not checked and the gate reports success having looked at nothing. That is how a `Directory.Packages.props` bump once merged with no build at all, the lesson `.github/build-affecting-paths.txt` exists to keep. The ignore list names only things that cannot change the wire contract, so anything unforeseen still runs the gate. A run costs about a minute.

**Two hygiene fixes in the same file**, both required by `github-actions.md` and both previously absent because the workflow never ran in a context where they mattered:

- `timeout-minutes: 30` — the default is six hours, and one hung run holds a slot in the organization's shared concurrency pool for all of it.
- a concurrency group keyed to this workflow's own name, not `github.workflow`, which in a called reusable workflow resolves to the *caller* and would put a publish's gate in the publish group where a later pull request could cancel it. `cancel-in-progress` is restricted to `pull_request` for the same reason.

**Please make this a required check** when merging — the trigger alone makes it report; the branch protection is what makes it gate.

## Verification

The gate was run locally exactly as the pull request path invokes it (no `--since`, since the floor names major 17 and the current major is 18):

```
$ dotnet run --project Source/Tools/WireCompatibility/WireCompatibility.csproj -- \
    --major 18 --withdrawn 18.4.0 --current Source/Kernel/Protobuf/chronicle.desc --allow-missing-baseline

  checked 18.0.0 / 18.1.0 / 18.2.0 / 18.3.0 / 18.5.0
  Every released baseline is still served.        (exit 0)
```

`Source/Kernel/Protobuf` is clean after a Release build of `Contracts.csproj`, so the committed descriptor set still matches the contracts source and the comparison above is against the real contract.

## What this does not do

This fixes *when* the gate runs, not *which question it asks*. It still only asks whether HEAD serves clients built against released contracts — old client against new kernel. The direction that caused #4058, a **new client against an older kernel**, is still unasked, and this change would not have caught that break. That is the next pull request, and it lands with the client-side fix, because the gate's fatality rule is defined by what the client tolerates: enabling it on its own would make `main` permanently red against contract additions that have already shipped.


einari:
Closing: the premise is false. `dotnet-build.yml:52` already calls this workflow on every pull request, and `dotnet-build` `needs: [wire-compatibility]` — the gate has been running in the right place all along. On #4042, the pull request that added `ReducerDefinition.Hash`, `wire-compatibility / wire-compatibility` reports `success`. It passed because `Check(baseline, current)` never reports an addition, not because it was absent.

I got here from a truncated `grep | head -20`; `publish.yml` has eighteen matches and the `dotnet-build.yml` caller was on line 21. Full reasoning and the corrected analysis are in #4058.

This change was also actively harmful: a second `pull_request` trigger creates a duplicate run that lands in the same concurrency group as the one `dotnet-build` calls, and the two cancel each other. That is visible in this PR's own checks — the standalone run was cancelled 1s in.

The one genuine finding here was that the job has no `timeout-minutes`, which `github-actions.md` requires. Splitting that out on its own.

